PR #3, opened 2015-06-04
Write HTML to fmt::Write instead of String.

Stebalien:
My motivation is that I would like to use this library in a static blogging
engine but I use a custom buffer type.

The downside is error handling. This may significantly impact performance (I
don't know what benchmarks you usually run). The upside is that users can write
directly to custom buffers as long as those buffers implement fmt::Write.

Error Handling Alternatives:
 * Panic on error and tell users to not pass in writers that can fail. Users
   can always record write errors on the side.
 * Provide a custom Write trait that doesn't support error reporting.

Also, I kept the current `fresh_line` behavior but we could probably make this
faster if we allow extra newlines.

Compatibility note: This requires rust 1.1 beta for Write::write_char.

FYI, I've signed the CLA.


Aatch:
I think it'd be better to use `io::Write` instead of `fmt::Write`, so you can write directly to a file or similar. Tests that me an @raphlinus did seem to suggest that writing the output is a surprisingly large amount of the time. Using `io::Write` is slight trickier, but for the most part the `write_str`s can be replaced with `write_all` and byte-string literals.

Aatch:
Actually, looking more carefully at the code, it'd be trivial to switch to `io::Write` given that you've already abstracted the "buffer".

Stebalien:
`io::Write` is won't work because it doesn't guarantee utf8 encoding which makes it impossible to safely write to a `String`. If you want to support `io::Write`, I could implement an adapter (like the one used in `io::Write::fmt`) and provide two functions: `push_html_fmt`, `push_html_io`.

Aatch:
@Stebalien but isn't almost everything pushed either already a `str` or is a literal? I'm not sure what the issue is.

Stebalien:
@Aatch It is. The problem is that `io::Write` throws away the fact that the data is utf8 so if you want to write to a String, you need to re-assure rust that the data really is utf8. Basically, if you want to be able to write to both files and strings, there are two ways to go:

1. Accept `io::Write` and implement `struct IPromiseThisIsUtf8WriteAdapter<W: fmt::Write>(&mut W)` (implements `io::Write`).
2. Accept `fmt::Write` and implement `struct WriteAdapter<W: io::Write>(&mut W)` (implements `fmt::Write`).

The second method is safer because it doesn't convert from utf8 to raw bytes and back again.

Aatch:
@Stebalien ah, I didn't think of that. I definitely think we should support both options, assuming that `push_html_io` is just a helper fn over `push_html_fmt` I can't see anything wrong with doing it that way.

raphlinus:
I wanted to benchmark this, but it seems to introduce logic bug. The easiest way to provoke it is `RUST_BACKTRACE=1 cargo run -- -s ~/dl/cmark/test/spec.txt 50`, where [spec.txt](https://raw.githubusercontent.com/jgm/cmark/master/test/spec.txt) is from [cmark](https://github.com/jgm/cmark). That's [Example 50](http://spec.commonmark.org/0.19/#example-50), which doesn't seem particularly exotic.

Backtrace below:

```
thread '<main>' panicked at 'arithmetic operation overflowed', src/html.rs:37
stack backtrace:
   1:        0x107192145 - sys::backtrace::write::hce8edc3761615a91fas
   2:        0x107195794 - panicking::on_panic::haf19346a6a6623d7vvw
   3:        0x10718c018 - rt::unwind::begin_unwind_inner::h84a637bb4fc9a013Ndw
   4:        0x10718c63c - rt::unwind::begin_unwind_fmt::he799ce251af7040fTcw
   5:        0x10719504c - rust_begin_unwind
   6:        0x1071b6af5 - panicking::panic_fmt::h16e5884fd9bf851ci5B
   7:        0x1071b63d4 - panicking::panic::h37c142b5f2d1829eP3B
   8:        0x107118b0c - html::FreshLineAdapter<'b, W>.Write::write_str::h5689971399085283464
   9:        0x10711b244 - escape::escape_html::h11118945503011952733
  10:        0x107115b48 - html::Ctx<'b, I, W>::run::h2540889576962225024
  11:        0x107115630 - html::push_html::h3609764159648968415
  12:        0x107113be7 - render_html::h56419a67795c2b60qaa
  13:        0x1071290d0 - run_spec::hcec7f0b6257f74f5Bja
  14:        0x10712c0ac - main::h073de8bb5db93de6bsa
  15:        0x1071970b8 - rust_try_inner
  16:        0x1071970a5 - rust_try
  17:        0x107195fa0 - rt::lang_start::h9fbea85581be889fzqw
  18:        0x10713f75e - main
[50]An unknown error occurred
```

I'm ok with using a Writer of some kind instead of a string, as long as it doesn't negatively impact performance. My sense is that the API with string is easier because you don't have to worry about errors, and pretty much just as flexible. I haven't measured it carefully, but I suspect that the actual cost of the extra string copy is pretty minimal, as it should all be done with wide SIMD instructions on modern CPU's.

src/html.rs
```diff
@@ -15,140 +15,161 @@
...
-            self.buf.push('\n');
+struct FreshLineAdapter<'b, W: 'b> {
+    writer: &'b mut W,
+    has_fresh_line: bool,
+}
+
+impl<'b, W> Write for FreshLineAdapter<'b, W> where W: Write + 'b {
+    #[inline]
+    fn write_str(&mut self, s: &str) -> fmt::Result {
+        self.has_fresh_line = s.as_bytes()[s.len()-1] == b'\n';
```
Aatch: This will fail if `s` is an empty string, `s.len() - 1` will underflow.

Stebalien:
Sorry about that. Fixed (although I'm still failing the 4 spec tests that master fails).

Stebalien:
I'm seeing a 0.96x speedup (i.e. slowdown) when benching the spec so it's not too bad. Also, I've added a method for pushing to an io::Write. Would you like that now or in a new pull request?

Stebalien:
Ping.

raphlinus:
I'm on vacation right now and away from my development machine. In any
case, I am skeptical about this patch. I'll try to give a full review when
I get back.

PR #10, opened 2015-08-16
Implement hoedown-style footnotes.

### LINKED ISSUE #4: footnotes support
Not sure what your intentions are with implementing things that aren't in CommonMark, so feel free to close these if they're not relevant, though personally I'd love to see them available here.

Footnotes is a big one for me.

notriddle:
Closes #4.

### DIFF
--- a/specs/footnotes.txt
+++ b/specs/footnotes.txt
@@ -1,0 +1,154 @@
+Run this with `cargo run -- -F -s specs/footnotes.txt`.
+
+This is how footnotes are basically used.
+
+.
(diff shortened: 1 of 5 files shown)


notriddle:
Are there any other extensions blocking this for rustdoc?

raphlinus:
Looking through <https://users.rust-lang.org/t/new-commonmark-parser/1690/11>, I think that covers all the flags passed to hoedown, except for superscript and strikethrough. I don't know to what extent either is used in Rust-land.

Reviewing now.

PR #30, opened 2016-08-05
[WIP] Update for spec v0.26

azerupi:
:warning: **Work in progress, do not merge**

This is an effort to update pulldown-cmark to the the [latest spec (v0.26)](http://spec.commonmark.org/0.26/spec.txt). It's not complete and I will probably not have the time and skill to finish this alone, so I thought I would post it so that it could be a  collaborative effort.

The test case format in the spec has been modified from

```
.
<input>
.
<output>
.
```

to

`````````````````````````````````
```````````````````````````````` example
<input>
.
<output>
````````````````````````````````
`````````````````````````````````

So I changed `main.rs` to correctly extract the new format. I wrote an iterator that returns the test cases from the spec and I think it turns out a little more readable than before.
## Failing tests

According to the official commonmark tool to test spec compliance, we pass `565/618` tests. But according to the pulldown-cmark test we pass only `504/618` tests. I think this is due to the fact that html is not normalized before testing equality in `pulldown-cmark` making it stricter than necessary.

Some (a lot?) regressions are du to tab handling. In the new spec some tabs are converted to spaces but some aren't. If I am not mistaken, pulldown-cmark converts all tabs to 4 spaces?

---

If you want to help, just fix some test(s) and create a PR against this PR :)

### DIFF
--- a/specs/spec-0.26.txt
+++ b/specs/spec-0.26.txt
@@ -1,0 +1,9305 @@
+# Introduction
+
+## What is Markdown?
+
+Markdown is a plain text format for writing structured documents,
+based on conventions used for indicating formatting in email and
+usenet posts.  It was developed in 2004 by John Gruber, who wrote
+the first Markdown-to-HTML converter in Perl, and it soon became
+ubiquitous.  In the next decade, dozens of implementations were
+developed in many languages.  Some extended the original
+Markdown syntax with conventions for footnotes, tables, and
+other document elements.  Some allowed Markdown documents to be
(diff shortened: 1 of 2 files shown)


azerupi:
I signed it!

raphlinus:
Thanks for doing this! I'm short on time right now, so really appreciate people from the community fixing this.

There's one tricky area around deeply nested inline markup. I worked out an algorithm for it, but haven't had time to implement it yet. Probably people should coordinate before taking this on.

Issue #35, opened 2016-11-05
Panic on invalid command line arguments


Victor-Savu:
It seemed really odd that the `pulldown-cmark` executable panics when the command line arguments passed are not recognized. (More surprising even when the arguments which failed were `--help` and `-h`)
I would propose two things: 
1. Add a command line option which prints the command line help menu to standard output (`h`, `help` would probably be in line with the current command line options style)
1. Handle the invalid command line arguments case by:
  a) printing a short message to standard error which contains at the very least the command to access the help information __and__
  b) exiting with an error code such that scripts which call the executable can react to the error

I would love to know what you think about this proposal:
 - if it is valid, what could be improved to make it great
 - if it is invalid, what are the drawbacks and what are better alternatives and why (one of the alternatives is, of course, the current state)

If desirable, I would be happy to contribute by implementing this change.

Victor-Savu:
Since [clap](https://github.com/kbknapp/clap-rs) has been seeing much more love from the ecosystem than [getopts](https://github.com/rust-lang-nursery/getopts), would it be acceptable to move to it?

PR #36, opened 2016-11-05
Add command line usage info

Victor-Savu:
Add the `help` (shorthand: `h`) option for printing command line usage information to standard output. On incorrect usage, print the usage information to standard error and exit with exit code 1 (before, the program would just panic).

Fixes: #35


Victor-Savu:
I signed it!

PR #37, opened 2016-11-15
Add table alignment parsing/rendering

kivikakk:
I'm not particularly happy with the location of `scanners::Alignment`, but it does the job.


raphlinus:
Generally looks good (and sorry for the delay). I'd prefer the `Alignment` to go into parse, that's where the whole public API is, so I moved it before merging. Thanks!

kivikakk:
Ah, lovely; thank you!

PR #38, opened 2016-12-06
Apply emphasis/strong when split across multiple lines in a container.

elahn:
```
> *two
> words*
```
now produces:
```
<blockquote>
<p><em>two
words</em></p>
</blockquote>
```
instead of:
```
<blockquote>
<p>*two
words*</p>
</blockquote>
```


elahn:
I signed it!

raphlinus:
Good catch, thanks!

PR #39, opened 2016-12-27
Generate tests from spec files

azerupi:
This PR builds upon #30 (and replaces it).

- ~~The 0.27 spec is added to the specs folder~~
- The other spec files are updated to the new test case format.
- A new build feature is introduced: `gen-tests`. When activated, the build script will generate unit tests from all the spec files in the specs directory.

This allows to run the spec with `cargo test` and updating the spec is just a matter of regenerating the test files with `cargo test --features=gen-tests`.

The run-time spec runner has also been updated to the new format.

------------

Currently pulldown-cmark passes `508 / 622`  tests from the spec.


raphlinus:
I like the idea of this. My organization has specific requirements for importing third party modules, which is how I believe the spec can be handled. There should be a single commit that adds third_party/CommonMark, which should be a clean import of a subset of files - I think at a minimum it should include LICENSE and spec.txt. I'm looking at my guideline document and it's specific enough I think it would save time if I did it, then you can rebase this change.

The build.rs file needs to have a copyright header matching the style of the project.

As I understand it, this build.rs will be invoked for all users of the library, which is something I'd like to avoid. Can it be made something like a dev dependency? Alternatively the binary could be a separate crate. But if the "work" it does is gated by a feature, maybe it's not so bad.

Is it necessary to check in tests/{table,footnote}.rs if they're autogenerated by the build script?

Thanks for your contribution, and sorry I didn't get around to looking at your earlier version.

azerupi:
>Is it necessary to check in tests/{table,footnote}.rs if they're autogenerated by the build script?

That depends, you could always run the tests with `cargo test --features=gen-tests` but it will needlessly regenerate the tests.

>As I understand it, this build.rs will be invoked for all users of the library, which is something I'd like to avoid. Can it be made something like a dev dependency?

I can understand that. 
I am wondering what the overhead is though. Considering that the whole file is `cfg`ed and there remains only an empty function when the feature isn't enabled, I would assume everything is optimized away and the overhead would be minimal. But I could be wrong. 

Do you worry about something other than time overhead?

Alternatively, we could avoid using a build script altogether. The script generating the tests could be executed with a command like [cargo-script](https://github.com/DanielKeep/cargo-script).

So instead of having a build script gated behind a feature, we would have an "orphan" script that you could run with `cargo script gen-tests` (if you have cargo-script installed) or `rustc gen-tests.rs && ./gen-tests`. 

Let me know what you prefer :)

raphlinus:
I added the spec (#40). Feel free to rebase this patch, and I'll merge it. Thanks!

azerupi:
I rebased.
I had to hardcode the path to the spec because it's no more alongside the others in the specs/ directory.  But I don't think it's a problem.

raphlinus:
Awesome, thanks! This comes at a very good time, I was just experimenting with a new algorithm for inline markup and finding that I couldn't run the tests because the format changed.

The path to the spec seems fine.

A few observations: I get a bunch of warnings when compiling with `--features=gen-tests`. I didn't dig too deeply.

Most of the table tests are failing. I'm not sure why, as when I ran the body of test 1 through the command line, I got the correct result.

The source file base should probably be incorporated into the name of the test, e.g. footnotes_test_1 instead of spec_test_1.

I was worried that cargo wouldn't know to rerun the build script if a spec changed, so you'd need a `cargo:rerun-if-changed` output, but it seems to magically work.

All this can be fixed later, I'm in no hurry.

PR #48, opened 2017-03-12
Full html spec

GuillaumeGomez:
Sorry, only implemented part of the HTML spec in my last PR. This time it's complete.


raphlinus:
Overall, I'm fine with these changes, but I do have one concern: it looks like you're editing `tests/spec.rs` by hand, where I believe this file is auto-generated from the spec (which itself is stored in `third_party/cmark`). I think it'd be fine if you moved the hand-written tests into a non-autogenerated file.

Feel free to correct me if I'm wrong about this, I didn't do the autogenerated spec tests myself.

GuillaumeGomez:
I have no problem moving the new tests to another file. If `tests/spec.rs` is autogenerated, might be nice to say it in the header or somewhere. ;)

GuillaumeGomez:
Ok, added a new file and moved my tests to this one.

notriddle:
> tests/hmtl.rs

GuillaumeGomez:
@notriddle: You haven't see anything. It's just between you and me.

raphlinus:
Thanks!

raphlinus:
Pushed 0.0.10 to crates.io at this commit.

PR #50, opened 2017-03-13
Ignore list items that overflow the int limit

### LINKED ISSUE #49: Overflow ParseIntError in scan_listitem
Running [this] fuzzer script for a few seconds returns

```
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: ParseIntError { kind: Overflow }', /checkout/src/libcore/result.rs:860
```

which according to the backtrace is from this code [here].

This input it crashed on was this:

> 4444444444444444444444444444444444444444444444444444444444!

or, as base64:

> NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NDQ0NCEK

[this]: https://github.com/rust-fuzz/targets/tree/d745b476a7128370adf78b72d1813e9e4c118ed1/pulldown-cmark
[here]: https://github.com/google/pulldown-cmark/blob/a35d1223afef0f01fc116afc57352b4dbf8252ed/src/scanners.rs#L393

notriddle:
Fixes #49

### DIFF
--- a/src/scanners.rs
+++ b/src/scanners.rs
@@ -390,8 +390,11 @@ pub fn scan_listitem(data: &str) -> (usize, u8, usize, usize) {
         b'0' ... b'9' => {
             let mut i = 1;
             i += scan_while(&data[i..], is_digit);
-            start = data[..i].parse().unwrap();
             if i >= data.len() { return (0, 0, 0, 0); }
+            start = match data[..i].parse() {
+                Ok(start) => start,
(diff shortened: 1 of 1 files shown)


raphlinus:
Thanks!

PR #55, opened 2017-03-21
Split out binary and lib build

GuillaumeGomez:
Otherwise we can't use it in rust because of the `getopts` duplicate.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "pulldown-cmark"
-version = "0.0.10"
+version = "0.0.11"
 authors = [ "Raph Levien <raph@google.com>" ]
 license = "MIT"
 description = "A pull parser for CommonMark"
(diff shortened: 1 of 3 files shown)


raphlinus:
I'm not 100% sure I understand why this is needed, but am ok with it if it's the cleanest way to solve the problem.

Any reason you didn't make the binary the default, and turn off the feature in clients?

GuillaumeGomez:
I can, it depends on what you prefer. But as far as I know, people use pulldown-cmark for its lib, not its binary.

And for the explanations: we need to remove the `getopts` duplicate (the version is different than the one used in rust) so rust tests can run on rustdoc without this failure. And I think you'd prefer avoid an update on your `getopts` version every time rust does it as well. ;)

raphlinus:
Thanks!

GuillaumeGomez:
Thanks to you! :)

GuillaumeGomez:
Can you ping me once it's published please?

raphlinus:
Now "cargo publish" fails, saying `error: main function not found`. Know what I should do? Seems to me defaulting to build the binary is one way to fix that...

GuillaumeGomez:
I think I know where it comes from, just a second.

Issue #57, opened 2017-03-22
-s option is more strict than official spec tester


kivikakk:
I tried `pulldown-cmark -s` against the current `master` spec (https://github.com/jgm/CommonMark/commit/f3ceed3013a6b719af91dc3323f3888e9dbe8406) and noted it reports 504/621, but `python3 test/spec_tests.py --program=pulldown-cmark` reports 564/621.

I suspect the difference might lie in the normalisation done in CommonMark's `spec_tests.py` by default, though disabling it with `--no-normalization` only makes a small difference to the results.

kivikakk:
(Further playing around reveals _completely ignoring_ whitespace in the comparison in `-s` brings us up to 562/621, but that's obviously not actually what we want to be doing.)

kivikakk:
The official spec tester uses a HTML parser to normalise the trees. We could try the same approach.

Would you be interested in a PR which attempted this? I'd be more inclined to rely on the official tester than reproduce its efforts (e.g. vendor `jgm/CommonMark` in as a submodule, and then have CI call the official tester with the `spec.txt` in its tree against the built `pulldown-cmark`), but I can also see the merits in the binary being able to test itself like this.

kivikakk:
(I got a little bored/interested, so I have some WIP at https://github.com/google/pulldown-cmark/compare/master...kivikakk:normalize-html.)

PR #58, opened 2017-03-22
Tab characters may appear in expected too

kivikakk:
Refs #57. This doesn't actually bump up the spec success count yet since the tests it's implicated in are failing for other reasons.


kivikakk:
Different email address, probably. Will use my company email address for commits here.

raphlinus:
Thanks for this. As you can see, there's a whole bunch of change in the spec I haven't gotten around to dealing with yet. But improvements of any scale are welcome.

PR #60, opened 2017-03-24
Put the correct feature

GuillaumeGomez:
It's me again. @alexcrichton said that the feature had to be this one and not a custom one. So here's the fix.


raphlinus:
Done, and published.

GuillaumeGomez:
Thanks a lot!

PR #62, opened 2017-03-28
Re-generate spec.rs tests

ScottAbbey:
It looks like spec.rs was modified by hand in
505b0a5510ad238c525c14c8af7280f60791d27d and possibly elsewhere.

Rerunning the build with `gen-tests` results in these changes to
spec.rs.

Making this change causes fewer tests to pass, but all the changes to
the tests appear to be correct -- they match the CommonMark spec
document now.

I will go through the CLA process if you approve this change.

### DIFF
--- a/tests/spec.rs
+++ b/tests/spec.rs
@@ -7,7 +7,8 @@ extern crate pulldown_cmark;
     #[test]
     fn spec_test_1() {
         let original = r##"	foo	baz		bim"##;
-        let expected = r##"<pre><code>foo	baz		bim</code></pre>
+        let expected = r##"<pre><code>foo	baz		bim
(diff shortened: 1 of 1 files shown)


raphlinus:
Yes, it's clearly wrong when the auto-gen stuff diverges from the input. One thing I think would help here is a strongly worded warning at the top of the auto-gen output not to edit it.

Thanks!

ScottAbbey:
Okay, I submitted the CLA form.

The top two lines of the spec.rs file do warn not to edit it, not sure what else can be done there.

raphlinus:
Right, I agree. I should have caught them being edited in the first PR :)

PR #63, opened 2017-03-30
Update testing procedures to pass trivially failing tests

ScottAbbey:
Hi Raph, I've sent 5 commits in this pull request but I don't think any of them depends on another, so if you aren't happy with one or more, they can be easily pulled out.  I wouldn't be surprised if I've overridden something that was done intentionally that I haven't yet picked up on.

I've included specifics in each commit message and won't repeat those here.  The basic goal is to update the testing procedures to pass about 58 tests that were trivially failing due to trailing whitespace in the expected compared to generated output.  Just calling `trim` before comparing would have only passed around 20 of the failing tests, a slightly different approach was needed that required modifying the tests generated by build.rs again.

The number of passing tests has increased from 506 to 564 after applying this set of commits.

These fixes will assist anyone aiming to correct individual test failures, because they have improved a large number of still-failing test cases that had multiple causes of failure.


raphlinus:
Thanks for working on this!

PR #66, opened 2017-04-03
Update spec.txt to jgm/CommonMark tagged 0.27

ScottAbbey:
I noticed there are 622 Examples on the [CommonMark 0.27 website](http://spec.commonmark.org/0.27) but only 621 in the test results for pulldown-cmark.

The spec.txt in this repo is not quite the same as the one published at the CommonMark site and jgm/CommonMark's 0.27 tag.  It's labeled 0.27 at the top, but one part has some changes.

There is one test case missing related to parentheses, and some parentheses rules have changed.

Is this just inadvertent error or something intentional?  Submitting this with a PR so you can see the changes easily.

### DIFF
--- a/tests/spec.rs
+++ b/tests/spec.rs
@@ -11437,7 +11437,49 @@ bar>)</p>
 
     #[test]
     fn spec_test_465() {
+        let original = r##"[link]((foo)and(bar))
+"##;
+        let expected = r##"<p><a href="(foo)and(bar)">link</a></p>
+"##;
+
+        use pulldown_cmark::{Parser, html, Options, OPTION_ENABLE_TABLES, OPTION_ENABLE_FOOTNOTES};
+
+        let mut s = String::new();
+
+        let mut opts = Options::empty();
+        opts.insert(OPTION_ENABLE_TABLES);
+        opts.insert(OPTION_ENABLE_FOOTNOTES);
+
+        let p = Parser::new_ext(&original, opts);
+        html::push_html(&mut s, p);
+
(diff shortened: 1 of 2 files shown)


raphlinus:
Almost certainly the spec changed since I imported it.

I'm happy to accept PR's to keep the spec sync'ed with upstream, but shouldn't the PR cover both the spec change and the autogen'ed file?

ScottAbbey:
I had just realized that. Updated.

raphlinus:
Thanks!

ScottAbbey:
@raphlinus Hmm, now I see I may have been overeager here. Apologies for my error.

It looks like this version is actually older than the one you had before, rather than newer.  This is the tagged 0.27 release, while the version I just overwrote was 0.27+some changes.

Should we be following jgm/CommonMark:master, or the latest tagged release?

Also, I see I should have updated the `Version:` tag on third_party/CommonMark/README.google to match the 0.27 tag's ID.

If you prefer to follow master, I will re-send this with current master's spec.txt. If you prefer to stick to tagged releases, I will send a new patch with the tagged commit's ID in the README.google file.

The latest version published and displayed on commonmark.org is the tagged 0.27 version (merged with this PR).

raphlinus:
I think as a general policy we should be following tagged releases.

I don't think it's necessary to keep the README.google up to date, the explicitly tagged version is primarily for licensing (ie, if the license doesn't change, I don't think it needs to update).

Thanks for looking into this!

PR #68, opened 2017-04-06
Remove extra backline

GuillaumeGomez:
So my point with this PR is trying to reduce the size of the generated HTML output by removing some useless characters, starting with '\n' between HTML tags.

### DIFF
--- a/src/escape.rs
+++ b/src/escape.rs
@@ -49,10 +49,10 @@ pub fn escape_href(ob: &mut String, s: &str) {
             match c {
                 b'&' => {
                     ob.push_str("&amp;");
-                },
+                }
                 b'\'' => {
                     ob.push_str("&#x27;");
-                },
+                }
                 _ => {
                     let mut buf = [0u8; 3];
                     buf[0] = b'%';
--- a/src/html.rs
+++ b/src/html.rs
@@ -59,7 +59,7 @@ impl<'a, 'b, I: Iterator<Item=Event<'a>>> Ctx<'b, I> {
                 Html(html) |
                 InlineHtml(html) => self.buf.push_str(&html),
                 SoftBreak => self.buf.push('\n'),
-                HardBreak => self.buf.push_str("<br />\n"),
+                HardBreak => self.buf.push_str("<br />"),
                 FootnoteReference(name) => {
                     let len = numbers.len() + 1;
                     self.buf.push_str("<sup class=\"footnote-reference\"><a href=\"#");
@@ -68,7 +68,7 @@ impl<'a, 'b, I: Iterator<Item=Event<'a>>> Ctx<'b, I> {
                     let number = numbers.entry(name).or_insert(len);
                     self.buf.push_str(&*format!("{}", number));
                     self.buf.push_str("</a></sup>");
-                },
+                }
             }
         }
     }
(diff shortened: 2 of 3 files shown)


ScottAbbey:
Here are some notes I thought up while checking out this PR.

1. Before this change, most (all?) output via the command line program ended with a newline.  Now it doesn't.  If I send a .md file that ends in a newline through the command line program, it outputs a .html file that does not end in a newline.   It seems like ending the output with a newline is probably correct through `push_html` as well, but maybe not.

2. This seems unexpected:

   [Example 291](http://spec.commonmark.org/0.27/#example-291) now outputs one line of HTML that is likely to be displayed by browsers as two lines:

   Input:
   ```
   foo\
   bar
   ```

   Output:
   ```
   <p>foo<br />bar</p>
   ```

   While the following outputs two lines of HTML that are likely to be displayed as one line in browsers:
   
   Input:
   ```
   foo
   bar
   ```

   Output:
   ```
   <p>foo
   bar</p>
   ```

3. This takes the number of passing tests via the `-s` option down from 566 to 17.  The jgm/CommonMark spec_tests.py script reports 568 passing tests both before and after this change, since it does do some normalization on the HTML output.  Since these are only superficial changes, I'm not saying this is an argument to not make this change, I just want to note that it is happening.  I have spent some time lately improving the built-in test procedures in order to enable and encourage contributions to fix some of the still-failing tests. 

   Adding a trailing newline to all output would get it back up to 435, but beyond that the internal tests will need to start doing more normalization if we want to keep them relevant.  Having them around is nice because they could be used to do some automatic testing when changes are submitted, but they aren't currently being used that way.

   I want to reiterate that this is a shortcoming of the testing procedure in this crate, not a complaint against this particular submission.  Ideally, the tests would be checking the internal representation rather than looking at one particular output format.

4. *Aesthetic notes only, not to be considered an argument for or against these changes*

   Before this change, at least all "ugly" output could be justified by saying it matched more or less exactly with the displayed output on the CommonMark spec.txt and webpage.  There was at least a certain amount of consistency -- opening tags that appeared on their own line tended to be closed by tags that appeared on their own line.  Now we have some rather asymmetrical results like Example 211, below.  HTML diffs tend to be more readable when there aren't multiple tags on the same line.

   The output for [Example 211](http://spec.commonmark.org/0.27/#example-211) is now:

   ```
   <blockquote>
   <blockquote>
   <blockquote>
   <p>foo
   bar</p></blockquote></blockquote></blockquote>
   ```

raphlinus:
Sorry to be negative (I really appreciate the work you've done integrating with rustdoc!), but this doesn't seem like something I'd want to merge. If I did, I'd need a good story around how the divergence from cmark output is handled. Further, where is the evidence that byte count in the generated HTML matters? If that's a fairly specialized concern, then is there some reason people can't use an HTML minifier tool?

GuillaumeGomez:
@ScottAbbey, @raphlinus: I think you're both right. The current HTML output actually bothers me because it's not supposed to be read by humans, therefore, no need to have all these backlines. However, we can completely debate and try to find a better solution, I'm all open to this possibility!

marcusklaas:
Closing this as the PR is now almost 2 years old and incompatible with the master branch.

PR #72, opened 2017-07-01
Add doc example for pulldown_cmark::html::push_html.

frewsxcv:


src/html.rs
```diff
@@ -245,6 +245,32 @@ impl<'a, 'b, I: Iterator<Item=Event<'a>>> Ctx<'b, I> {
...
+///
+/// * alpha
+/// * beta
+/// "#;
+/// let parser = Parser::new(markdown_str);
+///
+/// let mut html_buf = String::new();
+/// html::push_html(&mut html_buf, parser);
+///
+/// assert_eq!(html_buf, r#"<p>hello</p>
```
frewsxcv: Strange to me that this is a `<p>` tag and not `<h1>`

notriddle:
@raphlinus Are you going to look at this one?

src/html.rs
raphlinus: Yes, that seems like a bug. Has anyone looked into what's going on here?

frewsxcv: I haven't looked into it. I probably won't have much time in the coming days. Would you be okay shipping this PR as-is and opening a new issue for the `<h1>` thing?

raphlinus: Ok, I debugged this a bit. The problem is that the initial `# ` in `# hello` is being stripped by the doctest mechanism. That feels wrong to me, but I'm not excited about changing the behavior, as it likely means that tests will pass depending on the rust version.

Since this is not about header syntax per se, I feel the best resolution is just to remove the `#` from the markdown source. Then I'm happy to merge, and I'll cut a 0.1.0 version (satisfying #76 in the process).

frewsxcv: Good catch! Latest force push uses an alternate heading style

frewsxcv:
Also, at least for me, `master` currently has test failures. We should maybe look into getting CI turned on

raphlinus:
Thanks! Filed #79 to track the need to enable Travis, totally agree.

PR #74, opened 2017-07-12
Fix several spec tests

DemiMarie:
There are only 51 tests to go!  See commit messages for further information.


raphlinus:
Thanks! I think some of the other 51 tests will require a bit deeper work :). In any case, glad to have you working on this. It's something I'd like to do myself if I had more time, but I just don't right now.

PR #78, opened 2017-08-02
fix panic on incorrect code blocks

llogiq:
Due to an error in clippy, we wrongly parsed a `/**` comment as doc comment, leaving ` *` prefixes everywhere (see [issue #1920](https://github.com/rust-lang-nursery/rust-clippy/issues/1920). However, this trips up pulldown-cmark's scanner, which tries to get a byte from an empty slice.

This PR adds both a test case and a fix.


raphlinus:
Thanks, looks like a good fix!

llogiq:
You're welcome.

PR #21, opened 2016-04-10
End footnote definition with one blank line.

raphlinus:
According to the linked issue, a footnote definition should end with a
blank line. This is similar to the rule for lists, which end with two
blank lines. The code previously required two blank lines in both
cases, this patch changes it to just one for footnote definitions.

Fixes issue 20.

PR #84, opened 2017-08-09
Fix tests broken by footnote change

raphlinus:
The footnote change (issue #21) broke some tests. In some cases (just
the number of newlines) the tests needed to be updated, but in others
(allowing complex block structure) the code needed to be fixed.


raphlinus:
I'm pretty sure this is a good fix (verified that it doesn't change any of the spec tests), but I wanted to give other people a chance to take a look.

In the process of testing nested structure, I did change the behavior of one example:

```
* > a
  >  
  >
  > b
```

Previously it broke the list (two blank lines always broke a list), now it doesn't (the blockquote absorbs the two-blank-line behavior).

Looking at the spec, though, I think all this behavior has changed; I don't think two blank lines break a list anymore. That should probably be a separate change.

PR #85, opened 2017-08-23
Added Thematic Breaks to new_algo

petosegan:
I'm adding support for leaf blocks to the new algorithm, starting with Thematic Breaks

I also made parse_lines append text when it hits EOF. 

These changes increased the number of passing spec tests to 162/621.


petosegan:
OK, I fixed the commit author history to appease the bot.

PR #86, opened 2017-08-23
Handle Header Blocks for New Algorithm

petosegan:
The new algorithm now handles ATX Headers and Setext Headers. 

ATX Headers are handled in a similar way to the old algorithm.

Setext Headers are detected while parsing Paragraph nodes, and mutate the parent node to capture the paragraph text. Multiline Setext Headers are handled.

Spec tests for sections 4.1, 4.2, and 4.3 (Thematic Breaks, ATX Headers, and Setext Headers) are all passing, except those that rely on unimplemented blocks or escape characters. 

Now passing 189/621 spec tests.

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -245,10 +245,10 @@ impl<'a> RawParser<'a> {
     fn skip_blank_lines(&mut self) {
         loop {
             let ret = scan_blank_line(&self.text[self.off..]);
-            if ret == 0 {
+            if ret.is_none() {
                 break;
             }
-            self.off += ret;
+            self.off += ret.unwrap();
         }
     }
 
@@ -541,8 +541,8 @@ impl<'a> RawParser<'a> {
             }
         };
         let n = scan_blank_line(&self.text[off..]);
-        if n != 0 {
-            self.off = off + n;
+        if n.is_some() {
+            self.off = off + n.unwrap();
             return self.end();
         }
         self.state = State::TableRow;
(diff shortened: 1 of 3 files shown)


src/parse.rs
```diff
@@ -644,10 +644,10 @@ impl<'a> RawParser<'a> {
 
         if self.fence_char == b'\0' {
             let n = scan_blank_line(&self.text[off..]);
-            if n != 0 {
+            if n.is_some() {
```
raphlinus: As above.

src/parse.rs
```diff
@@ -541,8 +541,8 @@ impl<'a> RawParser<'a> {
             }
         };
         let n = scan_blank_line(&self.text[off..]);
-        if n != 0 {
-            self.off = off + n;
+        if n.is_some() {
```
raphlinus: Since you unwrap inside, I'd write `if let Some(n) = n` here.

src/prototype.rs
```diff
@@ -166,11 +167,65 @@ fn first_pass(s: &str) -> Tree<Item> {
...
+                tree.push();
+                let header_text_size = parse_line(&mut tree, s, ix);
+                ix += header_text_size;
+                tree.nodes[header_node_idx].item.end = ix;
+
+
+                // remove trailing matter from header text
+                let header_text = &s[header_start..];
+                let mut limit = ix - header_start;
+                while limit > 0 && header_text.as_bytes()[limit-1] == b' ' {
```
raphlinus: There are two alternatives to writing out these loops: [rposition](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.rposition) and [memchrr](https://docs.rs/memchr/1.0.1/memchr/fn.memrchr.html). In this case I'm not sure I'd change it, because the expected number of characters is small, but good to be aware of the choices.

raphlinus:
Great, thanks!

PR #89, opened 2017-09-05
Update spec tests & spec to commonmark 0.28

lukaramu:
Commonmark 0.28 released about a month ago; I auto-generated the spec tests with the new spec.


raphlinus:
Thanks, I saw this and was thinking about updating. Note, we'll also want to merge this into the new_algo branch, which thanks to @petosegan is seeing active development.

PR #90, opened 2017-09-06
Handle container blocks in new algorithm

petosegan:
I've implemented container blocks (blockquotes and lists) in the new algorithm. This ended up requiring many changes to leaf block handling as well. This now passes 363/621 spec tests.

The main loop (in first_pass) scans the text line-by-line. On each line, we first iterate over the spine of the tree to check that markings for containers are present. If not, we pop up the tree until all finished containers are closed. Then we scan for new containers and append them to the tree. Finally, we parse leaf blocks.

Code blocks and html blocks have been redone so that they are parsed one line at a time, because it simplifies the container logic.

Paragraphs are still parsed as multiple lines, because it seems like a straightforward way to implement lazy continuation of paragraph texts. 

Two key operations are handled as special cases during the second pass (tree parsing):
-   Loose list detection, and removal of trailing blank lines from indented code blocks. Loose lists are 
    detected with the help of non-html-generating BlankLine nodes. If a list is tight, we perform tree 
    surgery to excise paragraph tags.

-   Indented Code Block blankline removal is sort of split in two. The node containing the last non-blank node is recorded during tree construction, and everything after that is detached during tree parsing. Splitting it like this is favoring efficiency over readability of the code, and I'm not sure if it's the right trade-off.

I've also added a non-html-generating HTMLBlock node to parent raw html leaf blocks. This required a small change to html.rs, which I'm generally trying to avoid changing because it is client-facing. But I think it's clear that my change should not affect the correctness of HTML generation.

Most tests that should be passing are now passing. I'll note some exceptions that I found troublesome. The example numbers reference CommonMark v0.27:

**246**: The easy way to implement this is to make scan_paragraph_interrupt return false on blank list items. But this breaks example 244.

**265**: Here as well, it doesn't work to make scan_paragraph_interrupt return false on a numerical list item (with start_index != 1), which would seem like the easy fix. That would break example 263. John MacFarlane has expressed discomfort with this part of the spec (https://github.com/commonmark/cmark/issues/204).

**271 & 272**: I don't understand why the listitems with four spaces in front are legal... they seem to violate the definition.

**274**: This seems like it is in conflict with example 241. The blankline should terminate the list.

**284 & 285**: These point to a difficulty around handling blanklines at the end of lists. The blanklines should probably bubble up until they aren't the trailing node anymore. This has to be done during tree parsing since you don't know they are trailing until the first pass is over. I'm not sure yet how to do this correctly and efficiently.

I haven't implemented code fence info strings or HTML Type 7 yet. I'm not handling tabs correctly (which looks like it requires the synthesis of spaces in some cases). I haven't done anything about links. I believe the other failing tests before test 285 are due to parts of the inline handling that aren't implemented yet.

### DIFF
--- a/src/html.rs
+++ b/src/html.rs
@@ -178,6 +178,7 @@ impl<'a, 'b, I: Iterator<Item=Event<'a>>> Ctx<'b, I> {
                 self.buf.push_str(&*format!("{}", number));
                 self.buf.push_str("</sup>");
             }
+            Tag::HtmlBlock => {}
         }
     }
 
@@ -218,6 +219,7 @@ impl<'a, 'b, I: Iterator<Item=Event<'a>>> Ctx<'b, I> {
             Tag::Link(_, _) => self.buf.push_str("</a>"),
             Tag::Image(_, _) => (), // shouldn't happen, handled in start
             Tag::FootnoteDefinition(_) => self.buf.push_str("</div>\n"),
+            Tag::HtmlBlock => {}
         }
     }
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -171,7 +171,7 @@ fn run_spec(spec_text: &str, args: &[String], opts: Options) {
             print!(" ");
         }
 
-        let our_html = render_html(&test.input.replace("→", "\t").replace("\n", "\r\n"), opts);
+        let our_html = render_html(&test.input.replace("→", "\t"), opts);
 
         if our_html == test.expected.replace("→", "\t") {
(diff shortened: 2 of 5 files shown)


raphlinus:
Thanks for this, and also for the commentary about the limitations of the current state of parsing; that will help with future work. I didn't do a deep analysis but also didn't any find any problems in the quick pass over the code.

Fully correct handling of tabs will probably require some deeper thought, as this potentially touches many stages of the pipeline.

PR #91, opened 2017-09-08
Handle code fence info string and HTML type 7 on new algorithm

petosegan:
@elynnyap and I got code fence info strings and html type 7 working in the new algorithm. Now passing 370/621 spec tests.

We're a little unsure about the memory management of code info strings. Currently they are being cloned in item_to_tag.

### DIFF
--- a/src/prototype.rs
+++ b/src/prototype.rs
@@ -321,7 +321,7 @@ fn parse_html_line_type_1_to_5(tree : &mut Tree<Item>, s : &str, mut ix : usize,
     ix
 }
 
-fn parse_html_line_type_6(tree : &mut Tree<Item>, s : &str, mut ix : usize) -> usize {
+fn parse_html_line_type_6or7(tree : &mut Tree<Item>, s : &str, mut ix : usize) -> usize {
     let nextline_offset = scan_nextline(&s[ix..]);
     let htmlline_end_offset = scan_line_ending(&s[ix..]);
     tree.append_html_line(ix, ix+htmlline_end_offset);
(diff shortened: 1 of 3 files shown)


src/prototype.rs
```diff
@@ -633,10 +647,12 @@ fn parse_blocks(mut tree: &mut Tree<Item>, s: &str, mut ix: usize) -> usize {
 
     let (num_code_fence_chars, code_fence_char) = scan_code_fence(&s[ix..]);
     if num_code_fence_chars > 0 {
+        let nextline_offset = scan_nextline(&s[ix..]);
+        let info_string = s[ix+num_code_fence_chars..ix+nextline_offset].trim().to_string();
```
raphlinus: I _think_ this needs to be unescaped (this is done in the master branch), but I haven't double-checked with a test. (this is the kind of thing that arguably should be covered in the spec test, and if it isn't we should add one).

src/prototype.rs
```diff
@@ -104,7 +104,7 @@ enum ItemBody {
     Strong,
     Rule,
     Header(i32), // header level
-    FencedCodeBlock(usize, u8, usize), // number of fence chars, fence char, indentation
+    FencedCodeBlock(usize, u8, usize, String), // number of fence chars, fence char, indentation, info string
```
raphlinus: The quick answer to the lifetime question is that we _could_ make the enum `ItemBody<'a>` where `'a` is the lifetime of the source text, and make this a `Cow<'a, str>` instead. However, I'm not sure the performance gain is worth the extra hassle. I'm happy leaving it as is, and it's a potential optimization for later.

src/scanners.rs
```diff
@@ -93,6 +93,14 @@ pub fn is_ascii_alphanumeric(c: u8) -> bool {
     }
 }
 
+pub fn is_ascii_letterdigitdash(c: u8) -> bool {
```
raphlinus: Using match seems long-winded, I'd just write `c == b'-' || is_ascii_alphanumeric(c)`.

petosegan:
Made requested changes.

Escaping code info strings gave us two more passing tests.

raphlinus:
Awesome, thanks! Always nice to catch something in code inspection.

PR #92, opened 2017-09-18
Make generated cargo tests easier to work with

nelsonjchen:
While the main executable can also run tests, keeping the standard
`cargo test` function useful is also helpful.

* * *

Add end character to generated tests to work with substring only
matching of tests cases.

Regex test matching was removed in early 2015 in:

https://github.com/rust-lang/rust/pull/21458

You can only filter tests by substring. What if you wanted to run
`spec_test_6`? You would run `spec_test_64` too. Adding the character,
`_` to the end allows a workaround of specifying `spec_test_6_` as the
match to run if you wanted to only run `spec_test_6_`.

E.g. `cargo test spec_test_6_`

* * *

Tests are modified to generate tests with spec names in the tests

Previously, all tests were generated with `spec_test_#_` as their name,
even if they were not generated from `spec.txt`. The `tables.txt` spec
would also generate a `spec_test_#_`. This is an issue if you wanted to
test `spec_test_6_` and only wanted to run the one from `spec.txt` and
not `tables.txt`'s `spec_test_6_`. While you could build the test
executables and only run the specific test executable with the
`spec_test_#_` you want, it's kind of annoying.

This change derives the test names from the spec filename. For example,
`spec.txt` will generate `spec_test_6_` while `tables.txt` and
`footnotes.txt` will generate `tables_test_6_` and `footnotes_test_6_`
respectively. Future optional specifications that might be added will
utilize the same rule.

This allows testing of example 6 from the `footnotes.txt` spec by simply
doing:

`cargo test footnotes_test_6_`

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -54,6 +54,8 @@ fn generate_tests_from_spec() {
         let mut spec_rs = File::create(&rs_test_file)
                                .expect(&format!("Could not create {:?}", rs_test_file));
 
+        let spec_name = file_path.file_stem().unwrap().to_str().unwrap();
+
         let spec = Spec::new(&raw_spec);
         let mut n_tests = 0;
 
@@ -67,7 +69,7 @@ fn generate_tests_from_spec() {
                     r###"
 
     #[test]
-    fn spec_test_{i}() {{
+    fn {}_test_{i}_() {{
         let original = r##"{original}"##;
         let expected = r##"{expected}"##;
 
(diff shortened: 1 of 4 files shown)


nelsonjchen:
I was working on trying to fix Example 6 of the spec but I didn't notice the new-algo branch. I did do this though so maybe the effort wasn't a complete loss. 

The former is really handy if you want to attach a debugger. It would probably be more handy if I was working on `spec_test_64` and didn't want breakpoints to hit on `spec_test_6`!

nelsonjchen:
I just realized there was an `--exact` flag. 🤦

PR #93, opened 2017-09-18
Modify generated test names to include spec names

nelsonjchen:
Previously, all tests were generated with `spec_test_#` as their name,
even if they were not generated from `spec.txt`. The `tables.txt` spec
would also generate a `spec_test_#`. This is an issue if you wanted to
test `spec_test_6` and only wanted to run the one from `spec.txt` and
not `tables.txt`'s `spec_test_6` from `cargo test`. While you could
build the test executables and only run the specific test executable
with the `spec_test_#` you want, it's kind of annoying. Also, having
more specific names will make the output a little more clear as well in
a whole crate `cargo test`.

This change derives the test names from the spec filename. For example,
`spec.txt` will generate `spec_test_6` while `tables.txt` and
`footnotes.txt` will generate `tables_test_6` and `footnotes_test_6`
respectively. Future optional specifications added will utilize the same
rule.

This allows testing of example 6 from the `footnotes.txt` spec by simply
doing:

`cargo test footnotes_test_6`

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -54,6 +54,8 @@ fn generate_tests_from_spec() {
         let mut spec_rs = File::create(&rs_test_file)
                                .expect(&format!("Could not create {:?}", rs_test_file));
 
+        let spec_name = file_path.file_stem().unwrap().to_str().unwrap();
+
         let spec = Spec::new(&raw_spec);
         let mut n_tests = 0;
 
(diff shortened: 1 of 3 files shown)


raphlinus:
Great, that was annoying me too. Thanks for cleaning it up.

PR #94, opened 2017-10-17
Add support for inline HTML tags

notriddle:
Before: test result: FAILED. 372 passed; 249 failed; 0 ignored; 0 measured; 0 filtered out

After: test result: FAILED. 385 passed; 236 failed; 0 ignored; 0 measured; 0 filtered out


raphlinus:
Thanks, good to have you working on this.

PR #95, opened 2017-10-18
Consider backslash-escapes while parsing inline text

notriddle:
Before: test result: FAILED. 372 passed; 249 failed; 0 ignored; 0 measured; 0 filtered out

After: test result: FAILED. 386 passed; 235 failed; 0 ignored; 0 measured; 0 filtered out

n.b. This is completedly separate from the other PR.


raphlinus:
Thanks!

PR #97, opened 2017-10-18
Handle newlines and normalize the HTML

notriddle:
Brings the number of passing tests up to 409


raphlinus:
I chose not to normalize the HTML first time around, because it was relatively straightforward to match the cmark output byte-for-byte. Has that significantly changed? I haven't dug into this patch too carefully, but it's actually unexpected that the insignificant whitespace has changed, as that's mostly controlled by HTML generation and not parsing.

I'm totally fine with the general thrust, just want to make sure I understand what's going on with the whitespace. Probably that means I should dig into it myself, but I'm a bit busy the next couple of days...

notriddle:
There are two reasons why I think it's worthwhile to normalize HTML like this:

First of all, the official test harness does it. This tells me that cmark output doesn't match the specification byte-for-byte. If the reference implementation doesn't have to do it, then why do we have to do it?

2. The HTML produced by the code, as I have it written now, looks like this:

```
<!-- markdown -->
Example\
Code
<!-- html -->
Example
<br>Code
```

I could fix that, by doing some manual fiddling with the tree structure (it's the SynthesizeNewline item that's getting inserted into the wrong place) but it seems pointless.

notriddle:
@raphlinus Should I remove the HTML normalization stuff?

raphlinus:
Sorry for not responding earlier. I didn't know that cmark did normalization, that's a good argument. I'll merge this.

PR #98, opened 2017-10-19
Update bitflags to v1 and publish v0.2

chetgurevitch:
Breaking change, raises the minumum supported rust version to 1.20.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "pulldown-cmark"
-version = "0.1.0"
+version = "0.2.0"
 authors = [ "Raph Levien <raph@google.com>" ]
 license = "MIT"
 description = "A pull parser for CommonMark"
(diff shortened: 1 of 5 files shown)


raphlinus:
I'm all in favor of keeping deps up to date, but I'm not sure what the advantages / disadvantages of this move will be. Some people might want to compile with older Rust. Also, I don't _think_ incrementing the minimum language version is considered breaking by Rust semver conventions.

I might scoop this in when making other actually breaking changes.

ignatenkobrain:
@raphlinus main reason is that version is "1.0" which means it won't be braking anymore (until 2.0)

raphlinus:
Closing due to lack of CLA. Happy to accept a PR to upgrade bitflags. I don't think bumping to 0.2 is necessary, as upgrading bitflags shouldn't be considered a breaking change for semver purposes.

PR #99, opened 2017-10-19
Add some docs to some public tag variants with non-obvious fields

j3parker:
This is minor, but it wasn't immediately obvious to me which field would be which for `Image` and `Link`. I threw in a doc comment for `Header` as well .

This doesn't look that pretty when rendered at the moment but hopefully rustdoc improves that in the future :)


raphlinus:
Thanks!

PR #103, opened 2017-10-28
Support parsing without embedded HTML

maghoff:
This patch adds an option flag, `OPTION_DISABLE_HTML`, which configures the parser to avoid recognizing embedded HTML tags in the input. Thus, all would-be HTML in the input ends up escaped in the output: `<malicious-tag>` -> `&lt;malicious-tag&gt;`.

This is useful when accepting Markdown input from untrusted users as a simpler alternative to whitelisting of HTML.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -39,4 +39,4 @@ mod puncttable;
 mod utils;
 
 pub use passes::Parser;
-pub use parse::{Alignment, Event, Tag, Options, OPTION_ENABLE_TABLES, OPTION_ENABLE_FOOTNOTES};
+pub use parse::{Alignment, Event, Tag, Options, OPTION_ENABLE_TABLES, OPTION_ENABLE_FOOTNOTES, OPTION_DISABLE_HTML};
--- a/src/main.rs
+++ b/src/main.rs
@@ -25,7 +25,7 @@ extern crate getopts;
 extern crate pulldown_cmark;
 
 use pulldown_cmark::Parser;
-use pulldown_cmark::{Options, OPTION_ENABLE_TABLES, OPTION_ENABLE_FOOTNOTES};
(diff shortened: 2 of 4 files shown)


est31:
There is an alternative to this BTW, which is using an HTML sanitizer: https://github.com/notriddle/ammonia

It allows for the "fall back to HTML any time" feature of markdown while having similar security characteristics. Maybe the docs could point out that alternative?

maghoff:
Yes, I think sanitation is the most commonly used way to secure Markdown input. (I called it whitelisting in my above comment) As such, it could be reasonable to point it out in documentation.

For completeness, I'd like to mention a couple of advantages of disabling mixing of HTML into the Markdown that are not related to security:

 * Easier to keep tight control of how the generated HTML is structured (easier to style with CSS)
 * Easier on both the programmer and the CPU to tweak, for example to generate a table of contents
 * Easier to target other formats; in my case I want to generate a "plain text" output that is tweaked to work well for full text search indexing

These are not insurmountable obstacles when accepting HTML input, but hopefully I have conveyed some reasons why it is some times preferable to disable HTML input 🙂

marcusklaas:
Unfortunately this PR is no longer compatible with the current design of pulldown-cmark.

I think there's definitely merit in having such an option. If you are still interested in adding such functionality, it'd probably be good to try to adhere to an existing spec like [GFM's raw HTML extension](https://github.github.com/gfm/#disallowed-raw-html-extension-). Please feel free to open another PR if you are!

maghoff:
Thanks for the update. I will definitely be needing this going forward as well, but I won't have time to implement it anew for a while.

GFM's raw HTML extension is about filtering. My feature is about removing embedded HTML altogether.

marcusklaas:
I'm currently exploring options for [extensible rendering](https://github.com/raphlinus/pulldown-cmark/issues/116) and it looks like that should be able to cover HTML filtering as well. I'll keep you posted!

PR #113, opened 2017-11-13
Numbers refactor

RadicalZephyr:
Move the `numbers` piece of state from local to the `run` method into the `Ctx` struct with the other rendering state.

### DIFF
--- a/src/html.rs
+++ b/src/html.rs
@@ -42,7 +42,7 @@ struct Ctx<'b, I> {
     table_cell_index: usize,
 }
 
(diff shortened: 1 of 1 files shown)


RadicalZephyr:
I signed it!

RadicalZephyr:
Thanks for taking a look @raphlinus. I appreciate not wanting to merge a breaking change in to master, but I don't see how my proposed changes are breaking.  The only public functions in this module are `run` and `push_html` and I didn't change the signature of either function.  So I'm curious what you see as a breaking change.

PR #115, opened 2017-11-13
Extensible rendering 

RadicalZephyr:
This is a tracking/discussion PR for the changes that @reillysiemens and I proposed in #116.
The basic gist is that we want to make html rendering more extensible for library users.

[proposed changes]:

### DIFF
--- a/src/html.rs
+++ b/src/html.rs
@@ -29,9 +29,37 @@ use parse::Event::{Start, End, Text, Html, InlineHtml, SoftBreak, HardBreak, Foo
 use parse::Alignment;
 use escape::{escape_html, escape_href};
 
+pub trait IntoHtml<C> {
+    fn render(self, ctx: &mut C, buf: &mut String);
+}
+
 enum TableState {
     Head,
     Body,
+}
+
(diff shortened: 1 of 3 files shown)


raphlinus:
Sorry not to comment earlier.

Could you add a couple of sentences explaining what this is trying to do?

Commenting out the spec tests doesn't seem like a good idea, the file is autogenerated (and has a header at the top requesting it not be modified). We know these tests are failing. Part of the problem is a bit of uncertainty whether to patch the existing code (for which passing all tests is going to be challenging) or put development into the `new_algo` branch, which is far from done but has the potential to be better.

RadicalZephyr:
@raphlinus those commits marked "REVERT ME" aren't going to be in the final PR.  We commented them out as part of our dev process, because it's harder to tell if we've broken anything when there are a bunch of failing tests.

We only have a basic sketch so far, but the basic plan is to add an `IntoHtml` trait and rebuild the `Ctx` struct to make use of that trait to render html.  We're nowhere near done yet, but I thought I would posted the PR  to provide a place for discussion of our implementation as it evolves.

RadicalZephyr:
Closing because we're still flailing at this too much.  Sorry for the noise, but maybe we can discuss this idea in #116?

PR #118, opened 2017-12-11
PartialEq derives for Tag, Event, and Alignment

Michael-F-Bryan:
This adds a derive for `PartialEq` so you can compare two events/tags for equality.

In `mdbook` I want to be able to skip past some arbitrary tag (e.g. `while parser.next() != Some(Event::End(some_tag)) { /* ignore */ }`) and not being able to use `==` or `!=` makes this operation quite difficult.

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -79,7 +79,7 @@ pub struct ParseInfo<'a> {
     pub links: HashMap<String, (Cow<'a, str>, Cow<'a, str>)>,
 }
 
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, PartialEq)]
 pub enum Tag<'a> {
     // block-level tags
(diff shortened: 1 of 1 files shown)


Michael-F-Bryan:
@googlebot I signed it!

raphlinus:
Thanks, I can see how this is useful!

Michael-F-Bryan:
Thanks @raphlinus! Is there any chance you could do a patch release when you get a moment? 

I'm currently using the [patch] feature to override the published version of `pulldown-cmark` with the `master` branch, but once rust-lang-nursery/mdBook#491 is merged we'll need to update `Cargo.toml` to an official version before we can do another release of `mdbook`.

[patch]: http://doc.crates.io/manifest.html#the-patch-section

PR #123, opened 2018-01-23
Provide ability to provide a callback to be used in case of broken link references

### LINKED ISSUE #121: Option to insert reference link definitions on demand
It would be very helpful to be able to register a callback function that is called whenever the parser encounters a reference link it doesn't have a definition for.

If I'm not mistaken, it should suffice to add a new constructor `Parse::with_ref_link_callback(text, fn)` and call `fn`, if set, here instead of returning `None` directly:

https://github.com/google/pulldown-cmark/blob/c1ac9123f1653dc774403634696f848ffecbb01c/src/parse.rs#L1301

Please let me know if this is something you want to do in this project, and what your preferred API would look like.

## Context

@QuietMisdreavus and @Manishearth are currently working on implement [Rust RFC 1946](https://github.com/rust-lang/rfcs/blob/b9b71bbddd7ae54b338e3c17f47f30b9602cc338/text/1946-intra-rustdoc-links.md) "Intra rustdoc links" in <https://github.com/rust-lang/rust/pull/47046>.

The RFC's goal is to make it easy to link to Rust items in doc comments, and the main way to do that is by adding "implied ref links": You can write a doc comment containing

> This calls [foo()] with the given [::bar::Baz].

and rustdoc will generate the correct links to the function `foo` and the type `Baz`.

Rustdoc will soon use pulldown-cmark, and our goal is to land this feature once it does. Sadly, the current API of pulldown will return links without definitions as text events, so we can't dynamically generate and add these links. Adding all link definitions up front is not viable, as there may be links to external crates, which means the number of possible definitions is infinite.

Manishearth:
fixes #121

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -64,6 +64,10 @@ pub struct RawParser<'a> {
     containers: Vec<Container>,
     last_line_was_empty: bool,
 
+    /// In case we have a broken link/image reference, we can call this callback
+    /// with the reference name and use the link/title pair returned instead
+    broken_link_callback: Option<&'a Fn(&str) -> Option<(String, String)>>,
+
     // state for code fences
     fence_char: u8,
     fence_count: usize,
@@ -145,8 +149,10 @@ bitflags! {
 const MAX_LINK_NEST: usize = 10;
 
 impl<'a> RawParser<'a> {
-    pub fn new_with_links(text: &'a str, opts: Options,
-            links: HashMap<String, (Cow<'a, str>, Cow<'a, str>)>) -> RawParser<'a> {
+    pub fn new_with_links_and_callback(text: &'a str, opts: Options,
+            links: HashMap<String, (Cow<'a, str>, Cow<'a, str>)>,
+            callback: Option<&'a Fn(&str) -> Option<(String, String)>>)
+    -> RawParser<'a> {
         let mut ret = RawParser {
             text: text,
             off: if text.starts_with("\u{FEFF}") { 3 } else { 0 },
@@ -162,6 +168,8 @@ impl<'a> RawParser<'a> {
             fence_count: 0,
             fence_indent: 0,
 
+            broken_link_callback: callback,
+
             // info, used in second pass
             loose_lists: HashSet::new(),
             links: links,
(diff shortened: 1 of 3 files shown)


src/parse.rs
```diff
@@ -64,6 +64,10 @@ pub struct RawParser<'a> {
     containers: Vec<Container>,
     last_line_was_empty: bool,
 
+    /// In case we have a broken link/image reference, we can call this callback
```
raphlinus: I'd prefer the naming "link resolver"

src/passes.rs
```diff
@@ -33,7 +33,18 @@ impl<'a> Parser<'a> {
...
+    pub fn new_ext(text: &'a str, opts: Options) -> Parser<'a> {
+        Self::new_with_broken_link_callback(text, opts, None)
+    }
+
+    /// In case the parser encounters any potential links that have a broken
+    /// reference (e.g `[foo]` when there is no `[foo]: ` entry at the bottom)
+    /// the provided callback will be called with the reference name,
+    /// and the returned pair will be used as the link name and title if not
+    /// None.
+    pub fn new_with_broken_link_callback(text: &'a str, mut opts: Options,
```
raphlinus: I'm curious whether the callback might be a reference rather than a `Box`, as it's only used for the lifetime of the parser (similar to `text`).

Manishearth: It could, I didn't want to introduce another lifetime param. But reusing the same one may work.

Manishearth:
ping @raphlinus

raphlinus:
Sorry, I was waiting for my initial review to be addressed, and kinda dropped the ball. Do you think this is ready for merge?

Manishearth:
I think so

Manishearth:
Another thing we need is to make the HTML escaping stuff public, but I'll open a separate PR for that.

raphlinus:
Ok, I'll merge as is then. Thanks!

Manishearth:
@raphlinus can we get a version bump for this? https://github.com/rust-lang/rust/pull/48274 recently landed so I'd like to incorporate this into rustc

raphlinus:
@Manishearth Done. I can do another release when the other stuff has landed.

Manishearth:
Sounds good.  I'll have to wait for that since the feature as is can't be used.

Might want to actually do it in a single version bump (or yank)? The other PR is technically a breaking change, but only if there's a version between these two PRs.

PR #126, opened 2018-02-18
Pass non normalized reference down

Manishearth:


Manishearth:
We need both of these for this to work

raphlinus:
Ah ok, wasn't sure.

I'm going to do an 0.1.2 which is technically a semver violation, but I think ok as there are no users of the 0.1.1 API.

Manishearth:
You can just yank the version 😄

raphlinus:
Done.

PR #128, opened 2018-03-06
Inline code in table rows now render backslashes properly

Fierthraix:
Fixes https://github.com/rust-lang-nursery/mdBook/issues/637
Fixes https://github.com/google/pulldown-cmark/issues/127

I have tested with both rustdoc and and mdBook
`cargo test` still fails the same 56 spec tests.

## Changes

If you put inline code in a table cell using code backticks (<code>\`</code>) then *everything* within those backticks is treated the same way it is *outside* tables.

The main effect is that having a `|` in inline code now renders correctly.

### Examples

```markdown
|           Regex             |     Matches     |
|-----------------------------|-----------------|
|`foo-(1|[3-5])|bar-([39]|39)`|`foo-1`, `bar-39`|
```
#### Old Rendering

<table><thead><tr><td>           Regex             </td><td>     Matches     </td></tr></thead><tr><td>`foo-(1</td><td>[3-5])</td><td>bar-([39]</td><td>39)`</td><td><code>foo-1</code>, <code>bar-39</code></td></tr></table>

#### New Rendering

<table><thead><tr><td>           Regex             </td><td>     Matches     </td></tr></thead><tr><td><code>foo-(1|[3-5])|bar-([39]|39)</code></td><td><code>foo-1</code>, <code>bar-39</code></td></tr></table>

<hr />

### <code>\`</code> with an escape `\` before pipes

This one doesn't change (which is to be expected as `|` doesn't need an escape in inline code)

```markdown
|           Regex             |     Matches     |
|-----------------------------|-----------------|
|`foo-(1\|[3-5])\|bar-([39]\|39)`|`foo-1`, `bar-39`|
```
<table><thead><tr><td>           Regex             </td><td>     Matches     </td></tr></thead><tr><td><code>foo-(1\|[3-5])\|bar-([39]\|39)</code></td><td><code>foo-1</code>, <code>bar-39</code></td></tr></table>

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -935,9 +935,16 @@ impl<'a> RawParser<'a> {
             return self.end();
         }
         let mut n = 0;
+        let mut in_code = false;
         while i < limit {
             let c = bytes[i];
-            if c == b'\\' && i + 1 < limit && bytes[i + 1] == b'|' {
+            if c == b'`' {
+                in_code = !in_code;
+            }
+            if in_code {
+                i += 1;
+                continue;
+            } else if c == b'\\' && i + 1 < limit && bytes[i + 1] == b'|' {
                 i += 2;
                 continue;
             } else if c == b'|' {


ScottAbbey:
1. We're trying to implement CommonMark here
2. CommonMark still doesn't have tables
3. At jgm/CommonMark and commonmark.org, there have been some calls for GitHub Flavored Markdown's tables spec to be officially adopted by CommonMark
4. The tables spec here is very under-defined

I'd hope to see at least some of the following:

- This PR to update the tables spec with at least whatever this change affects
- This PR to conform to GFM Tables at least on this specific point, in anticipation of future adoption

I'm not saying we need to rewrite the entire tables implementation and spec here to match GFM, I'm just trying to say that creating intentional differences from the most likely official tables candidate seems very unwise.

That said, see [GFM tables #193](https://github.github.com/gfm/#example-193):

----
> Include a pipe in a cell’s content by escaping it, including inside other inline spans:
----
```
| f\|oo  |
| ------ |
| b `\|` az |
| b **\|** im |
```
----
```
<table>
<thead>
<tr>
<th>f|oo</th>
</tr>
</thead>
<tbody>
<tr>
<td>b <code>|</code> az</td>
</tr>
<tr>
<td>b <strong>|</strong> im</td>
</tr></tbody></table>
```
----

Both the examples you have given here disagree with GFM tables.

My suggestion:

- Add any Examples from GFM's table spec that specifically deal with pipes to specs/table.txt
- Implement the handling of pipes to pass those tests

Fierthraix:
@ScottAbbey 
The reason I initially didn't want to escape when in code backticks is because it changes the behaviour of what happens inside <code>\`</code>s if they're in a table cell. The second reason is you now have to pass that information for whether a code block needs to ignore backslashes inside it depending on context.

While it is true that my commit doesn't follow the GFM table spec, it is still better than the current state (see rust-lang-nursery/mdBook#637).

I will however add your suggestions and push them soon.

raphlinus:
I'm going to get back to this, but after making new_algo handle the current master CommonMark spec. Actually I'm much more interested in some tests, illustrating what's correct behavior, than I am in a code patch.

marcusklaas:
Closing this as the change is no longer compatible with the new master, and it also seems to be inconsistent with the GFM spec, as @ScottAbbey noted.

Issue #137, opened 2018-06-24
Upgrade of bitflags (0.9 → 1.0)


silwol:
In an attempt to package `pulldown-cmark` for Debian, I found out that it still uses `bitflags` 0.9, when a 1.0 version was already released.
Migrating the dependency to 1.0 is mostly a straight-forward process, prefixing enum values with their type name. A quick- and dirty experiment showed that it works properly, and the same set of tests seem to pass as it is currently the case.
`bitflags` made breaking changes in [version 1.0](https://github.com/bitflags/bitflags/blob/master/CHANGELOG.md#100), requiring at least `rustc` version 1.20.
Debian has the policy to package the latest version of each required crate, except if older versions are required by the dependency tree. Because packaging multiple versions of crates causes significant overhead, it should be avoided if possible (putting it through the NEW queue controlled by the Debian FTP masters because it's effectively a new package, and cleaning up orphaned crate packages later).
*If* you would accept a patch upgrading the `bitflags` dependency to 1.0, I would like to do the work, which implies raising the required `rustc` version for `pulldown-cmark` to 1.20. If you deliberately chose to stay on `bitflags` 0.9 for keeping compatibility with older `rustc` versions, then please close this bug with an according comment, and I will attempt to package `bitflags` 0.9 for debian.

PR #139, opened 2018-06-27
Upgrade bitflags dependency from 0.9 to 1.0

silwol:
Closes: #137


blackgear:
I think you should raise the semver of pulldown_cmark itself, since the parser options got a break change.

raphlinus:
Yes, this is a breaking change. I've been thinking of batching all the pending breaking changes together, to minimize churn for clients, but I don't have a lot of time to work on this now.

Issue #109, opened 2017-11-11
HTML should write to any writer, not just strings.


quadrupleslap:
It would be really useful, and the only thing stopping it is some `fresh_line` function.

droundy:
Should this be closed since #144 was merged?

quadrupleslap:
Yes. Thanks!

PR #144, opened 2018-08-13
Allow writing to any Write instance

ferristseng:
Adds a `write_html` function that can write to any thing that implements `Write`. Also, changes `push_html` to use the `write_html` method internally.

#109


ferristseng:
I signed it!

ferristseng:
Ok try again!

rbuckland:
:+1: to merge please.

ferristseng:
I think that should be fine. I can preserve the signature of `push_html` by unwrapping the result.

ferristseng:
Sorry it took so long, but this should be ready now. `push_html` should work the same as before now!

PR #147, opened 2018-09-13
Add LinkType to differenciate different link types

### LINKED ISSUE #141: Not all different link types are distinguishable
With the values about link provided by this crate there is no way to tell link references to existing definitions with and without text apart.

There are the following different link types with their respective values as provided by this crate. As you can see from the `dst`, `title` and `text` values, I can't tell 5 and 7, and 6 and 8 apart.

Would it be possible to either add a `tag` value to the `Tag::Link` variant holding the referenced tag, or create an enum containing the different link types which is held by `Tag::Link`? Also with such a change I think it would make sense to make the existing `title` and the future `tag` `Option<Cow>` and have the value `None` if they aren't specified instead of using the empty string to represent those.

1. Reference to non-existent definition without text:
```md
[Foo]
dst="foo", title="Foo", text="Foo"
```
2. Reference to non-existent definition with text:
```md
[Text][Foo]
dst="foo", title="Foo", text="Text"
```
3. Inline link without title:
```md
[Text](Foo)
dst="Foo", title="", text="Text"
```
4. Inline link with title:
```md
[Text](Foo "Title")
dst="Foo", title="Title", text="Text"
```
5. Reference to existing definition without text without title:
```md
[Foo]
[fOo]: Link
dst="Link", title="", text="Foo"
```
6. Reference to existing definition without text with title:
```md
[Foo]
[fOo]: Link "Title"
dst="Link", title="Title", text="Foo"
```
7. Reference to existing definition with text without title:
```md
[Text][Foo]
[fOo]: Link
dst="Link", title="", text="Text"
```
8. Reference to existing definition with text with title:
```md
[Text][Foo]
[fOo]: Link "Title"
dst="Link", title="Title", text="Text"
```

oberien:
Fix #141

### DIFF
--- a/src/html.rs
+++ b/src/html.rs
@@ -148,7 +148,7 @@ impl<'a, 'b, I: Iterator<Item=Event<'a>>> Ctx<'b, I> {
             Tag::Emphasis => self.buf.push_str("<em>"),
             Tag::Strong => self.buf.push_str("<strong>"),
             Tag::Code => self.buf.push_str("<code>"),
-            Tag::Link(dest, title) => {
+            Tag::Link(_, dest, title) => {
                 self.buf.push_str("<a href=\"");
                 escape_href(self.buf, &dest);
                 if !title.is_empty() {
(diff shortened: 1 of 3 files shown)


oberien:
Ok. Should I resolve the merge conflicts now or are you planning on adding more code that could result in further conflicts?

oberien:
I rebased the `linktype` branch on the current master.

PR #150, opened 2018-10-07
Update spec.txt

Keats:
Updates to the latest version of spec.txt from the CommonMark repo

### DIFF
--- a/third_party/CommonMark/spec.txt
+++ b/third_party/CommonMark/spec.txt
@@ -514,8 +514,8 @@ one block element does not affect the inline parsing of any other.
 ## Container blocks and leaf blocks
 
 We can divide blocks into two types:
-[container block](@)s,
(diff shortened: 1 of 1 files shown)


raphlinus:
A couple of things. First, my practice is generally to update to numbered releases, rather than tracking the repo, to make it easier to track differences. But I see upstream has gone well over a year without a release, so I might reconsider. Second, we need to update the tests as well, see #89 for an example.

The CLA is no longer necessary, I just deleted the hook.

Thanks for contributing!

raphlinus:
No worries. I don't have a very strong feeling either way. If you'd like to update, I'm ok with it; as we're fixing things it does make sense to go with the most updated spec.

Keats:
Looks like the tests themselves didn't change, only text clarifications were added.

PR #152, opened 2018-11-02
Prepare for 0.2 release

raphlinus:


raphlinus:
I'm putting this up as a PR largely to get feedback before doing the actual release. Since this is a breaking change (see #139), are there other changes people would like to see?

Do people want a community site to discuss this and other issues? I'd be happy to create a channel on [xi.zulipchat.com](https://xi.zulipchat.com).

raphlinus:
@polyzen Good suggestion. That one is complicated, as currently tests are failing because the draft cmark spec got updated and the code hasn't followed. I started the new_algo branch in an attempt to do a serious rework, but haven't followed through. I may or may not have time for it; I would be willing to mentor somebody to either fix the failing tests in the current codebase or push new_algo through.

In the meantime, I'm not sure how much value there would be in having CI when the tests are failing. I'm inclined not to block the 0.2 release on it.

polyzen:
Please update the lock file:
`error: the lock file needs to be updated but --locked was passed to prevent this`

raphlinus:
@polyzen I'm not able to repro; `cargo update` creates a null diff and I tried a couple other things. Can you provide more info; maybe you're using some optional features I'm not?

polyzen:
Actually I just assumed every project had a `Cargo.lock`. `--locked` is suggested for reproducible builds.

https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html

raphlinus:
I just ran `cargo build --locked` from this branch and it ran fine with no warnings or errors. So I'm trying to figure out why you're seeing an error and I'm not.

polyzen:
I'm guessing you have a `Cargo.lock` generated locally.

raphlinus:
Ohhhh. Right. So you're suggesting that the `Cargo.lock` be committed? I can do that.

polyzen:
Actually I'm not sure, lol:
https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries

raphlinus:
Well, it is a binary too, and since then there has been some movement to including Cargo.lock in libraries too, if for no other reason than to document the environment in which they were tested. I think I'll add it.

raphlinus:
I'll merge this and do a 0.2 release if I don't hear squeals of complaint soon. Also, I have a pull request to the new_algo branch (#153) to bring it up to date with this one. My thinking is that it's ultimately going to be an easier path getting to 100% spec compliance with the new algorithm than trying to patch this, so I'd like serious further development there.

PR #153, opened 2018-11-05
Merge branch 'release_0_2' into new_algo

raphlinus:
This patch brings the new_algo branch up to date with changes made up to the 0.2 release. Of course, the new_algo branch is still quite incomplete.


tests/footnotes.rs
```diff
@@ -3,40 +3,40 @@
...
-        opts.insert(OPTION_ENABLE_FOOTNOTES);
+        opts.insert(Options::ENABLE_TABLES);
+        opts.insert(Options::ENABLE_FOOTNOTES);
 
         let p = Parser::new_ext(&original, opts);
         html::push_html(&mut s, p);
 
-        assert_eq!(normalize_html(&expected), normalize_html(&s));
+        assert_eq!(expected, s);
     }
```
ScottAbbey: Normalization was added to `new_algo` in #97. Probably just need to re-run the gen-tests.

raphlinus: Right, I didn't re-gen, as that was broken, and fixed after this merge commit. I'm wondering what's the best process - get this merge in and fix things up after, or try to get master to a good state and redo this merge? Leaning towards the former because it seems like it blocks things least, but open to other ideas.

Issue #79, opened 2017-08-02
Enable Travis


raphlinus:
Some tests are failing in master, very likely because the footnote change didn't update the test. That's a strong demonstration of the need for CI.

PR #155, opened 2018-11-06
Update gen-tests feature after #139

ScottAbbey:
In #139, the spec tests were all updated with the new changes, but they
must have been changed with a global replace operation rather than an
update to the gen-tests build script.

These updates to the script result in the same generated tests after
running `cargo build --features gen-tests` so hopefully that's enough to
get things back in order.

This does not help more of the tests run, just helps keep them in order.

Ref: #79, some sort of automatic script running the gen-tests feature on
every update would probably have helped in this case.

Ref:
https://xi.zulipchat.com/#narrow/stream/149610-pulldown-cmark/subject/new-algo/near/146837971

> Raph Levien: The gen-tests feature may be broken, I
think the bitflags breaking change didn't update that.


raphlinus:
Thanks!

PR #156, opened 2018-11-07
Fix typo: incorrect apostrophe.

waywardmonkeys:


ScottAbbey:
> Do we have data on which form is most widely accepted. Can we bikeshed this some more maybe?

....challenge accepted? ( :-\ )

> But do not use apostrophes for plurals of abbreviations without periods, or for plurals formed from figures: TVs, PCs, DVDs; 1990s, 747s, size 7s.

So the two possible forms would be `ASTs` and `A.S.T.'s`.....some evidence is needed to establish whether `AST` or `A.S.T.` is more correct. Let's try using a book because those seem somewhat authoritative in establishing the language programmers use.

On pg. 227 of Engineering a Compiler, 2nd Ed., by Cooper & Torczon, the "abstract syntax tree (AST)" is defined. All references on that page use `AST` and none use `A.S.T.`.

waywardmonkeys:
I think this has been sufficiently bikeshedded! Can we merge this now? :)

PR #164, opened 2018-11-13
Scan attributes one line at a time

ScottAbbey:
*Not necessarily expecting this code to get merged, at least as-is, but it seems like the Pull Request makes easier to have a discussion with code examples attached than the Issue....*

I was checking on the recent progress on the `new_algo` branch and the [first failing test (number 60)](https://spec.commonmark.org/dingus/?text=%60Foo%0A----%0A%60%0A%0A%3Ca%20title%3D%22a%20lot%0A---%0Aof%20dashes%22%2F%3E) seemed worth taking a closer look at.  All the elements in it seemed like things that this branch should have covered already.

It was failing because the process of scanning inline HTML attributes in `scan_inline_html_open()` was spanning multiple lines without checking whether those lines affected block structure.

This particular change flips test number 60 from failing to passing and doesn't change the pass/fail of any other tests. I also tried [a few hand-crafted tests](https://spec.commonmark.org/dingus/?text=%3Ca%20href%3D%0A---%0A%22this%20is%20a%20href%22%0A%2F%3E%0A%0A%3Ca%20href%3D%22%0A---%0Athis%20is%20a%20href%22%0A%2F%3E%0A%0A%3Ca%20href%3D%0A%22this%20is%20a%20href%22%0A---%0A%2F%3E%0A%0A%3Ca%0Ahref%3D%22this%20is%20a%20href%22%0A---%0A%2F%3E%0A%0A%3Ca%0A---%0Ahref%3D%22this%20is%20a%20href%22%0A%2F%3E) that seemed to work.

...But I'm more interested in whether or not this particular phenomenon exists elsewhere and what we should do about it.  

So:

* Do you recall how these interactions were handled in the mainline branch?
* How committed are we to digging deeper into inline structures like this in the first pass?
* This change doesn't actually work the way it should when the examples are placed inside a block quote or list item...we probably need to shed the container marks (with `scan_containers()`?) on each line of these multi-line scans.

### DIFF
--- a/src/scanners.rs
+++ b/src/scanners.rs
@@ -872,15 +872,24 @@ pub fn scan_attribute_value(data: &str) -> Option<usize> {
     if size == 0 { return None; }
     let mut i = 0;
     match data.as_bytes()[0] {
-        b'\'' => {
+        quote @ b'\'' | quote @ b'"' => {
             i += 1;
-            i += scan_while(&data[i..], |c| c != b'\'');
-            i += 1;
-        },
-        b'"' => {
-            i += 1;
-            i += scan_while(&data[i..], |c| c != b'"');
-            i += 1;
+            // Block structure such as a setext heading must
+            // be determined before inlines such as HTML tags,
+            // so attributes must be processed one line at a time
+            loop {
+                if i >= data.len() { return None; }
+                if scan_setext_heading(&data[i..]).is_some() {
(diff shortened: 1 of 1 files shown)


ScottAbbey:
Sounds good! I think that is probably the best route to follow, it could get pretty messy otherwise.

PR #166, opened 2018-11-27
[new_algo] Handle HTML blocks

ScottAbbey:
This takes the passing tests count from 392 to 435.

The main action is basically rewriting the three HTML block situations to match the new patterns, mostly similar to the handling for indented code blocks.

It is passing all the tests specifically written for HTML blocks, but I can certainly imagine that I've missed something in the handling of HTML blocks that come inside other blocks. I tried a couple tests and things seemed to be working so it seems good enough for now.

I've intentionally avoided removing the now obsolete functions these new ones are replacing, they can probably all be yanked out in the future.

I've also avoided changing other return types and behaviors that might not be "Rust-y"...it seems like we're mostly pushing off optimization and reorganization until we've got a fully passing test suite, so I left all that alone in the interest of keeping the total size of these changes small.

Note that the `cargo run -- -s third_party/CommonMark/spec.txt`, which does not do normalization, starts to fail sooner than the `cargo test --test spec` now.  We probably don't want to add all the extra dependencies for normalization to the built binary though?

Let me know if you have any suggestions -- or just fix them yourself!

### DIFF
--- a/src/prototype.rs
+++ b/src/prototype.rs
@@ -159,8 +159,8 @@ impl<'a> FirstPass<'a> {
         let line_end_ix = nonspace_ix + scan_line_ending(&self.text[nonspace_ix..]);
         // Types 1-5 are all detected by one function and all end with the same
         // pattern
-        if let Some(_html_end_tag) = get_html_end_tag(&self.text[nonspace_ix..=line_end_ix]) {
-            // return after parsing type 1 to 5
+        if let Some(html_end_tag) = get_html_end_tag(&self.text[nonspace_ix..=line_end_ix]) {
+            return self.parse_html_block_type_1_to_5(ix, html_end_tag);
         }
 
         // Detect type 6
@@ -238,6 +238,49 @@ impl<'a> FirstPass<'a> {
 
         self.tree.pop();
         self.tree.nodes[self.tree.cur].item.end = ix;
+        ix
+    }
+
+    /// When start_ix is at the beginning of an HTML block of type 1 to 5,
+    /// this will find the end of the block, adding the block itself to the
+    /// tree and also keeping track of the lines of HTML within the block.
+    ///
+    /// The html_end_tag is the tag that must be found on a line to end the block.
+    fn parse_html_block_type_1_to_5(&mut self, start_ix: usize, html_end_tag: &'static str) -> usize {
+        self.tree.append(Item {
+            start: start_ix,
+            end: 0, // set later
+            body: ItemBody::HtmlBlock(Some(html_end_tag)),
+        });
+        self.tree.push();
+
+        let mut ix = start_ix;
+        let end_ix;
+        loop {
+            let line_start_ix = ix;
+            ix += scan_nextline(&self.text[ix..]);
+            self.append_html_line(line_start_ix, ix);
+
+            let mut line_start = LineStart::new(&self.text[ix..]);
+            let n_containers = self.scan_containers(&mut line_start);
+            if n_containers < self.tree.spine.len() {
+                end_ix = ix;
+                break;
+            }
+
+            if (&self.text[line_start_ix..ix]).contains(html_end_tag) {
+                end_ix = ix;
+                break;
+            }
+
+            let next_line_ix = ix + line_start.bytes_scanned();
+            if next_line_ix == self.text.len() {
+                end_ix = next_line_ix;
+                break;
+            }
+            ix = next_line_ix;
+        }
+        &self.pop(end_ix);
         ix
     }
 
@@ -349,6 +392,26 @@ impl<'a> FirstPass<'a> {
         }
     }
 
+
+    /// Appends a line of HTML to the tree.
+    fn append_html_line(&mut self, start: usize, end: usize)
+    {
+
+        if end > start {
+            self.tree.append(Item {
+                start: start,
+                // Normalize CRLF to LF
+                end: if self.text.as_bytes()[end - 2] == b'\r' { end - 2 } else { end - 1},
+                body: ItemBody::Html,
+            });
+        }
+        self.tree.append(Item {
+            start: end,
+            end: end,
+            body: ItemBody::SynthesizeNewLine,
+        });
+    }
+
     /// Returns number of containers scanned.
     fn scan_containers(&self, line_start: &mut LineStart) -> usize {
         let mut i = 0;
@@ -368,6 +431,7 @@ impl<'a> FirstPass<'a> {
                 ItemBody::Paragraph => (),
                 ItemBody::IndentCodeBlock(_) => (),
                 ItemBody::FencedCodeBlock(_) => (),
+                ItemBody::HtmlBlock(_) => (),
                 _ => panic!("unexpected node in tree"),
             }
             i += 1;
@@ -1066,7 +1130,7 @@ fn get_html_end_tag(text : &str) -> Option<&'static str> {
     static END_TAGS: &'static [&'static str; 3] = &["</script>", "</pre>", "</style>"];
 
     for (beg_tag, end_tag) in BEGIN_TAGS.iter().zip(END_TAGS.iter()) {
-        if 1 + beg_tag.len() < text.len() &&
+        if 1 + beg_tag.len() <= text.len() &&
             text.starts_with(&beg_tag[..]) {
             let pos = beg_tag.len();
             let s = text.as_bytes()[pos];
@@ -1078,7 +1142,7 @@ fn get_html_end_tag(text : &str) -> Option<&'static str> {
     static ST_BEGIN_TAGS: &'static [&'static str; 3] = &["<!--", "<?", "<![CDATA["];
     static ST_END_TAGS: &'static [&'static str; 3] = &["-->", "?>", "]]>"];
     for (beg_tag, end_tag) in ST_BEGIN_TAGS.iter().zip(ST_END_TAGS.iter()) {
-        if 1 + beg_tag.len() < text.len() &&
+        if 1 + beg_tag.len() <= text.len() &&
            text.starts_with(&beg_tag[..]) {
             return Some(end_tag);
         }
--- a/src/prototype.rs
+++ b/src/prototype.rs
@@ -166,12 +166,12 @@ impl<'a> FirstPass<'a> {
         // Detect type 6
         let possible_tag = scan_html_block_tag(&self.text[nonspace_ix..=line_end_ix]).1;
         if is_html_tag(possible_tag) {
-            // return after parsing type 6
+            return self.parse_html_block_type_6_or_7(ix);
         }
 
         // Detect type 7
         if let Some(_html_bytes) = scan_html_type_7(&self.text[nonspace_ix..=line_end_ix]) {
-            // return after parsing type 7
+            return self.parse_html_block_type_6_or_7(ix);
         }
 
         // Remaining blocks aren't impacted by leading spaces, so we can consume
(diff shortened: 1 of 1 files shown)


src/prototype.rs
```diff
@@ -144,7 +144,38 @@ impl<'a> FirstPass<'a> {
             return self.parse_indented_code_block(ix, remaining_space);
         }
 
+
+        // HTML Blocks
+
+        // HTML blocks may have 1-3 spaces in front of them, but our scanning
```
raphlinus: I suspect this isn't going to be quite right in the case where spaces can be created from a tab, i.e. `">\t<html>"`, which yields two spaces before the html tag. This is a very subtle problem, but maybe should be handled the way code blocks do it (this is what `remaining_space` is for).

src/prototype.rs
```diff
@@ -144,7 +144,38 @@ impl<'a> FirstPass<'a> {
...
+        // functions don't know that. The spaces are preserved in the output,
+        // so we don't want to move `ix` yet.
         line_start.scan_all_space();
+        let nonspace_ix = start_ix + line_start.bytes_scanned();
+
+        // HTML block openings always happen on a single line, but our scanning
```
raphlinus: The scanning functions should never consume past the end of a line, I think. The first pass parses the input into lines, and the resulting items in the chain are only within a line (with the partial exception of SoftBreak and HardBreak). Inline HTML scanning happens from an InlineScanner which only sees inside the items produced in the first pass. It's entirely possible there are cases this isn't strictly true, but those are probably bugs and maybe the scanners should be changed. In any case, this comment is likely to be confusing.

It is true the scanners only need to see to the end of the line, so I'm not particularly worried the code is wrong.

src/prototype.rs
```diff
@@ -144,7 +144,38 @@ impl<'a> FirstPass<'a> {
...
+        // functions are also used for inline HTML and may consume multiple lines,
+        // so we should only allow them to see one line at a time.
+        let line_end_ix = nonspace_ix + scan_line_ending(&self.text[nonspace_ix..]);
+        // Types 1-5 are all detected by one function and all end with the same
+        // pattern
+        if let Some(html_end_tag) = get_html_end_tag(&self.text[nonspace_ix..=line_end_ix]) {
```
raphlinus: Why ..= here? Does it need to actually see the newline character? This could panic if the string ends without a newline (so `line_end_ix == text.len()`).

src/prototype.rs
```diff
@@ -144,7 +144,38 @@ impl<'a> FirstPass<'a> {
             return self.parse_indented_code_block(ix, remaining_space);
         }
 
+
+        // HTML Blocks
+
+        // HTML blocks may have 1-3 spaces in front of them, but our scanning
+        // functions don't know that. The spaces are preserved in the output,
+        // so we don't want to move `ix` yet.
         line_start.scan_all_space();
```
raphlinus: I think the `scan_all_space` here is a no-op, any remaining space should have been consumed by the `scan_space_upto(4)` above. Maybe try removing it and verify nothing breaks?

src/prototype.rs
```diff
@@ -210,6 +241,86 @@ impl<'a> FirstPass<'a> {
...
+                end_ix = ix;
+                break;
+            }
+
+            if (&self.text[line_start_ix..ix]).contains(html_end_tag) {
+                end_ix = ix;
+                break;
+            }
+
+            let next_line_ix = ix + line_start.bytes_scanned();
```
raphlinus: I think you want to advance `ix` here. If last line doesn't end in a newline, then I think `ix` will be left at the _beginning_ of that line on return.

src/prototype.rs
```diff
@@ -210,6 +241,86 @@ impl<'a> FirstPass<'a> {
...
+            let mut line_start = LineStart::new(&self.text[ix..]);
+            let n_containers = self.scan_containers(&mut line_start);
+            if n_containers < self.tree.spine.len() || line_start.is_at_eol()
+            {
+                end_ix = ix;
+                break;
+            }
+
+            let next_line_ix = ix + line_start.bytes_scanned();
+            if next_line_ix == self.text.len() || scan_eol(&self.text[next_line_ix..]).1 {
```
raphlinus: `scan_eol` isn't exactly the detection of a blank line, it will fail to catch a line consisting of whitespace.

src/prototype.rs
```diff
@@ -318,6 +429,26 @@ impl<'a> FirstPass<'a> {
         }
     }
 
+
+    /// Appends a line of HTML to the tree.
+    fn append_html_line(&mut self, start: usize, end: usize)
```
raphlinus: Suggestion, this could take a remaining_space argument, as does the `append_code_text`.

src/prototype.rs
```diff
@@ -318,6 +429,26 @@ impl<'a> FirstPass<'a> {
...
+    {
+
+        if end > start {
+            self.tree.append(Item {
+                start: start,
+                // Normalize CRLF to LF
+                end: if self.text.as_bytes()[end - 2] == b'\r' { end - 2 } else { end - 1},
```
raphlinus: This looks like it will clip the last byte in the case where the line doesn't end in a newline.

src/prototype.rs
```diff
@@ -318,6 +429,26 @@ impl<'a> FirstPass<'a> {
...
+                body: ItemBody::Html,
+            });
+        }
+        self.tree.append(Item {
```
raphlinus: If there's a newline in the source, I prefer to output it as text rather than synthesizing a new line, for sourcemap purposes. I suspect we should probably synthesize a newline in the case where the input ends without a newline.

Possibly some of this logic can be combined with the code case, but I'm not too worried, a little duplication is fine.

src/prototype.rs
```diff
@@ -1035,7 +1167,7 @@ fn get_html_end_tag(text : &str) -> Option<&'static str> {
     static END_TAGS: &'static [&'static str; 3] = &["</script>", "</pre>", "</style>"];
 
     for (beg_tag, end_tag) in BEGIN_TAGS.iter().zip(END_TAGS.iter()) {
-        if 1 + beg_tag.len() < text.len() &&
+        if 1 + beg_tag.len() <= text.len() &&
```
raphlinus: I think this length check (and the one below) should be removed. According to the spec, the start condition is satisfied if the line is just the begin tag and a newline.

Also, spec says matching is case insensitive, so `starts_with` isn't quite right.

src/prototype.rs
ScottAbbey: I had copied this part from the `parse_indented_code_blocks()`. It does seem correct to advance it here. 

Do you recall why you didn't want to advance `ix` in the similar case there?

src/prototype.rs
ScottAbbey: :-$

I suppose `scan_blank_line()` would be more appropriate.

src/prototype.rs
ScottAbbey: I was just trying to mimic the existing `fn append_html_line()` (impl on `Tree`). I assumed it was done for a reason over there.  Looks like I should have been mimicking `append_code_text` instead!

src/prototype.rs
ScottAbbey: > spec says matching is case insensitive,

Something like `text[..beg_tag.len()].to_lowercase().starts_with()` ? In which case we do need the length check after all? Is there a better way to do that?

src/prototype.rs
raphlinus: I don't think `scan_attribute_value` is being used for inline HTML. It's all being done by `scan_inline_html` which in this case calls into `scan_inline_attribute`. Both of those use an `InlineScanner` struct, not the source text.

I think the goal here is that old-style scanners (ie not those using InlineScanner) can be used to scan block structure, and they should always stop at the end of a line. They should be separate from the inline case (there might be a bit of logic duplication, but I think there will be important subtle differences anyway). Does that make sense?

src/prototype.rs
raphlinus: I misunderstood the first time I read this, thought there was a line that was to be included. I think the code is ok here, but probably want to set `end_ix` here for sourcemap. Actually I'm not sure you need a separate `end_ix`, it's good enough to `self.pop(ix)` at the end. The reason the indented code blocks need that is that blank line is not considered part of the block (so `end_ix` points before the blank line and `ix` after).

src/prototype.rs
raphlinus: Yeah, `append_code_text` was done after the infrastructure was put in place for the fun line start handling with tabs.

src/prototype.rs
raphlinus: Feel free to leave this as a TODO. That code snippet is almost correct but can panic if the tag length slices a Unicode codepoint in half. It also allocates a String which is not great for performance. Doing this fully correctly in the Unicode case is tricky, but fortunately the actual logic only needs ASCII. This is much simpler, but I'm not sure that function exists in the Rust std lib.

Also, on consulting the spec, I don't think the condition is getting it exactly right; it should check that there's whitespace etc after the tag name, so for example will reject `<prepare>`.

ScottAbbey: Thanks for the suggestion, I tried something different in the latest update.

ScottAbbey:
I've updated this with several new commits, each targeting one of your notes. I added the commits on top of my existing submission rather than remixing the whole thing. Let me know if you want that changed.

I pressed the Resolve button on the notes I was trying to address. If there are any still open I probably didn't do anything about them or investigate more.

If you have additional suggestions I'll be glad to look at this more!

Still passes the same number of official tests as the original submission but I can see things are working much better with some ad-hoc tests that were not working properly beforehand.

src/prototype.rs
```diff
@@ -1174,21 +1174,31 @@ fn get_html_end_tag(text : &str) -> Option<&'static str> {
     static BEGIN_TAGS: &'static [&'static str; 3] = &["<script", "<pre", "<style"];
     static END_TAGS: &'static [&'static str; 3] = &["</script>", "</pre>", "</style>"];
 
-    for (beg_tag, end_tag) in BEGIN_TAGS.iter().zip(END_TAGS.iter()) {
-        if 1 + beg_tag.len() <= text.len() &&
-            text.starts_with(&beg_tag[..]) {
+    'type_1: for (beg_tag, end_tag) in BEGIN_TAGS.iter().zip(END_TAGS.iter()) {
+        if text.len() >= beg_tag.len() && text.starts_with("<") {
+            for (i, c) in beg_tag.as_bytes()[1..].iter().enumerate() {
+                if ! (&text.as_bytes()[i+1] == c || &text.as_bytes()[i+1] == &(c - 32)) {
```
ScottAbbey: So far I haven't seen any of the ways this isn't safe, but I just noticed there's a `utils::strcasecmp()` function....any preference?

src/prototype.rs
```diff
@@ -138,34 +138,35 @@ impl<'a> FirstPass<'a> {
...
 
 
         // HTML Blocks
 
-        // Our HTML scanning functions don't know what to do with leading spaces,
-        // so start them out at the first nonspace character
-        let nonspace_ix = start_ix + line_start.bytes_scanned();
+        // Start scanning at the first nonspace character, but remember how many
+        // spaces there are because HTML blocks need to preserve them.
+        let remaining_space = indent;
```
raphlinus: This will have, I think, the slight disadvantage of not including the leading spaces on the first line in the sourcemap. But it's such a tiny nit I wouldn't block on it.

src/prototype.rs
raphlinus: Not a strong preference. I do think the code would be improved if the case-insensitive comparison was split out into a different function, but using `strcasecmp` is also not trivial, as it takes `&str` and here we're working with bytes. If you were going to do that, I think the way to do it is scan the keyword, detecting the end condition (eol, whitespace, `>`) _first_, then strcasecmp if that passes. To my ears, that's slightly cleaner than this, but either way seems fine.

src/prototype.rs
```diff
@@ -169,10 +169,6 @@ impl<'a> FirstPass<'a> {
             return self.parse_html_block_type_6_or_7(ix, remaining_space);
         }
 
-        // Remaining blocks aren't impacted by leading spaces, so we can consume
```
raphlinus: If you want to fix the initial whitespace sourcemap issue, the way to do it is: leave this here, remove the `ix` advance above (before the html blocks), _save_ the line_start (by cloning) before the `let indent = line_start.scan_space_upto(4)`, and pass in the non-advanced `ix` and the remaining space from the saved line_start. (Alternatively, now that I think about it, you can save just the remaining_space).

As I say, this is a small issue, but I'm trying to think about the "right way" to do these kinds of things.

src/prototype.rs
ScottAbbey: Resolved this one for now with a `TODO` comment.

src/prototype.rs
ScottAbbey: Let's just make sure I haven't blundered here as well:

```
<pre></pre>

Text
```

Gives an `HtmlBlock(0, 12)` and I believe that means bytes 0-11 are in the block and byte 12 is not in the block? So the HtmlBlock contains from the opening `<` until the newline at the end of that line, but not the newline for the blank line.

Does that seem right?

PR #168, opened 2018-12-19
Start work on inline links

raphlinus:
It should be able to handle inline links and titles.

462/633 tests passed

### DIFF
--- a/src/prototype.rs
+++ b/src/prototype.rs
@@ -45,12 +45,18 @@ enum ItemBody {
     Inline(usize, bool, bool), // Perhaps this should be MaybeEmphasis?
     MaybeCode(usize),
     MaybeHtml,
+    MaybeLinkOpen,
+    MaybeLinkClose,
     Backslash,
 
     // These are inline items after resolution.
     Emphasis,
     Strong,
     Code,
+    InlineHtml,
+    // Link params: destination, title.
+    // TODO: get lifetime in type so this can be a cow
+    Link(String, String),
 
     Rule,
     Header(i32), // header level
@@ -59,7 +65,6 @@ enum ItemBody {
     SynthesizeNewLine,  // TODO: subsume under SynthesizeText, or delete
     HtmlBlock(Option<&'static str>), // end tag, or none for type 6
     Html,
-    InlineHtml,
     BlockQuote,
     List(bool, u8, Option<usize>), // is_tight, list character, list start index
     ListItem(usize), // indent level
@@ -242,7 +247,9 @@ impl<'a> FirstPass<'a> {
     /// tree and also keeping track of the lines of HTML within the block.
     ///
     /// The html_end_tag is the tag that must be found on a line to end the block.
-    fn parse_html_block_type_1_to_5(&mut self, start_ix: usize, html_end_tag: &'static str, mut remaining_space: usize) -> usize {
+    fn parse_html_block_type_1_to_5(&mut self, start_ix: usize, html_end_tag: &'static str,
+            mut remaining_space: usize) -> usize
+    {
         self.tree.append(Item {
             start: start_ix,
             end: 0, // set later
@@ -284,7 +291,9 @@ impl<'a> FirstPass<'a> {
     /// When start_ix is at the beginning of an HTML block of type 6 or 7,
     /// this will consume lines until there is a blank line and keep track of
     /// the HTML within the block.
-    fn parse_html_block_type_6_or_7(&mut self, start_ix: usize, mut remaining_space: usize) -> usize {
+    fn parse_html_block_type_6_or_7(&mut self, start_ix: usize, mut remaining_space: usize)
+        -> usize
+    {
         self.tree.append(Item {
             start: start_ix,
             end: 0, // set later
@@ -308,7 +317,9 @@ impl<'a> FirstPass<'a> {
             }
 
             let next_line_ix = ix + line_start.bytes_scanned();
-            if next_line_ix == self.text.len() || scan_blank_line(&self.text[next_line_ix..]).is_some() {
+            if next_line_ix == self.text.len()
+                || scan_blank_line(&self.text[next_line_ix..]).is_some()
+            {
                 end_ix = next_line_ix;
                 break;
             }
@@ -434,6 +445,7 @@ impl<'a> FirstPass<'a> {
             self.tree.append(Item {
                 start: start,
                 end: start,
+                // TODO: maybe this should synthesize to html rather than text?
                 body: ItemBody::SynthesizeText(Borrowed(&"   "[..remaining_space])),
             });
         }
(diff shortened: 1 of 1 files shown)


src/prototype.rs
```diff
@@ -1546,51 +1594,223 @@ fn make_code_span(tree: &mut Tree<Item>, s: &str, open: usize, close: usize) {
     }
 }
 
-/// Handle inline HTML and code spans.
+fn scan_link_destination_plain(scanner: &mut InlineScanner) -> Option<String> {
```
marcusklaas: Wouldn't it be nicer to implement these as methods on `InlineScanner`?

src/prototype.rs
```diff
@@ -1322,11 +1357,24 @@ impl<'a> InlineScanner<'a> {
 
     fn to_node_and_ix(&self) -> (usize, usize) {
         let mut cur = self.cur;
-        if self.tree.nodes[cur].item.end == self.ix {
+        if cur != NIL && self.tree.nodes[cur].item.end == self.ix {
```
marcusklaas: Why the extra condition?

src/prototype.rs
```diff
@@ -1322,11 +1357,24 @@ impl<'a> InlineScanner<'a> {
 
     fn to_node_and_ix(&self) -> (usize, usize) {
```
marcusklaas: What do these numbers represent?

src/prototype.rs
marcusklaas: Also, do these methods replace functions in `scanners.rs`?

src/prototype.rs
```diff
@@ -1546,51 +1594,223 @@ fn make_code_span(tree: &mut Tree<Item>, s: &str, open: usize, close: usize) {
...
-                // TODO: this logic isn't right if the replaced chain has
-                // tricky stuff (skipped containers, replaced nulls).
-                tree.nodes[cur].item.body = ItemBody::InlineHtml;
-                tree.nodes[cur].item.end = ix;
-                tree.nodes[cur].next = node;
-                cur = node;
-                if cur != NIL {
-                    tree.nodes[cur].item.start = ix;
+        match tree.nodes[cur].item.body {
+            ItemBody::MaybeHtml => {
```
marcusklaas: We should also handle autolinks in this block at some point right? Add a TODO?

src/prototype.rs
```diff
@@ -1546,51 +1594,223 @@ fn make_code_span(tree: &mut Tree<Item>, s: &str, open: usize, close: usize) {
...
+}
+
+/// Handle inline HTML, code spans, and links.
 ///
 /// This function handles both inline HTML and code spans, because they have
-/// the same precedence. (Maybe rename?)
-fn handle_inline_html(tree: &mut Tree<Item>, s: &str) {
+/// the same precedence. It also handles links, even though they have lower
+/// precedence, because the URL of links must not be processed.
+fn handle_inline_pass1(tree: &mut Tree<Item>, s: &str) {
```
marcusklaas: I feel like this functions and many other functions in this module really dig deep into the internals of `Tree`. Is there not a way to abstract this out to the tree module?

src/prototype.rs
```diff
@@ -1546,51 +1594,223 @@ fn make_code_span(tree: &mut Tree<Item>, s: &str, open: usize, close: usize) {
...
+                let maybe_html = {
+                    let next = tree.nodes[cur].next;
+                    let mut scanner = InlineScanner::new(tree, s, next);
+                    if scan_inline_html(&mut scanner) {
+                        Some(scanner.to_node_and_ix())
+                    } else {
+                        None
+                    }
+                };
+                if let Some((node, ix)) = maybe_html {
```
marcusklaas: Ugh! NLL make this so much nicer.

src/prototype.rs
```diff
@@ -1546,51 +1594,223 @@ fn make_code_span(tree: &mut Tree<Item>, s: &str, open: usize, close: usize) {
...
+        }
+    }
+    if !scanner.scan_ch(b')') {
+        return None;
+    }
+    Some((url, title))
+}
+
+// TODO: if it's just node, get rid of struct. But I think there probably will be
+// state related to normalization.
```
marcusklaas: What kind of normalization?

src/prototype.rs
```diff
@@ -45,12 +45,18 @@ enum ItemBody {
     Inline(usize, bool, bool), // Perhaps this should be MaybeEmphasis?
     MaybeCode(usize),
     MaybeHtml,
+    MaybeLinkOpen,
```
marcusklaas: Would it be nice to at some point have different enums for the pre- and post resolution stage?

src/prototype.rs
raphlinus: They do replace functions in `scanners.rs`. In general, I haven't been aggressive in deleting old code, but a major cleanup is due soon.

Regarding whether they're methods on `InlineScanner`, that's something of a style choice, but I think of `InlineScanner` as providing access to text that might be non-contiguous because it doesn't include block structure (eg `>` for blockquote), and a function like this as consuming it.

src/prototype.rs
raphlinus: Because it's quite possible for `cur` to be `NIL` here if the client has scanned all the way to the end. I'm pretty sure if you revert this you can see a panic on one of the tests.

src/prototype.rs
raphlinus: This is a good point, and a strong argument can be made to have at least type aliases. The first is a "node index", which is an offset into the `Vec` underlying the tree. The second is a byte offset in the source text.

src/prototype.rs
raphlinus: I'm using the spec tests as the main engine driving development. So we know we're going to need autolinks without having an explicit TODO. I do use TODO for things that might not be covered by the spec tests or temporary code that I know needs to be reworked.

src/prototype.rs
raphlinus: I'd like to do that, yes. Part of the reason this hasn't happened yet is that it hasn't been 100% clear exactly what access to the tree is needed.

src/prototype.rs
raphlinus: Indeed, and feel free to clean it up :)

src/prototype.rs
raphlinus: The [label matching algorithm](https://spec.commonmark.org/0.28/#matches). My thinking here is that we compute the normalized form of the link label at the time that it's scanned here, then refer to that normalized string later when determining matches. But maybe that can be done as a normalized-compare.

src/prototype.rs
raphlinus: I'm not sure that's really possible, when you're in the middle of processing you're going to have some of both. The order nodes are processed is somewhat chaotic because of the requirements of the spec, it's not like a simple tree traversal where you can take a tree of one type as input and produce a tree of another type.

PR #172, opened 2019-02-06
Housekeeping

marcusklaas:
* moves to 2018 edition
* initial use of NLL to simplify
* clippy run
* rustfmt run


raphlinus:
So before digging into this, lemme tell why it hasn't been done before. There are occasional changes to master that don't affect the core parsing algorithm, and we want to be able to pull those into this branch. Formatting changes like these basically make automatic pulls impossible. There are, I think, a couple things that haven't been pulled yet, like the ability to use any `Write` for html output, not just a `String`.

That said, I'm in favor of these changes, and do all of them in the other code bases I'm actively working on (except for clippy - I found mostly false positives and am not a fan of the annotation noise; I'd certainly be willing to consider it though). I'm just not 100% sure what order things should be done.

One intermediate step, just to suggest something: do the changes on `prototype.rs` only for now (which won't introduce merge conflicts), then plan on a major cleanup when we merge back to master. That might not work with actually switching edition, but my gut feeling is that the edition changes are mostly in the `use` declarations.

marcusklaas:
Sounds fair - I'll make the changes!

ScottAbbey:
> There are occasional changes to master that don't affect the core parsing algorithm, and we want to be able to pull those into this branch.

How trustworthy is rustfmt? Can we run it on `master`, `new_algo`, and on top of any in-progress PRs and end up with something that is consistent across all of them?

The number of open PRs is pretty small right now, I don't have anything I'm working on locally, and you two are the only others who have been sharing any new code changes recently. Could be as good a chance as ever.

> - moves to 2018 edition

Do we need to consider the impact of bumping the "minimum rust version" for this project? If 2018 edition requires 1.31, how far is that from what the minimum rust version already is? Is there a tradeoff function between how much it helps and the impact it will have? (Hopefully minimal?)

marcusklaas:
Rustfmt'ing this repo went pretty smoothly. Only needed two `#[rustfmt_skip]` attributes on large static arrays.

raphlinus:
Right, rustfmt is pretty safe. I'd be ok with rustfmt on master also, and that should take care of most of the merging concerns. The edition and nll changes are farther reaching.

Here's my thinking re minimum rust version. I'd prefer not to change it for the existing 0.2.0 release branch, but I'm fine with 2018 and nll for the `new_algo` work. People are going to have to do some work to upgrade to this, so snapping to a recent Rust version sounds fine.

marcusklaas:
Closed in favor of https://github.com/raphlinus/pulldown-cmark/pull/173.

PR #173, opened 2019-02-06
Implement autolinks on new_algo

marcusklaas:
Also moves to edition 2018. Brings passing tests to 480/633.

Replaces https://github.com/raphlinus/pulldown-cmark/pull/172.

### DIFF
--- a/src/prototype.rs
+++ b/src/prototype.rs
@@ -23,9 +23,9 @@
 use std::borrow::Cow;
 use std::borrow::Cow::Borrowed;
 
-use parse::{Event, Tag, Options};
-use scanners::*;
-use tree::{NIL, Node, Tree};
+use crate::parse::{Event, Tag, Options};
+use crate::scanners::*;
+use crate::tree::{NIL, Node, Tree};
 
 #[derive(Debug)]
 struct Item {
--- a/src/prototype.rs
+++ b/src/prototype.rs
@@ -1314,6 +1314,8 @@ impl<'a> InlineScanner<'a> {
         self.ix -= 1;
     }
 
+    // Consumes byte if it was next and returns true. Does
+    // nothing and returns false otherwise.
     fn scan_ch(&mut self, c: u8) -> bool {
         self.scan_if(|scanned| scanned == c)
     }
@@ -1323,8 +1325,9 @@ impl<'a> InlineScanner<'a> {
         self.scan_while(|scanned| scanned != c)
     }
 
-    fn scan_if<F>(&mut self, f: F) -> bool
-        where F: Fn(u8) -> bool
+    fn scan_if<F>(&mut self, mut f: F) -> bool
+    where
+        F: FnMut(u8) -> bool,
     {
         if let Some(c) = self.next() {
             if !f(c) {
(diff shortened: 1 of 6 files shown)


src/prototype.rs
```diff
@@ -1697,6 +1701,125 @@ fn scan_link_title(scanner: &mut InlineScanner) -> Option<String> {
...
+    Some(uri)
+}
+
+fn scan_email(scanner: &mut InlineScanner) -> Option<String> {
+    // using a regex library would be convenient, but doing it by hand is not too bad
+
+    // FIXME: don't use strings?
+    let mut uri: String = "mailto:".into();
+
+    while let Some(c) = scanner.next() {
```
raphlinus: Will this do the right thing if the scanner terminates early (without `@`)?

src/prototype.rs
```diff
@@ -1697,6 +1701,125 @@ fn scan_link_title(scanner: &mut InlineScanner) -> Option<String> {
...
+    };
+    scanner.unget();
+
+    Some(uri)
+}
+
+fn scan_email(scanner: &mut InlineScanner) -> Option<String> {
+    // using a regex library would be convenient, but doing it by hand is not too bad
+
+    // FIXME: don't use strings?
```
raphlinus: I think `String` is ok here. We can look at it in a performance pass later.

src/prototype.rs
```diff
@@ -1697,6 +1701,125 @@ fn scan_link_title(scanner: &mut InlineScanner) -> Option<String> {
...
+                _ => {
+                    scanner.unget();
+                    break;
+                }
+            }
+            fresh_label = false;
+        }
+
+        if uri.len() == label_start || uri.len() - label_start > 63
+            || uri.chars().last() == Some('-') {
```
raphlinus: Here, `chars().last()` will scan the entire string. More efficient to do `as_bytes`.

marcusklaas: Good catch!

src/prototype.rs
marcusklaas: Ok

src/prototype.rs
marcusklaas: No, it will not! Will fix

marcusklaas:
Updated!

src/prototype.rs
```diff
@@ -1766,25 +1765,22 @@ fn scan_uri(scanner: &mut InlineScanner) -> Option<String> {
...
-            }
-            _ => {
-                return None;
-            }
+            b'=' | b'?' | b'^' | b'_' | b'`' | b'{' | b'|' | b'}' | b'~' | b'-' => (),
+            _ => break,
         }
     }
 
+    if uri.as_bytes()[uri.len() - 1] != b'@' {
```
raphlinus: I think there *might* be a possibility `uri` is empty at this point, which would be a panic.

src/prototype.rs
```diff
@@ -1766,25 +1765,22 @@ fn scan_uri(scanner: &mut InlineScanner) -> Option<String> {
...
+        return None;
+    }
```
raphlinus: And I think this changes the logic, leaves the `@` at the end. I didn't test it, but might be worth checking.

src/prototype.rs
marcusklaas: no!

src/prototype.rs
marcusklaas: Not sure of what you mean by this.. shouldn't `@` be at the end at that point?

marcusklaas:
Sweet! Thanks for the review :-)

PR #174, opened 2019-02-06
Implement inline images

marcusklaas:
Raises the number of passing tests to 488/633. I think there's only 1 relevant test that currently isn't passing as it's particularly nasty:

```md
![[[foo](uri1)](uri2)](uri3)
```
which should produce
```html
<p><img src="uri3" alt="[foo](uri2)" /></p>
```

I haven't figured out a way to solve in a way that's not *too* appalling. Grateful for any pointers!

### DIFF
--- a/src/prototype.rs
+++ b/src/prototype.rs
@@ -47,6 +47,7 @@ enum ItemBody {
     MaybeHtml,
     MaybeLinkOpen,
     MaybeLinkClose,
+    MaybeImage,
     Backslash,
 
     // These are inline items after resolution.
@@ -55,8 +56,9 @@ enum ItemBody {
     Code,
     InlineHtml,
     // Link params: destination, title.
-    // TODO: get lifetime in type so this can be a cow
+    // TODO: get lifetime in type so these strings can be cows
     Link(String, String),
+    Image(String, String),
 
     Rule,
     Header(i32), // header level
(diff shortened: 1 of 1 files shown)


raphlinus:
Yeah, this nested stuff is going to be a bit tricky. I'm not sure I have much in the way of pointers, except to note that the logic for completing a link clears the link stack (`link_stack.clear()`) but the logic for an image shouldn't, because a link body can contain an image. I haven't dug into your code yet to see whether that's the problem.

marcusklaas:
Yeah that's exactly what I'm doing :sweat_smile:

marcusklaas:
Sounds good to me :-)

src/prototype.rs
```diff
@@ -766,11 +768,16 @@ fn parse_line(tree: &mut Tree<Item>, s: &str, mut ix: usize) -> (usize, Option<I
                 begin_text = ix;
             }
             b'[' => {
-                tree.append_text(begin_text, ix);
+                let (link_type, start) = if ix >= begin_text + 1 && bytes[ix - 1] == b'!' {
```
ScottAbbey: This is currently causing a panic on spec_test_573:

```
\![foo]

[foo]: /url "title"
```

[Published as #564](https://spec.commonmark.org/0.28/#example-564)

Looking backwards by bytes in this "safe" ASCII-in-UTF-8 environment isn't quite as straightforward as looking forward by bytes.  In this case, if there are an odd number of `\` before the `!`, it's a link, but if there are an even number of `\` before, it's an image.

marcusklaas: Thanks for the heads-up! I hadn't considered that. I will try to issue a fix :-)

PR #175, opened 2019-02-07
Make images work with escaped characters

marcusklaas:
Should fix the issue highlighted by @ScottAbbey [here](https://github.com/raphlinus/pulldown-cmark/pull/174#discussion_r254558886). That test just fails now instead of panicking.


marcusklaas:
:+1:

PR #176, opened 2019-02-07
Add initial implementation for reference definitions (WIP)

marcusklaas:
This isn't ready to merge yet since its implementation is pretty bad in some places (scanning is messy!) and I'm missing still many edge cases, but I was hoping to get feedback on the general design.

This pushes the number of passing tests to 508/633.


src/prototype.rs
```diff
@@ -82,16 +84,18 @@ struct FirstPass<'a> {
     text: &'a str,
     tree: Tree<Item>,
     last_line_blank: bool,
+    references: Vec<RefDefScan<'a>>,
```
raphlinus: I think this is a performance concern as the number of definitions grows. I think the win is to use [unicase](https://crates.io/crates/unicase) rather than caseless, so that the wrapped string can be used as a key in either a HashMap or a BTreeMap.

src/scanners.rs
```diff
@@ -456,6 +456,143 @@ pub fn calc_indent(text: &str, max: usize) -> (usize, usize) {
...
+    pub(crate) bytecount: usize,
+    pub(crate) label: &'a str,
+    pub(crate) dest: &'a str,
+    pub(crate) title: Option<&'a str>
+}
+
+pub fn scan_refdef(data: &str) -> Option<RefDefScan> {
+    let mut i = scan_whitespace_no_nl(data);
+
+    // defs can only be indented up to three spaces
```
raphlinus: It seems unlikely this will handle tabs correctly. I'm 90% certain you want to use the `LineStart` mechanism for initial indent, and assume all initial whitespace has been consumed by the time you get to the ref scanning.

src/scanners.rs
```diff
@@ -456,6 +456,143 @@ pub fn calc_indent(text: &str, max: usize) -> (usize, usize) {
...
+    if i > 3 {
+        return None;
+    }
+
+    // scan label
+    if scan_ch(&data[i..], b'[') == 0 {
+        return None;
+    }
+    i += 1;
+    let label_length = scan_while(&data[i..], |c| c != b']');
```
raphlinus: This doesn't seem to be taking into account bracket nesting. I think you'll have to write a little state machine, it'll probably need to do backslash processing too.

src/scanners.rs
```diff
@@ -456,6 +456,143 @@ pub fn calc_indent(text: &str, max: usize) -> (usize, usize) {
...
+    let label = &data[i..(i + label_length)];
+    i += label_length;
+    // FIXME: out of bounds possibly?
+    if scan_ch(&data[i..], b']') == 0 || scan_ch(&data[(i + 1)..], b':') == 0 {
+        return None;
+    }
+    i += 2;
+
+    // whitespace between label and url (including up to one newline)
```
raphlinus: If you process a newline, you'll need to do `scan_containers` at the beginning of the next line. Note that this means it needs to be a method of `FirstPass`, not a standalone function.

src/scanners.rs
```diff
@@ -456,6 +456,143 @@ pub fn calc_indent(text: &str, max: usize) -> (usize, usize) {
...
+    let mut newlines = 0;
+    i += scan_while(&data[i..], |c| {
+        if c == b'\n' {
+            newlines += 1;
```
raphlinus: *Probably* want to break early on `newlines > 1`, but that's a minor optimization.

src/scanners.rs
```diff
@@ -456,6 +456,143 @@ pub fn calc_indent(text: &str, max: usize) -> (usize, usize) {
...
+    }
+
+    // scan link dest
+    let (dest_length, dest) = scan_link_dest(&data[i..])?;
+    i += dest_length;
+
+    // scan whitespace between dest and label
+    newlines = 0;
+    let whitespace_bytes = scan_while(&data[i..], |c| {
+        if c == b'\n' {
```
raphlinus: Same re scan_containers on newline.

marcusklaas:
Thanks for the feedback Raph! I'll get to work :-)

marcusklaas:
558/633 now. Making progress!

Not yet ready for another review though..

marcusklaas:
Ok, I think this is now ready for a full review. It's currently at 563/633 with nearly all link/ image related tests passing. Most of the the ones that do not pass yet are due to escaping issues, which shouldn't be too hard to correct with more "cow magic".

Your most recent commit was a great source of inspiration and label comparisons are now implemented as `UniCase<Cow<'a, str>>`, which should be close to optimal indeed. I ended up not using the exact link label scanning code you recently pushed for two reasons:
  * the builder setup where it's possible to add multiple pieces of text seems unnecessary for current usecases where we don't ever need more than 1
  * I believe it wasn't invalidating the right labels, for example allowing those with only whitespace or nested brackets

There's one particular issue that I've struggled with and haven't found an elegant solution for: there's often two variants of scanning functions. One for `&str`s and one that works on `InlineScanner`s with obviously very similar logic. This seems like a source of clutter and duplication. Any ideas on how we could solve this in elegant manner would be much appreciated!

And of course I'll be very grateful for any other feedback you may have :-)

src/linklabel.rs
```diff
@@ -0,0 +1,90 @@
...
+        return None;
+    }
+    let mut i = 1;
+    let mut only_white_space = true;
+    let mut still_borrowed = true;
+    let bytes = text.as_bytes();
+    // no worries, doesnt allocate until we push things onto it
+    let mut label = String::new();
+
+    // TODO: loop over chars instead of bytes to be more utf-8 compatible?
```
raphlinus: I think so, see below. The way commonmark is designed, you almost never need to look at chars instead of bytes, pretty much everything non-ASCII is a pass-through. But this is probably the exception.

src/linklabel.rs
```diff
@@ -0,0 +1,90 @@
...
+    loop {
+        if i >= bytes.len() || i >= 1000 { return None; }
```
raphlinus: Here the spec calls for counting Unicode code points, not utf-8 code units.

src/linklabel.rs
```diff
@@ -0,0 +1,90 @@
...
+                }
+                i += whitespaces;
+            }
+            _ => {
+                only_white_space = false;
+                i += 1;
+            }
+        }
+
+        if !still_borrowed {
```
raphlinus: This can panic. There are basically two things you can do here: iterate characters, or keep track of a range of original string to push as a slice. *Generally* I've tended to do the latter, as pushing a slice is probably a bit more efficient than pushing individual codepoints, but we should probably focus on code clarity and correctness for now.

src/scanners.rs
```diff
@@ -964,38 +964,48 @@ pub fn is_html_tag(tag: &str) -> bool {
 pub fn scan_html_type_7(data: &str) -> Option<usize> {
     let mut i = scan_ch(data, b'<');
     if i == 0 {
-        return None; }
-    i += scan_ch(&data[i..], b'/');
+        return None;
+    }
+
+    let close_tag_bytes = scan_ch(&data[i..], b'/');
```
raphlinus: Why am I seeing this diff? It doesn't seem related to link defs. Maybe some branch confusion?

marcusklaas: You're absolutely right. It fixes a test unrelated to links. I will make a separate PR for this.

src/linklabel.rs
marcusklaas: Good catch!

marcusklaas:
599/633 now. itshappening.gif

marcusklaas:
Addressed all your comments. Should be in better shape now.

PR #178, opened 2019-02-11
Various emphasis fixes

marcusklaas:
This fixes 25 emphasis related tests.

### DIFF
--- a/src/prototype.rs
+++ b/src/prototype.rs
@@ -688,12 +688,30 @@ fn is_left_flanking(bytes: &[u8], run_len: usize, ix: usize) -> bool {
     if is_ascii_whitespace(next_byte) {
         return false;
     }
-    if !is_ascii_punctuation(next_byte) {
-        return true;
+    let delim = bytes[ix];
+    if delim == b'*' {
+        if !is_ascii_punctuation(next_byte) {
+            return true;
+        }
+        if ix == 0 {
+            return false;
+        }
+    } else {
+        // delim must be b'_'
+        if !is_ascii_punctuation(next_byte) {
+            if ix == 0 {
+                return true;
+            }
+
+            // this mess is to prevent intra word emphasis with _
+            // TODO: function name no longer representative
+            let prev_byte = bytes[ix - 1];
+            if !(is_ascii_whitespace(prev_byte) || is_ascii_punctuation(prev_byte)) {
+                return false;
+            }
+        }
     }
-    if ix == 0 {
-        return false;
-    }
+
     // FIXME: we really should get the prev char!
     let prev_byte = bytes[ix - 1];
 
--- a/src/prototype.rs
+++ b/src/prototype.rs
@@ -678,53 +678,54 @@ fn dump_tree(nodes: &Vec<Node<Item>>, mut ix: usize, level: usize) {
 }
 
 /// Determines whether the delimiter run starting at given index is
-/// left-flanking, as defined by the commonmark spec (and isn't introword
-/// for _ delims)
-fn delim_run_can_open(bytes: &[u8], run_len: usize, ix: usize) -> bool {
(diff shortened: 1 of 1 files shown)


src/prototype.rs
```diff
@@ -675,6 +677,56 @@ fn dump_tree(nodes: &Vec<Node<Item>>, mut ix: usize, level: usize) {
     }
 }
 
+/// Determines whether the delimiter run starting at given index is
+/// left-flanking, as defined by the commonmark spec (and isn't introword
```
raphlinus: "intraword"

src/prototype.rs
```diff
@@ -675,6 +677,56 @@ fn dump_tree(nodes: &Vec<Node<Item>>, mut ix: usize, level: usize) {
...
+    }
+    if ix == 0 {
+        return true;
+    }
+    let delim = bytes[ix];
+    if delim == b'*' && !is_ascii_punctuation(next_byte) {
+        return true;
+    }
+
+    // FIXME: we really should get the prev char!
```
raphlinus: Yup. I'm not sure off the top of my head whether there's code for this in pulldown-cmark. I know I've written it but it might be in xi :)

That said, if this function is changed to take `&str` rather than `&[u8]`, then you can use the stdlib and do it with `s[..ix].chars().rev().next()`. Similarly for next_char (don't use `rev`).

Expanding to Unicode whitespace here can be filed as an issue if it's not properly tested in the spec, and not block this PR.

src/prototype.rs
```diff
@@ -729,20 +781,29 @@ fn parse_line(tree: &mut Tree<Item>, s: &str, mut ix: usize) -> (usize, Option<I
...
-                let can_open = ix + count < s.len() && !is_ascii_whitespace(bytes[ix + count]);
-                let can_close = ix > start && !is_ascii_whitespace(bytes[ix - 1]);
-                // TODO: can skip if neither can_open nor can_close
-                for i in 0..count {
-                    tree.append(Item {
-                        start: ix + i,
-                        end: ix + i + 1,
-                        body: ItemBody::Inline(count - i, can_open, can_close),
-                    });
+                // FIXME: it looks like we're mixing byte indices with char indices here.
```
raphlinus: I don't think there's a fixme here, there are no char indices here, only byte indices. You're correct there's no panic risk because the delim run is guaranteed to be ASCII.

marcusklaas: The delim run is ASCII but any characters before it may not be, right? And then we wouldn't reach the delim run from the start of the string.

src/prototype.rs
marcusklaas: It is tested in the spec and we can get 1 or 2 more passes by fixing it :-) I'll try to do it in this pr.

src/prototype.rs
raphlinus: I'm not understanding your concern here, could you spell it out maybe?

marcusklaas: Yes! Say `s = "ö*"`, then by the time we reach line of the FIXME, the byte counter `ix` will be `2`, and we will panic on `&s[(ix + 1)..]` I think.

raphlinus: I'm still not seeing a problem, and I also tested with some Unicode strings. That string is `[0xc3, 0xb6, 0x2a]`, ix is 2 as you say, so ix + 1 is 3 and s[3..] is the empty string. This seems to me as it should be.

marcusklaas: Aaaah - I completely misunderstood how string indexing works. I thought it was character based! Then the behaviour here is fine indeed. I shall remove the FIXME. Thanks for the clarification!

marcusklaas:
Implemented all your suggestions. We now pass an additional test :-)

Thanks once again for your guidance!

PR #181, opened 2019-02-15
Implement footnotes for new_algo

marcusklaas:
Passes all 8 tests.


src/linklabel.rs
```diff
@@ -35,6 +40,15 @@ pub(crate) fn scan_link_label<'a>(text: &'a str) -> Option<(usize, LinkLabel<'a>
     let mut byte_index = 1;
     // no worries, doesnt allocate until we push things onto it
     let mut label = String::new();
+    let is_footnote = if let Some(&'^') = char_iter.peek() {
```
raphlinus: Footnotes are optional (they will cause divergence with the CommonMark spec), so I'd really like to see the option plumbed through.

marcusklaas:
Footnotes are feature gated now and I split the parsing of footnotes/ refdefs so it will bail early when looking for one but finding the other.

PR #183, opened 2019-02-16
Encoding fixes

marcusklaas:
This fixes 6 encoding/ escape tests, bringing the score to 622/633. There will be another PR fixing up to 8 more, mostly on href encoding, which this pull request doesn't touch.

### DIFF
--- a/src/escape.rs
+++ b/src/escape.rs
@@ -38,10 +38,11 @@ static HREF_SAFE: [u8; 128] = [
 
 static HEX_CHARS: &'static [u8] = b"0123456789ABCDEF";
 
-pub fn escape_href(ob: &mut String, s: &str) {
+pub fn escape_href(ob: &mut String, s: &str, backslash_escapes: bool) {
     let mut mark = 0;
+    let bytes = s.as_bytes();
     for i in 0..s.len() {
-        let c = s.as_bytes()[i];
+        let c = bytes[i];
         if c >= 0x80 || HREF_SAFE[c as usize] == 0 {
             // character needing escape
 
@@ -53,6 +54,9 @@ pub fn escape_href(ob: &mut String, s: &str) {
                 b'&' => {
                     ob.push_str("&amp;");
                 },
+                b'\\' if backslash_escapes && i + 1 < s.len() && is_ascii_punctuation(bytes[i + 1]) => {
+                    // skip \
+                }
                 b'\'' => {
                     ob.push_str("&#x27;");
                 },
--- a/src/escape.rs
+++ b/src/escape.rs
@@ -100,12 +100,20 @@ pub fn escape_html(ob: &mut String, s: &str, secure: bool) {
     let bytes = s.as_bytes();
     let mut mark = 0;
     let mut i = 0;
+
     while i < size {
         match bytes[i..].iter().position(|&c| HTML_ESCAPE_TABLE[c as usize] != 0) {
             Some(pos) => {
                 i += pos;
             }
             None => break
+        }
+        if let Some((bytes, c)) = parse_unicode_lit(&bytes[i..]) {
+            ob.push_str(&s[mark..i]);
+            ob.push(c);
+            i += bytes;
+            mark = i;
+            continue;
         }
         let c = bytes[i];
         let escape = HTML_ESCAPE_TABLE[c as usize];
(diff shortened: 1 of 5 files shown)


marcusklaas:
Ok, decided to do the rest of the encoding fixes in this PR as well. This should take of all of them. 629/633 tests are passing now.

There's some differences in backslash escaping behaviour between autolinks and other links, so in order to avoid unnecessary allocations by escaping at parse time, so extra information needs to be added to the AST/ events. Currently, this is solved by adding bools to the `Text` and `Link` tags, but it may be more elegant to just add a `Autolink` variant to the events.

marcusklaas:
Yisss, was able to squash the final encoding bug as well. 630/633.

raphlinus:
630 is a very sweet number. I've got a busy schedule so will try to get to the backlog of PR's soon, but can't promise a specific time.

src/escape.rs
```diff
@@ -35,10 +38,12 @@ static HREF_SAFE: [u8; 128] = [
 
 static HEX_CHARS: &'static [u8] = b"0123456789ABCDEF";
 
-pub fn escape_href(ob: &mut String, s: &str) {
+pub fn escape_href(ob: &mut String, s: &str, backslash_escapes: bool) {
```
raphlinus: I suspect that this is the wrong place to be special-casing on backslash escaping, that it should have been caught upstream. One way to tell for sure is to look at the AST generated by the [CommonMark testing tool](https://spec.commonmark.org/dingus/).

src/escape.rs
```diff
@@ -86,30 +101,60 @@ static HTML_ESCAPE_TABLE: [u8; 256] = [
...
+
     while i < size {
         match bytes[i..].iter().position(|&c| HTML_ESCAPE_TABLE[c as usize] != 0) {
             Some(pos) => {
                 i += pos;
             }
             None => break
         }
-        let c = bytes[i];
+        if !skip_unescape {
```
raphlinus: The input `&lt;` should be in the AST as `Text("<")`. I'm about 90% sure there doesn't need to be special case escaping here for code blocks. Of course, when *parsing* the rules are different - if the input `&lt;` is in a code block, then the AST holds `Text("&lt;")`.

src/escape.rs
```diff
@@ -118,3 +163,73 @@ pub fn escape_html(ob: &mut String, s: &str, secure: bool) {
     }
     ob.push_str(&s[mark..]);
 }
+
+fn parse_entity(s: &str) -> Option<(usize, &'static str)> {
```
raphlinus: About 90% sure these don't belong here, entity and literal parsing should be done during parsing.

src/html.rs
```diff
@@ -40,6 +40,9 @@ struct Ctx<'b, I> {
     table_state: TableState,
     table_alignments: Vec<Alignment>,
     table_cell_index: usize,
+    // number of code blocks we're in. this may never become
```
raphlinus: From what I can tell, changes won't be needed in this file, the unescaping logic should be done during parsing.

src/parse.rs
```diff
@@ -114,8 +114,8 @@ pub enum Tag<'a> {
     Strong,
     Code,
 
-    /// A link. The first field is the destination URL, the second is a title
-    Link(Cow<'a, str>, Cow<'a, str>),
+    /// A link. The first field is the destination URL, the second is a title, third is autolink indicator
```
raphlinus: I'm not convinced the rendering is different for autolinks; I strongly suspect that the differences are in parsing. I haven't dug into this too deeply though.

src/parse.rs
```diff
@@ -125,7 +125,7 @@ pub enum Tag<'a> {
 pub enum Event<'a> {
     Start(Tag<'a>),
     End(Tag<'a>),
-    Text(Cow<'a, str>),
+    Text(Cow<'a, str>, bool), // bool = true -> block backslash escapes
```
raphlinus: I'd need to be convinced of the need for this.

marcusklaas:
Thanks for the review Raph! Big agree that the code needs work. However, there is an idea behind deferring all the escape work/ differentiation between autolinks and regular links etc until we get to the html. This way, we avoid any intermediate allocations. If instead we replace things like `&#92;` or non-autolink urls at parse time, that's an extra allocation and an extra copy of the entire enclosing text that could be skipped. I think we can clean up the code so that we get both the speed of deferred escapes and convenience/ safety of parse time escapes, for example by providing "escaped text" getters on the events that do allocate, may consumers need it.

raphlinus:
I would focus on correctness and clarity first, and the optimization of deferring escaping to later, but I understand the desire.

marcusklaas:
Yeah, the escape bit seems like a good idea to me as well. I'm closing this pr and starting anew from the current `new_algo` so there's not a bunch of unnecessary commits (and I don't have to solve the merge conflicts! :sweat_smile:)

PR #185, opened 2019-02-25
Encoding fixes take two

marcusklaas:
It took some convincing, but I now also see that doing things like this makes more sense ^^. The code should be much cleaner now. Upgrading to max performance should be a relatively small change from here, should we opt for it.

This PR additionally moves all remaining `ItemBody` variants to `Cow<str>`s, although some parsing functions may still only return `Owned` variants. I left this for now.

Again at 630/633.

### DIFF
--- a/src/prototype.rs
+++ b/src/prototype.rs
@@ -1003,6 +1003,10 @@ fn delim_run_can_close(s: &str, suffix: &str, run_len: usize, ix: usize) -> bool
     next_char.is_whitespace() || next_char.is_ascii_punctuation()
 }
 
+// TODO: we could have a different ItemBody for this kind of text, 
+// producing a different node which makes getting either raw or unescaped
+// text explicit. This would eliminate an allocation/ copy when parsing
+// to HTML for example.
 fn unescape_and_append<'a>(tree: &mut Tree<Item<'a>>, text: &'a str, start: usize, end: usize) {
     if end > start {
         match unescape(&text[start..end]) {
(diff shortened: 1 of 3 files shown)


src/prototype.rs
```diff
@@ -995,10 +1035,11 @@ fn parse_line<'a>(tree: &mut Tree<Item<'a>>, s: &'a str, mut ix: usize) -> (usiz
         match bytes[ix] {
             b'\n' | b'\r' => {
                 let mut i = ix;
-                if ix >= begin_text + 1 && bytes[ix - 1] == b'\\' {
+                let eol_bytes = scan_eol(&s[ix..]).0;
+                if ix >= begin_text + 1 && bytes[ix - 1] == b'\\' && ix + eol_bytes < s.len() {
                     i -= 1;
-                    tree.append_text(begin_text, i);
-                    ix += scan_eol(&s[ix..]).0;
+                    unescape_and_append(tree, s, begin_text, i);
```
raphlinus: From an efficiency perspective, it seems like this ends up scanning plain text twice - once in this method, once in unescape. It's possible to do it once by including unescape triggers (`&`, probably add 0x00, the other two are `\` and `\r`, which are already covered) in the match.

But since this version is pretty clear and seems correct, I'm willing to take that up in a future performance pass.

In rough benchmarking, I'm seeing something like a 15% hit from this PR. As I say, acceptable for now, room for future improvement.

marcusklaas: Oof, I hadn't considered that. 15% is a pretty big hit. I guess we'll always have to make the trade-off between either scanning twice or allocating :thinking: Then the optimal strategy isn't as clear cut as I thought. Oh well, I learn!

marcusklaas:
Thanks for your patience and guidance!

PR #186, opened 2019-02-26
No double scan

marcusklaas:
This should fix the double scan issue in `parse_line` pointed out during review of the previous PR. Also tries to avoid allocation for link titles and destinations.


marcusklaas:
This reduces time on the rendering the [crdt document](https://github.com/xi-editor/xi-editor/blob/master/docs/docs/crdt-details.md) to html from ~729µs to ~622µs on my machine.

PR #187, opened 2019-02-26
Add first benchmark

marcusklaas:
Not comprehensive by any means, but should give us some indication of whether we're moving in the right direction.

### DIFF
--- a/benches/crdt.md
+++ b/benches/crdt.md
@@ -1,0 +1,1097 @@
+---
+layout: page
+title: CRDT - The Xi Text Engine
+site_nav_category_order: 206
+is_site_nav_category2: true
+site_nav_category: docs
+---
+
+<!-- See https://www.figma.com/file/UGOAcpKR5WIP81t3DGPIP2dR/CRDT-Merge-Diagrams for the source of the diagrams -->
+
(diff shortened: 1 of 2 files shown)


benches/lib.rs
```diff
@@ -0,0 +1,39 @@
...
+        let mut s = String::new();
+        match file.read_to_string(&mut s) {
+            Err(why) => panic!("couldn't open {}: {}", path.display(), why),
+            Ok(_) => s
+        }
+    }
+
+    #[bench]
+    fn crdt_empty_options(b: &mut test::Bencher) {
+        let input = read_file("./benches/crdt.md");
```
raphlinus: As an FYI, this can be made more robust (not depending on current directory at runtime) using the [`include_bytes!`](https://doc.rust-lang.org/1.8.0/std/macro.include_bytes!.html) macro.

marcusklaas: This is an excellent tip. I shall incorporate into the PR.

benches/crdt.md
```diff
@@ -0,0 +1,1097 @@
+---
```
ScottAbbey: Where did you get this file from? I think if you got it from https://github.com/xi-editor/xi-editor/, then we probably need to make some acknowledgement of that fact due to the Apache 2.0 license that repository seems to be released under.

Maybe file it under third_party/xi-editor similar to what is already there at third_party/CommonMark?

Issue #188, opened 2019-02-27
What table extension spec to choose?


marcusklaas:
The [markdown-extra](https://michelf.ca/projects/php-markdown/extra/#table) and [GFM](https://github.github.com/gfm/#tables-extension-) markdown table extensions don't agree on whether to use `<th>` or `<td>` for cells in table headers. It seems we must make a choice.

Any requirements/ preference on your side, @GuillaumeGomez ?

Possibly relevant: https://github.com/raphlinus/pulldown-cmark/pull/8

GuillaumeGomez:
I have a preference for `<th>` for header's cells. The rustdoc CSS uses it as well so that might be better for me. :3

marcusklaas:
GFM it is. We should render most commonmark-extra tables ok as well. The difference between the spec besides the `th`s in the header is that we fill table rows with empty cells, whereas commonmark-extra does not expect this.

PR #189, opened 2019-02-27
Implement table extension for new_algo

marcusklaas:
It currently fails almost all table tests in our suite, but I believe that's only because it expects `<td>`s in table headers whereas this currently produces `<th>`s.

Relevant issue: https://github.com/raphlinus/pulldown-cmark/issues/188


marcusklaas:
@raphlinus I'd like to add the GFM table tests to the repo, but it seems that the python scripts that generates the tests from the spec is not part of the repository.. is this hosted somewhere else?

ScottAbbey:
I think the test files are created by a build.rs file, not a Python script.

raphlinus:
Oops, for some reason I missed that this is in the review queue, I'll get to it soon. Yes, tests are generated by `build.rs`.

marcusklaas:
I added the table tests generated from GFM's spec and there are some legitimate failures. I'll try to fix those first :-)

marcusklaas:
Note: this is still not ready, but it shouldn't be too long anymore. Normalization of whitespace inside tables and normalization of th elements is done. I'll send you the batsignal when this is ready for a review ^^

marcusklaas:
This is ready for review now!

It passes all GFM's table tests. I needed to add some extra normalization, by dropping whitespace from table elements and removing empty tbody's, but I am fairly certain that these operations don't cause any semantic changes to the html.

Concerning commonmark-extra compliance: we pass the tests wherever commonmark-extra and GFM agree. There's two points on which they are irreconcilable: `td` vs `th` in `thead`s and adding filler cells. To pass commonmark-extra's table tests, I've transformed their `td`s to `th`s in the tests and disabled one test that expected no fillers to be added.

PR #190, opened 2019-02-28
Proof of concept for SIMD html escapes

marcusklaas:
This is a proof of concept to try and speed up some code using SIMD to see how much we could gain. HTML escapes seemed least complex, so that's what I tried. It's pretty much my first adventure using SIMD instructions, so there's probably a lot to improve still. There's a very limited set of bytes we're matching on and I tried to exploit some of their properties (most importantly, they have distinct lower nybbles!). Very much hoping to get additional insights through this PR though.

Overall speed gains are modest. I added two micro benchmarks for html escapes: one with lots of bytes which should be escaped (dense) and one without any (sparse).

test | baseline | simd
----------------|----|-------
test to_html::crdt_empty_options   |   646,362 ns/iter (+/- 26,357) | 631,209 ns/iter (+/- 27,171) |
test to_html::escape_html_dense   |    591 ns/iter (+/- 19) |   548 ns/iter (+/- 46) |
test to_html::escape_html_sparse   |     499 ns/iter (+/- 101) |   196 ns/iter (+/- 13) |

Which seems to provide only a 3% speedup overall, even though it look liked quite some time was spent in `escape_html` from the profile I ran...

Being a PoC, it doesn't do feature detection and only works on x64. Also doesn't leverage AVX2 and its 256 bit operations, since I don't have a machine supporting those instructions. It should be trivial to upgrade the code though.

Very, very grateful for any pointers or guidance!


ScottAbbey:
> Also doesn't leverage AVX2 and its 256 bit operations, since I don't have a machine supporting those instructions.

I've never used SIMD in Rust, and I don't have plans to help on this directly right now, but I can at least run this on a Skylake-SP machine supporting avx, avx2 and [several avx512 extensions](https://en.wikipedia.org/wiki/Advanced_Vector_Extensions#CPUs_with_AVX-512), if that would help you.

marcusklaas:
Thanks for that Scott! I don't think Rust exposes avx512 extension intrinsincs yet, but I may well take you up on that offer to test avx2 :-)

Keats:
There is https://crates.io/crates/v_htmlescape that does SIMD html escaping, you can have a look at the code/compare

marcusklaas:
Wow, that is a fascinating crate. Very generic. I think I can learn a lot from it. Thanks for the reference!

marcusklaas:
OK, I think this ready for serious review now. Ended up using a trick inspired by the v_htmlescape crate that prevents us from having to fall back to scalar code. I left some comments inline explaining. This speeds up the common case of short strings quite a bit. The crdt benchmark went from 575µs -> 515µs on my machine.

There's now also some conditional compilation for x64 architectures. I believe this does not limit binary portability as the required SSE instructions are all part of the x64 standard.

There's quite a number of experimental commits in this PR, so if we choose to merge, it's probably best to sqash them.

marcusklaas:
Closing this, as the `new_algo` branch is no longer used!

PR #191, opened 2019-03-01
Fix link nesting

raphlinus:
Need to preserve link nesting rather than just clear the stack.

Fixes spec test 500, bringing us to 631/633


raphlinus:
I think there might be some code cleanup, a bit of duplication in the code that forms links, but I wanted to PR a working version.

marcusklaas:
Looks good to me! This is cause for celebration! :-)

PR #192, opened 2019-03-01
Fix list errors

raphlinus:
Rework handling of list items starting with a blank line.

Don't set `last_line_blank` when inside a blockquote.

In combination with #191 this should get us to 633/633 spec tests.


marcusklaas:
Nice solution!

PR #193, opened 2019-03-02
Remove old parsing module and other unused code

marcusklaas:


src/prototype.rs
```diff
@@ -953,18 +1016,6 @@ impl<'a> Tree<Item<'a>> {
     }
 }
 
-#[allow(dead_code)]
```
raphlinus: I'm hesitant to delete it, as I've found it useful, but having unused code around for a long period of time is not great.

src/lib.rs
```diff
@@ -42,5 +41,4 @@ mod prototype;
 mod tree;
 mod linklabel;
 
-pub use crate::prototype::Parser;
-pub use crate::parse::{Alignment, Event, Tag, Options};
```
raphlinus: I'm not sure how much of a problem deleting the `parse` mod creates for merging with master, maybe it would be good to experiment. But in the not too distant future (maybe as a followup PR after the merge), it makes sense to rename `prototype` to `parse`.

PR #195, opened 2019-03-06
Big merge preparation

marcusklaas:
This merges master into new_algo and lays the groundwork for the new_algo -> master merge. There were no big conflicts.

There were two issues that I found:
 * the `push_html` function in the master branch writes the entire output to an intermediate buffer before copying all that to the given target string. This, together with ...
 * ... manual written byte counting seemed to incur a ~30% performance loss.

By writing to the string directly and foregoing manual byte counting, the performance regression from new_algo is limited to ~5%. Since we only push utf-8 to the output buffer, this should be safe. For those interested in the number of bytes written, it's easy to wrap a writer with an implementation of `Seek` - which should be faster to boot. This would be a breaking change.

Open to input of course!

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ construct an AST. Another advantage is that source-map information (the mapping
 between parsed blocks and offsets within the source text) is readily available;
 you basically just call `get_offset()` as you consume events.
 
-While manipulating AST's is the most flexible way to transform documents,
+While manipulating ASTs is the most flexible way to transform documents,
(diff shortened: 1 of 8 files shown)


marcusklaas:
I guess we could also do the wrapping of the `Writer` ourselves and preserve backwards compatibility.

marcusklaas:
Sounds good. That `WriteStr` trait in particular seems like a promising avenue. Agreed with making an issue to gather feedback - I'll open one tomorrow.

PR #196, opened 2019-03-07
String experiment for faster unescaping

marcusklaas:
As promised, here is a quick proof of concept for a new string type. It has two benefits over `Cow<str>`:

1. it does not allocate for single characters, which are used exclusively for unescaping
2. it takes up three instead of four words of memory, which could shrink our `Node<Item>` type by 16 bytes on x64 which is currently determined by the `Link` and `Image` variants, containing two such values. Shrinking this datatype reduces the memory footprint and hence improves the L1 cache performance of our AST tree

Downsides: 
* custom string type is exposed in API, which breaks backwards compatibility. However, by implementing `Deref` and `Borrow`, it can be used as a `&str` much like `Cow<str>`. Note that in the PR we didn't need to touch any code consuming the events.

Performance:
* on input without unescapes, performance difference is currently too small to measure. When converting more cows, we could gain additional throughput through improved cachability as described above
* on a heavy unescaped output, of which one example has been added to the benchmark suite, iteration time dropped from 32,000ns -> 20,000ns on my machine

Is this enough of a jump to justify adding a custom string type? Let's discuss.

### DIFF
--- a/src/strings.rs
+++ b/src/strings.rs
@@ -16,6 +16,12 @@ pub struct InlineStr {
     inner: [u8; DOUBLE_WORD_SIZE],
 }
 
+impl<'a> AsRef<str> for InlineStr {
+    fn as_ref(&self) -> &str {
+        self.deref()
+    }
+}
+
 impl Hash for InlineStr {
     fn hash<H: Hasher>(&self, state: &mut H) {
         self.deref().hash(state);
--- a/src/strings.rs
+++ b/src/strings.rs
@@ -1,8 +1,8 @@
 use std::ops::Deref;
-use std::borrow::Borrow;
+use std::borrow::{ToOwned, Borrow};
 use std::str::from_utf8_unchecked;
 
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub struct StackStr {
     inner: [u8; 4],
     len: usize,
@@ -29,10 +29,17 @@ impl Deref for StackStr {
     }
 }
 
+#[derive(Debug, Clone)]
 pub enum SmortStr<'a> {
     Boxed(Box<str>),
     Borrowed(&'a str),
     Stacked(StackStr),
+}
+
+impl<'a> std::cmp::PartialEq<SmortStr<'a>> for SmortStr<'a> {
+    fn eq(&self, other: &SmortStr) -> bool {
+        self.deref() == other.deref()
+    }
 }
 
 impl<'a> From<&'a str> for SmortStr<'a> {
@@ -71,6 +78,16 @@ impl<'a> Borrow<str> for SmortStr<'a> {
     }
 }
 
+impl<'a> SmortStr<'a> {
+    fn to_string(self) -> String {
+        match self {
+            SmortStr::Boxed(b) => b.into(),
+            SmortStr::Borrowed(b) => b.to_owned(),
+            SmortStr::Stacked(s) => s.deref().to_owned(),
+        }        
+    }
+}
+
 #[cfg(test)]
 mod test {
     use super::*;
@@ -95,10 +112,10 @@ mod test {
     }
 
     #[test]
-    fn smortstr_char_to_owned() {
+    fn smortstr_char_to_string() {
         let c = '藏';
         let smort: SmortStr = c.into();
-        let owned: String = smort.to_owned();
+        let owned: String = smort.to_string();
         let expected = "藏".to_owned();
         assert_eq!(expected, owned);
     }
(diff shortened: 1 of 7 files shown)


oberien:
I like that you guys care a lot about performance, and especially about possible performance regressions with the new_algo compared to the old parser. But I'm not sure if introducing a custom string type in the API makes sense here for some unescaping optimizations, and here are my reasons why:

1. For plain generators, this change might be negligible. They take the event stream, use everything as `&str`, and generate output. For my use-case (I'm writing a biased markdown → latex (and possibly markdown → html / blog / mdbook) converter; still a WIP, but close to its 0.1.0) I perform some heavy transformations on the event stream, before passing it to the generator. In fact [I wrote a `CowExt` trait](https://github.com/oberien/pundoc/blob/b153654c373db04b5caed95c04dd235604dc4eca/src/ext.rs#L29-L42) for easier transformations.
2. I'm not sure that microoptimizations like better caching will have a huge impact for most applications of pulldown_cmark. I imagine most uses to perform heavy file IO, reading from a file and writing back to a file. (Don't get me wrong, I love these kinds of micro-optimizations and tend to put way too much time into them :) )
3. I imagine unescaping single characters won't happen often in the wild. In fact until this PR I didn't even know they were a thing in CommonMark :) I always thought that unescaping in the markdown realm means unescaping backslash-escaped ascii punctuation characters ([which I really don't like](https://github.com/oberien/pundoc/issues/19#issuecomment-468943142), but that' a rant for another day :) ).

In the end on the one hand I don't think that a possibly rarely used feature and the cache micro-optimization are enough to justify adding this new string type.  
On the other hand, I'm already performing some `str` → `String` transformations ([concatenating all consecutive `Event::Text` into a single `Event::Text`](https://github.com/oberien/pundoc/blob/b153654c373db04b5caed95c04dd235604dc4eca/src/frontend/concat.rs), so adding another step of transforming `SmortStr<'a>` to `Cow<'a, str>` wouldn't be too much of a hassle in my case. This might also be true for most other non-simple-generator applications.

src/strings.rs
```diff
@@ -0,0 +1,122 @@
+use std::ops::Deref;
+use std::borrow::{ToOwned, Borrow};
+use std::str::from_utf8_unchecked;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct StackStr {
+    inner: [u8; 4],
+    len: usize,
```
raphlinus: A thought, `usize` is the natural size but in this case it might make sense to use a smaller type, `u8` is sufficient.

src/strings.rs
```diff
@@ -0,0 +1,122 @@
...
+    type Target = str;
+
+    fn deref(&self) -> &str {
+        unsafe {
+            from_utf8_unchecked(&self.inner[..self.len])
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
```
raphlinus: The name is whimsical, but I think confusing, I wonder if we can come up with a better one.

marcusklaas:
Thanks for the comments @oberien @raphlinus!

Do you have thoughts on this, @GuillaumeGomez?

src/strings.rs
marcusklaas: Yup, sounds v sensible. Went for the minimal implementation that worked for arbitrary `char`s, but indeed there is room for other short strings. Doing the box → stack check on clone is pretty smart and something we could definitely do!

src/strings.rs
marcusklaas: Yes! Naming things is hard. Very open to any other name :-)

GuillaumeGomez: "Smort" is a bit weird.

GuillaumeGomez:
Except for the naming ("SmortStr", really? XD), looks good to me!

marcusklaas:
Okay, I performed the following changes.

1. Renamed `StackStr` -> `InlinedStr` since it could technically be on the heap as well (although it wouldn't be very sensible) and the whimsical `SmortStr` is now the more descriptive `CowStr`. Considered calling it the even more descriptive `SmallCowStr`, but feared that'd be a bit long. Still very open to this and other options if anyone has an opinion/ suggestion.

2. Implemented @raphlinus' suggestions allowing more short inline strings (up to 15 bytes on 64bit architectures) and inlining of short strings on clone.

3. Moved all `Cow<str>`s to `CowStr`s. That should ensure forward compatibility/ consistency. As predicted, this dropped the size of the tree nodes from 104 bytes to 88 bytes on x64. On my machine, this alone improved the main benchmark time from ~525µs → ~510µs, that's the case *without* any unescapes. Unescapes themselves have between 60-75% higher throughput.

I hope this is now enough to make a decision on whether to go forward with this. I believe it's the last real blocker for merging `new_algo` to `master` and a 0.3 release :-)

marcusklaas:
:crown:

PR #197, opened 2019-03-08
Various optimizations

marcusklaas:
This does the following optimizations:
 * when appending text, checks if current node is a text node and tries to extend it. This ended up being pretty significant, dropping the render time on `crdt.md` from ~575µs → ~515µs on my machine
 * returning borrowed strings for refdef titles and autolinks whenever possible

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -84,6 +84,8 @@ pub enum LinkType {
     ShortcutUnknown,
     /// Autolink like `<http://foo.bar/baz>`
     Autolink,
+    /// Email address in autolink like `<john@example.org>`
+    Email,
 }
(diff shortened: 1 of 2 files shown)


src/parse.rs
```diff
@@ -1330,6 +1352,12 @@ fn unescape_cow<'a>(c: Cow<'a, str>) -> Cow<'a, str> {
 impl<'a> Tree<Item<'a>> {
     fn append_text(&mut self, start: usize, end: usize) {
         if end > start {
+            if let TreePointer::Valid(ix) = self.cur() {
```
raphlinus: Yay!

Issue #198, opened 2019-03-09
new_algo branch: panic on input ending with <


andersk:
```console
$ echo -n '<' | cargo run
    Finished dev [unoptimized + debuginfo] target(s) in 0.05s
     Running `target/debug/pulldown-cmark`
thread 'main' panicked at 'index out of bounds: the len is 1 but the index is 18446744073709551614', src/parse.rs:2069:8
note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

marcusklaas:
Thanks for filing this!

PR #199, opened 2019-03-09
Don't crash on < as input

marcusklaas:
Closes https://github.com/raphlinus/pulldown-cmark/issues/198.


raphlinus:
Thanks to both for the report and the fix!

PR #200, opened 2019-03-12
New algo -> master

marcusklaas:
Is this the time?

I think we can configure CI after this.

Maybe let's wait just a bit for initial feedback before tagging it 0.3?


Keats:
I can't promise any timeline but I'll try the branch with [Zola](https://github.com/getzola/zola)

marcusklaas:
That'd be amazing!

Cargo.toml
```diff
@@ -1,12 +1,14 @@
 [package]
 name = "pulldown-cmark"
-version = "0.2.0"
+# Note: This not a valid release, intended to be filled in when it's ready.
+version = "0.0.0"
```
raphlinus: I'm wondering whether this should be changed to "0.3.0" in this PR.

marcusklaas: Let's do it in a separate PR in a day or so?

marcusklaas:
:rocket:

Keats:
It's looking good!
I think I found a bug with footnotes: they are not processed if the footnote doesn't have a newline after, eg:

```rust
#"
Hello [My site][world]

<!-- more -->

Bla bla

[world]: https://vincentprouillet.com
"#
```
works but

```md
#"
Hello [My site][world]

<!-- more -->

Bla bla

[world]: https://vincentprouillet.com"#
```
doesn't.

marcusklaas:
Thanks for the test run @Keats! I'll open an issue for the footnotes bug.

Issue #162, opened 2018-11-11
Using `<details>` element results in extraneous and unbalanced paragraph elements


harding:
v0.2.0 via cargo

For this input:

```markdown
<details><summary>Testing 1..2..3..</summary>

This is a test of the details element.

</details>
```

I receive the following output from pulldown-cmark with unexpected extraneous paragraph tags that cause HTML5 validation to fail on the first and last `</p>` closing tags, as they're separated by nesting from their opening `<p>` tag and so are considered unbalanced:

```html
<p><details><summary>Testing 1..2..3..</summary></p>
<p>This is a test of the details element.</p>
<p></details></p>
```

Compare this to the expected output that I indeed receive from jgm cmark:

```html
<details><summary>Testing 1..2..3..</summary>
<p>This is a test of the details element.</p>
</details>
```

PR #201, opened 2019-03-13
Update list of html tags to 0.28 sepc

marcusklaas:
Fixes https://github.com/raphlinus/pulldown-cmark/issues/162 and introduces a regression test spec file.


src/scanners.rs
```diff
@@ -29,12 +29,13 @@ use crate::strings::CowStr;
...
-    "body", "button", "canvas", "caption", "col", "colgroup", "dd", "div",
-    "dl", "dt", "embed", "fieldset", "figcaption", "figure", "footer", "form",
-    "h1", "h2", "h3", "h4", "h5", "h6", "header", "hgroup", "hr", "iframe",
-    "li", "map", "object", "ol", "output", "p", "progress","section", "table",
-    "tbody", "td", "textarea", "tfoot", "th", "thead", "tr", "ul", "video"];
+const HTML_TAGS: [&'static str; 63] = ["address", "article", "aside", "base", "basefont",
+    "blockquote", "body", "caption", "center", "col", "colgroup", "dd", "details", "dialog",
+    "dir", "div", "dl", "dt", "fieldset", "figcaption", "figure", "footer", "form", "frame",
+    "frameset", "h1", "h2", "h3", "h4", "h5", "h6", "head", "header", "hr", "html", "iframe",
+    "legend", "li", "link", "main", "menu", "menuitem", "meta", "nav", "noframes", "ol",
```
ScottAbbey: Including `meta` here is inconsistent with the spec.txt currently included in this project, which has seen significant changes since the 0.28 release.

Ref: https://github.com/commonmark/CommonMark/commit/cb136a48ca97e3eb4dde7df775ebfe2cfcf087dd

Ref: https://github.com/raphlinus/pulldown-cmark/commit/5aa0f26bc80dc69253cb5dd1ae15b6f6e872df48#diff-da326c7eea772a16886c94e27b77a472R2045

marcusklaas: I'm confused.. there's a new version of the spec than 0.28? I've used [this list](https://spec.commonmark.org/0.28/#html-blocks) as a reference.

marcusklaas: It appears so: https://github.com/commonmark/CommonMark/commits/master

That's good to know. I'll use the most recent master list then :-) Thanks @ScottAbbey!

PR #206, opened 2019-03-14
Implement broken link callbacks

marcusklaas:
Hopefully takes care of the last remaining blocker for a 0.3 release.

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -2396,8 +2396,8 @@ impl<'a> Parser<'a> {
                                 } else if let Some(callback) = self.broken_link_callback {
                                     // looked for matching definition, but didn't find it. try to fix
                                     // link with callback, if it is defined
-                                    if let Some(val) = callback(link_label.as_ref(), link_label.as_ref()) {
-                                        Some((link_type.to_unknown(), val.0.into(), val.1.into()))
+                                    if let Some((url, title)) = callback(link_label.as_ref(), link_label.as_ref()) {
+                                        Some((link_type.to_unknown(), url.into(), title.into()))
                                     } else {
                                         None
                                     }
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -88,6 +88,17 @@ pub enum LinkType {
     Email,
 }
 
+impl LinkType {
+    fn to_unknown(self) -> Self {
+        match self {
+            LinkType::Reference => LinkType::ReferenceUnknown,
+            LinkType::Collapsed => LinkType::CollapsedUnknown,
+            LinkType::Shortcut => LinkType::ShortcutUnknown,
+            _ => unreachable!(),
+        }
+    }
+}
+
 #[derive(Clone, Debug, PartialEq)]
 pub enum Event<'a> {
     Start(Tag<'a>),
@@ -2189,6 +2200,7 @@ pub struct Parser<'a> {
     text: &'a str,
     tree: Tree<Item<'a>>,
     refdefs: HashMap<LinkLabel<'a>, LinkDef<'a>>,
+    broken_link_callback: Option<&'a Fn(&str, &str) -> Option<(String, String)>>,
 }
 
 impl<'a> Parser<'a> {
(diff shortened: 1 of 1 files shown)


src/parse.rs
```diff
@@ -2189,6 +2199,7 @@ pub struct Parser<'a> {
     text: &'a str,
     tree: Tree<Item<'a>>,
     refdefs: HashMap<LinkLabel<'a>, LinkDef<'a>>,
+    broken_link_callback: Option<&'a Fn(&str, &str) -> Option<(String, String)>>,
```
oberien: Would it be possible to change the broken_link_callback to return `Option<(Cow<'a, str>, Cow<'a, str>)>` instead for performance reasons? In my use-case I don't require changing the destination, but I use the broken_link_callback more as a filter of "valid" invalid destinations.

marcusklaas:
So the previous release did some normalization and passed this to the callback, but this is handled by UniCase now and hence we don't actually have access to a normalized string.. What to do here? We could import the normalization from the previous release, but then we'd be doing normalization twice. Ideally, the consumer takes care of normalization if they are interested in it, but that'd be a BC break. Guidance needed!

cc @GuillaumeGomez

src/parse.rs
marcusklaas: That sounds sensible. It's a BC break technically, but we're already backwards incompatible with regards to the `CowStr`. Maybe we can then also change the signature to take a single `&str` for which there is some reasoning in my other comment.

More input appreciated!

GuillaumeGomez:
Better twice than never I guess? But why is UniCase handling it directly?

marcusklaas:
UniCase has a more sophisticated implementation than we previously had. It is able to compare two string (slices) case insensitively without allocation. This is pretty tricky, so that's why we were happy to use that instead of writing our own. It also fundamentally prevents us from getting a normalized `&str`, because that'd require some sort of allocation anyway.

Do you rely on normalized input in rustdoc, @GuillaumeGomez?

GuillaumeGomez:
We take the link as is (we have to when we generate intra-docs links).

src/parse.rs
raphlinus: Looking at the code, I don't think a `Cow` here would actually reduce allocations, I think it would need to be converted into a `String` anyway after the call. In any case, I would worry about the complexity of the type (I think you need two lifetimes, the strings separate from `'a`).

src/parse.rs
```diff
@@ -2363,50 +2387,61 @@ impl<'a> Parser<'a> {
...
+                            else if let Some(ReferenceLabel::Link(link_label)) = label {
+                                let type_url_title = if let Some(matching_def) = self.refdefs.get(&UniCase::new(link_label.as_ref().into())) {
+                                    // found a matching definition!
+                                    let title = matching_def.title.as_ref().cloned().unwrap_or("".into());
+                                    let url = matching_def.dest.clone();
+                                    Some((link_type, url, title))
+                                } else if let Some(callback) = self.broken_link_callback {
+                                    // looked for matching definition, but didn't find it. try to fix
+                                    // link with callback, if it is defined
+                                    if let Some(val) = callback(link_label.as_ref(), link_label.as_ref()) {
```
raphlinus: This would probably be clearer as `(name, title)` rather than `val`, so `name.into()` instead of `val.0.into()`.

marcusklaas: I don't think the strings need to separate from `&'a` when we annotate the return type with the same lifetime? E.g.,
```rust
broken_link_callback: Option<&'a Fn(&'a str) -> Option<(Cow<'a, str>, String)>>
```
Fully agreed on your second comment. Will change that.

marcusklaas: Thinking about it a bit more, I'm starting to question if `Cow`-like return types would even make sense here. The closure is passed a label, not an url, so it's not natural to expect it to be returned as either the url or title...

raphlinus: Unifying lifetimes can cause borrow-checking problems for callers. This type of problem is subtle, so I'd only want to do it if we had evidence it was ok. Also what @marcusklaas said. I vote we keep it `String` and then possibly revisit this later.

marcusklaas: Ok! I'll fix up the other thing and merge it :-)

PR #208, opened 2019-03-14
Some clippy assisted cleanups

marcusklaas:


src/linklabel.rs
```diff
@@ -32,7 +32,7 @@ pub enum ReferenceLabel<'a> {
 pub type LinkLabel<'a> = UniCase<CowStr<'a>>;
 
 /// Returns number of bytes (including brackets) and label on success.
-pub(crate) fn scan_link_label<'a>(text: &'a str) -> Option<(usize, ReferenceLabel<'a>)> {
+pub(crate) fn scan_link_label(text: &str) -> Option<(usize, ReferenceLabel)> {
```
oberien: Using an anonymous lifetime like `ReferenceLabel<'_>` may be clearer here. There is actually the allow-by-default lint for this in the compiler (`elided_lifetime_in_path`) which is planned to be warn-by-default some time in the future iirc.

src/linklabel.rs
```diff
@@ -46,7 +46,7 @@ pub(crate) fn scan_link_label<'a>(text: &'a str) -> Option<(usize, ReferenceLabe
 
 /// Assumes the opening bracket has already been scanned.
 /// Returns the number of bytes read (including closing bracket) and label on success.
-pub(crate) fn scan_link_label_rest<'a>(text: &'a str) -> Option<(usize, CowStr<'a>)> {
+pub(crate) fn scan_link_label_rest(text: &str) -> Option<(usize, CowStr)> {
```
oberien: Same here with `CowStr<'_>`.

src/scanners.rs
```diff
@@ -670,12 +669,12 @@ pub fn scan_attribute_value(data: &str) -> Option<usize> {
...
     }
     if i >= data.len() { return None; }
-    return Some(i);
+    Some(i)
 
 }
 
 // Remove backslash escapes and resolve entities
-pub fn unescape<'a>(input: &'a str) -> CowStr<'a> {
+pub fn unescape(input: &str) -> CowStr {
```
oberien: ~

src/linklabel.rs
marcusklaas: I recall reading something about this. Good catch, will change!

src/parse.rs
```diff
@@ -1481,9 +1481,9 @@ fn get_html_end_tag(text : &str) -> Option<&'static str> {
 
     // TODO: Consider using `strcasecmp` here
     'type_1: for (beg_tag, end_tag) in BEGIN_TAGS.iter().zip(END_TAGS.iter()) {
-        if text.len() >= beg_tag.len() && text.starts_with("<") {
+        if text.len() >= beg_tag.len() && text.starts_with('<') {
             for (i, c) in beg_tag.as_bytes()[1..].iter().enumerate() {
-                if ! (&text.as_bytes()[i+1] == c || &text.as_bytes()[i+1] == &(c - 32)) {
+                if ! (&text.as_bytes()[i+1] == c || text.as_bytes()[i+1] == c - 32) {
```
raphlinus: I think I'd change the first conjunct to `text.as_bytes()[i+1] == *c`, or possibly `for (i, &c)` above.

src/strings.rs
```diff
@@ -149,7 +149,7 @@ impl<'a> Borrow<str> for CowStr<'a> {
 }
 
 impl<'a> CowStr<'a> {
-    pub fn to_string(self) -> String {
+    pub fn into_string(self) -> String {
```
raphlinus: Agree but note this is a semver break. It's fine if it's scooped into the 0.4 release.

marcusklaas: I don't think that's the reasoning behind the lint. `to_string` used to be slower because it went through the `fmt::Display` machinery, but that's not relevant here.

The lint that triggered this change seemed to warn against `to_*` methods taking self by value instead of by reference, that's all.

src/parse.rs
marcusklaas: Agreed, that's a better look.

PR #211, opened 2019-03-14
Bump version, update readme, implement source offset method

marcusklaas:
While going through the README, I found that `get_offset` was there to provide source mapping for blocks. However, it wasn't implemented yet. There is some ambiguity as how to do it for leaves; I've currently set the offset to their start, but setting it to their end seems just as feasible..

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -168,7 +168,7 @@ dependencies = [
 
 [[package]]
 name = "pulldown-cmark"
-version = "0.0.0"
+version = "0.3.0"
 dependencies = [
  "bitflags 1.0.4 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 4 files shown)


raphlinus:
The `get_offset` stuff doesn't match existing behavior. The result of `echo '*hello'* | cargo run -- -e` should be:

```
0: Start(Paragraph)
0: Start(Emphasis)
1: Text("hello")
6: End(Emphasis)
7: End(Paragraph)
8: EOF
```

I haven't dug into the code to see what's going on, but figured I would post this early.

marcusklaas:
Yeah I think the difference is in the leaf items. I'll play with it a bit.

marcusklaas:
Ok, I see what was going on. The offset is set to the start of the next item. Should be better now.

It passes your basic test atleast.

marcusklaas:
:ship:

PR #219, opened 2019-03-16
Make CowStr and InlineStr public

### LINKED ISSUE #218: CowStr and InlineStr are private
This makes it impossible to store them without allocating.

raphlinus:
Fixes #218


marcusklaas:
Looks good! r+

Issue #28, opened 2016-07-05
add checkboxes


bestouff:
Hi,
- [x] Would it be possible to add Tasks Lists please ?

DarkoDamjanovic:
Seems like task list/todo items/checkboxes are not part of the CommonMark spec:
https://spec.commonmark.org

notriddle:
Neither are tables or footers. We can add them as an extension.

DarkoDamjanovic:
Thanks, that would be great.

bestouff:
Nice, thanks a lot !

PR #220, opened 2019-03-16
Implement GFM's task lists spec

marcusklaas:
Fixes https://github.com/raphlinus/pulldown-cmark/issues/28.

- [x] add [GFM's task list tests](https://github.github.com/gfm/#task-list-items-extension-)
- [x] introduce new feature flag `ENABLE_TASKLISTS`
- [x] write initial implementation of spec
- [x] pass all tests in current suite
- [ ] review
- [ ] update PR


src/parse.rs
```diff
@@ -290,6 +294,15 @@ impl<'a> FirstPass<'a> {
                     self.begin_list_item = true;
                     return after_marker_index + n;
                 }
+                if self.options.contains(Options::ENABLE_STRIKETHROUGH) {
```
raphlinus: This needs to be ENABLE_TASKLISTS

marcusklaas: Woops - good catch!

tests/gfm_tasklist.rs
```diff
@@ -0,0 +1,68 @@
...
+
+include!("normalize_html.rs.inc");
+
+
+    #[test]
+    fn gfm_tasklist_test_1() {
+        let original = r##"- [ ] foo
+- [x] bar
+"##;
+        let expected = r##"<ul>
```
jasonm23, line 14: This test should expect list / list items with relevant classes.

```
<ul class="contains-task-list">
<li class="task-list-item"><input type="checkbox" id="" disabled="" class="task-list-item-checkbox">test item</li>
</ul>
```
This detail is missing from the GFM spec.  However there's no CSS way to locate the items and apply the style to get rid of the list item symbols.

https://github.github.com/gfm/#task-list-items-extension-

Issue #160, opened 2018-11-07
Iterator with offset


crawford:
I'd prefer if I didn't have to call get_offset() in order to read the location of the parser (I have to dump out the location of errors in my application). If the iterator was a tuple of Event and position, I wouldn't need to "manually" iterate over the parser.

Currently, I'm doing something like this:

```rust
while let Some(event) = parser.next() {
    let offset = parser.get_offset();
    ...
}
```

It would be great if I could do something like this instead:

```rust
for (event, offset) in parser.iter_with_offset() {
    ...
}
```

raphlinus:
This'll probably be in the batch of features that lands with new_algo. Here's a question for you though: I'm tracking both start and end offsets, not just start as before. Would it be useful to have both while I'm making the API? I don't believe the cost would be significant.

crawford:
Definitely. That would be helpful for syntax highlighting. Currently, I use the start position to calculate the line number, which is unhelpful for minified YAML (one line). Having a range would let me handle all cases elegantly.

Issue #140, opened 2018-07-03
Question: source maps supported or not?


akavel:
The readme says:

> - Versatile; in particular source-maps are supported

Is this actually supported as of current version, or not? If yes, how? Is it via the `get_offset` method of Parser? If yes, do I understand correctly that this doesn't let me for example detect what are the offsets of various parts of a link definition, thus in: `[some random link](https://github.com)` I can only get the offset where the whole thing starts and where it ends? Also cannot e.g. easily differentiate it from `[ref link][some-ref]`?

I'm searching for a Rust library which could let me write an editor with syntax highlighting and/or hiding of markdown syntax, so I need to be able to precisely position all such minute details...

raphlinus:
It's through `get_offset`, yes. I agree that's only fine-grained down to the level of individual AST nodes, not finer. I'd consider an extension that would report more detail. Also, I'd be happy for a PR that would clarify the language if it's confusing.

marcusklaas:
A new offset iterator with a range for each event has been implemented!

PR #222, opened 2019-03-16
Implement an event iterator that includes source offsets

marcusklaas:
Should be a bit more ergonomic than calling `get_offset` before each call to `next` on the iterator.

Fixes https://github.com/raphlinus/pulldown-cmark/issues/160.

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -21,6 +21,7 @@
 //! Tree-based two pass parser.
 
 use std::collections::HashMap;
+use std::ops::Range;
 
 use unicase::UniCase;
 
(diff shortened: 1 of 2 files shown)


marcusklaas:
cc https://github.com/raphlinus/pulldown-cmark/issues/140

src/parse.rs
```diff
@@ -2552,6 +2552,69 @@ impl<'a> Parser<'a> {
         }
         stack.pop_to(&mut self.tree, 0);
     }
+
+    /// Returns the next event in a pre-order AST walk, along with its
+    /// start and end offset in the source.
+    // TODO: this should probably be destructive. actually remove items from the tree
```
raphlinus: There might be scope for both. Non-destructive is one step towards AST access. In any case, since we use the cow pattern so much, I don't think the cloning of strings will be too bad.

marcusklaas: Those are good points. It's not as clear cut as I thought (I feel like I'm saying this a lot :sweat_smile:) - I will remove the TODO.

src/parse.rs
```diff
@@ -2552,6 +2552,69 @@ impl<'a> Parser<'a> {
...
+    // so we don't have to clone owned items (Strings)
+    fn next_event(&mut self) -> Option<(Event<'a>, usize, usize)> {
```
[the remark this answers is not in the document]
marcusklaas: I like it!

PR #224, opened 2019-03-17
Experiment with job syntax

raphlinus:


raphlinus:
If there's only one job, it's not needed. Closing.

PR #225, opened 2019-03-17
Add a CI status badge

raphlinus:
This is partly to test whether it will run CI properly, partly because it's actually useful.


marcusklaas:
Looks like it works working! :confetti_ball:

marcusklaas:
I hope we can cache images with rust built and perhaps even dependencies. It takes 3-4 min now to build. Something I can look into.

For now, this is good!

raphlinus:
Caching is coming soon: https://devblogs.microsoft.com/devops/adding-caching-to-azure-pipelines/

PR #230, opened 2019-03-19
Update scanner::scan_hrule to return Option<usize>

joshleeb:
rather than a `usize`, where `None` was represented by returning `0`.


joshleeb:
I also added a few `TODO` comments (along the same lines of returning `Option<T>` instead of `T`) which I'll fix up in subsequent PRs 😄

marcusklaas:
This looks good, thanks!

PR #232, opened 2019-03-19
Update scanner::scan_eol to return Option<usize>

joshleeb:
rather than a `(usize, bool)`


marcusklaas:
I like it, thanks!

PR #233, opened 2019-03-19
Update scanner::scan_blockquote_start to return Option<usize>

joshleeb:


marcusklaas:
Thanks!

PR #234, opened 2019-03-19
Update scanner::scan_code_fence to return Option

joshleeb:
This PR updates the `scanner::scan_code_fence` fn to return an
`Option<(usize, u8)>` rather than a `(usize, u8)`


marcusklaas:
Good stuff!

PR #235, opened 2019-03-19
Html escapes using SIMD

marcusklaas:
Adds SIMD html escaping for x64 platforms. Should provide a 5-10% speedup based on my tests.

It is now opt-in behind a SIMD feature flag.


src/escape.rs
```diff
@@ -131,6 +155,116 @@ where
...
         i += 1;
     }
-    w.write_all(&bytes[mark..])?;
+    w.write_all(&bytes[mark..])
+}
+
+/// Calls the given function with the index of every byte that is in
+/// [", &, <, >] and for no other byte.
+#[cfg(all(target_arch = "x86_64", feature="simd"))]
+fn scan_simd<F>(bytes: &[u8], mut offset: usize, mut callback: F) -> io::Result<()>
```
raphlinus: The name has a couple problems. First, its signature is not like the other scan methods. Second, it doesn't say what it does. I might suggest "foreach_special_simd".

src/escape.rs
```diff
@@ -104,7 +108,27 @@ static HTML_ESCAPES: [&'static str; 6] = [
...
+    })?;
+    w.write_all(&bytes[mark..])
+}
+
+#[cfg(not(all(target_arch = "x86_64", feature="simd")))]
+pub fn escape_html<W: Write>(w: W, s: &str) -> io::Result<()> {
+    escape_html_secure(w, s, false)
+}
+
+pub fn escape_html_secure<W>(mut w: W, s: &str, secure: bool) -> io::Result<()>
```
raphlinus: Not clear yet why we have this "secure" flag.

src/escape.rs
```diff
@@ -131,6 +155,116 @@ where
...
+        while mask != 0 {
+            let ix = mask.trailing_zeros();
+            callback(offset + ix as usize)?;
+            mask ^= mask & -mask;
+        }
+
+        offset += 16;
+    }
+
+    // final iteration. there's two scenario's:
```
raphlinus: This is a deep question, whether it's defined behavior to read from the same page. My personal feeling is that it's never ok to read a byte past the slice.

Also, I'm pretty sure this is not handling the empty case.

Here's the strategy I would employ:

If len < 16, use scalar fallback. There is basically no safe way to read SIMD here.

Process len / 16 chunks. Then if len % 16 != 0, read last chunk starting at len - 16, compute mask then shift right len % 16.

src/lib.rs
```diff
@@ -61,7 +61,7 @@ extern crate unicase;
 
 mod scanners;
 mod entities;
-mod escape;
+pub mod escape;
```
raphlinus: My understanding is this increases the public API surface? If so, worth mentioning in commit message.

src/escape.rs
marcusklaas: Right. I don't yet have hard evidence for this, but I've read `libc`'s `strlen` basically does this as well. They kind of have to use this technique as C strings are zero terminated and the length is not known in advance (that would make the function kind of trivial :p). It supposedly reads past the end of the buffer, but never past a 16 byte bound.

I can dig into this further tho. I think I have seen it in other libraries as well. `jetscii` comes to mind.

src/lib.rs
marcusklaas: That is again an on oversight on my part. That was only public for benchmarking purposes. Will remove.

src/escape.rs
raphlinus: I can see libc needing it, but in our case we have the length. This is an area where I'd really love input from an expert. @burntsushi ?

src/escape.rs
marcusklaas: Me neither, it's never used. Hysterical raisins, I guess. Can remove.

src/escape.rs
BurntSushi: I ran into a similar decision point in memchr, and my conclusion was "absent compelling evidence to the contrary, this is UB, so don't do it." I ended up [falling back to a simple scalar loop](https://github.com/BurntSushi/rust-memchr/blob/d70ea8a1dedd89be028e11f09c99423d4393e26f/src/x86/sse2.rs#L115-L123) when the input was too small to fit into a vector. Note, though, that the scalar loop only kicks in when the overall input is too small, and _not_ when the number of remaining bytes to look at is too small. Instead, we just accept that the last check might overlap with a previous check, but otherwise ensure that we are correctly reading in a full vector's worth of data: https://github.com/BurntSushi/rust-memchr/blob/d70ea8a1dedd89be028e11f09c99423d4393e26f/src/x86/sse2.rs#L182 --- I haven't looked at this code too carefully, but I imagine you can do the same by shifting some things around.

In my research, I came across this relevant SO answer: https://stackoverflow.com/questions/37800739/is-it-safe-to-read-past-the-end-of-a-buffer-within-the-same-page-on-x86-and-x64 However, with phrases like "Reading outside an object is of course Undefined Behaviour in C, but it's well-defined for C-targeting-x86," it feels like the accepted answer is on dubious ground.

From what I can tell, one way this might be safe is if you dropped down into Assembly and did reads past the end of the buffer there. But up in Rust or C land, I feel like you're on much shakier footing.

With respect to jetscii, I had thought [PCMPESTRI](https://www.felixcloutier.com/x86/pcmpestri) had a special exception for this, but I don't remember where I saw it. Personally, I feel like if you at any point have a pointer to some data that is out of bounds in a Rust program, you've probably entered UB territory.

I'd like to stress that I do not actually consider myself an expert in this, and that I feel like a true canonical answer would probably need to come from someone who is involved in rustc/lllvm development with respect to these special vendor intrinsics. (I imagine that's a fairly small number of people.) In general, my stance is a reflection of 1) no conclusive evidence that this isn't UB and 2) generally being conservative.

src/escape.rs
```diff
@@ -131,6 +155,116 @@ where
...
+    }
+
+    let lower_vec = unsafe { _mm_set_epi8( 0, 62, 0, 60, 0, 0, 0, 0, 0, 38, 0, 0, 0, 34, 0, 127 ) };
+    let upperbound = bytes.len().saturating_sub(16);
+
+    while offset < upperbound {
+        let mut mask = unsafe {
+            let raw_ptr = transmute(bytes.as_ptr().offset(offset as isize));
+            let v = _mm_loadu_si128(raw_ptr);
+            let expected = _mm_shuffle_epi8(lower_vec, v);
```
BurntSushi: `_mm_shuffle_epi8` is only available in SSSE3 or later, so you need, at minimum, a `#[target_feature(enable = "ssse3")]` and you need to ensure that you're only executing this code on `x86_64` platforms that support SSSE3, probably via [`is_x86_feature_detected`](https://doc.rust-lang.org/std/arch/index.html#dynamic-cpu-feature-detection), which I don't see here at all.

src/escape.rs
```diff
@@ -131,6 +155,116 @@ where
...
+        // that is potentially not ours and crash
+        return Ok(());
+    }
+
+    let lower_vec = unsafe { _mm_set_epi8( 0, 62, 0, 60, 0, 0, 0, 0, 0, 38, 0, 0, 0, 34, 0, 127 ) };
+    let upperbound = bytes.len().saturating_sub(16);
+
+    while offset < upperbound {
+        let mut mask = unsafe {
+            let raw_ptr = transmute(bytes.as_ptr().offset(offset as isize));
```
BurntSushi: You don't need a transmute here. See the `memchr` crate for how to use pointer casts instead.

src/escape.rs
```diff
@@ -131,6 +155,116 @@ where
...
+            let v = _mm_loadu_si128(raw_ptr);
+            let expected = _mm_shuffle_epi8(lower_vec, v);
+            let matches = _mm_cmpeq_epi8(expected, v);
+            _mm_movemask_epi8(matches)
```
BurntSushi: Hmmm, does this lead to false positives? Since you're checking only the nybbles of `&"<>`, I think it would be possible for non-ASCII bytes to be incorrectly reported as a position here. For example, I think _any_ byte whose 4 least significant binary digits is `0010` will be reported as a `"` by this code.

src/escape.rs
```diff
@@ -131,6 +155,116 @@ where
...
+        }
+    } else {
+        unsafe {
+            let raw_ptr = transmute(bytes.as_ptr().offset(offset as isize - alignment as isize));
+            // yay, aligned load!
+            let v = _mm_load_si128(raw_ptr);
```
BurntSushi: I don't think I'd bother with an aligned load here. Can you observe a perf difference from this? If you're going to worry about aligned loads, then it should probably be in the loop above, but even then I've found it difficult to observe a difference.

src/escape.rs
marcusklaas: Yeah, it's an oversight - will fix!

src/escape.rs
marcusklaas: :+1:

src/escape.rs
marcusklaas: I've read that on many modern CPUs there's even absolutely no difference. The aligned load is there for safety reasons (to not cross a 16-byte boundary and no page boundary by extension, at least, that's the idea), not performance. I guess the aligned intrinsic is there just for signalling.

src/escape.rs
marcusklaas: Are you sure? I believe it is sound. Indeed any byte with lower nibble `0010` will get mapped to `"`, but only `"` will test positive in the subsequent comparison?

src/escape.rs
BurntSushi: Ah gotya.

src/escape.rs
BurntSushi: Oh, I see! Gotya. Clever. So I'll turn this into another piece of feedback, but obviously you should defer to @raphlinus here since I'm just a driveby commenter. Namely, I would suggest:

1. Adding some high level docs inside this function describing how this routine works. If I may be allowed to continue to link to memchr... :-) https://github.com/BurntSushi/rust-memchr/blob/d70ea8a1dedd89be028e11f09c99423d4393e26f/src/x86/sse2.rs#L19
2. Add a test that specifically checks that false positives are prevented, e.g., use a byte with the same nybble as `"`. (Maybe you already have this test, but I can't tell easily.)

src/escape.rs
marcusklaas: Thanks so much for your detailed considerations Andrew! All the references are very helpful too.

Just a single note regarding the below.

> Personally, I feel like if you at any point have a pointer to some data that is out of bounds in a Rust program, you've probably entered UB territory.

I'm unsure if you meant this, but in this case all pointers should point to a safe range. It's just that a 16 byte read from that pointer may read from outside that range.

src/escape.rs
marcusklaas: The feedback is certainly appreciated man. Thank you. I will take it to heart :-) :bowing_man:

src/escape.rs
BurntSushi: Ah yes, right, that's what I meant. :-)

raphlinus:
Just a couple comments, and many thanks to @BurntSushi for the valuable feedback.

On modern x86, unaligned reads are almost as efficient as aligned, but I believe there is a tiny advantage. It's not worth contorting the code or adding more cases, though. On older x86 there used to be more, and on ARM I think it's also more (I wouldn't worry too much about ARM here).

I agree that more comments and doc would be useful. I was able to follow along because I've been thinking about this a lot, but even so it wasn't immediately clear. Somebody else trying to change the code, or adapt the ideas to other applications, would benefit from description. Also keep in mind that adding more special characters requires a code change (for example, this is likely to happen with math). Even now, we could be more sophisticated, for example not treating `~` as a special character unless strikethrough is enabled.

Looking forward to doing another review pass when the CPU detection etc is done, but no rush. Thanks again!

marcusklaas:
Ok, this should be a lot better now. No more out-of-bound reads, dynamic feature detection, better modularization and improved documentation.

src/escape.rs
```diff
@@ -123,14 +127,166 @@ where
...
+        table[(b'"' & 0x0f) as usize] = b'"';
+        table[0] = 0b0111_1111;
+        table
+    }
+
+    #[target_feature(enable = "ssse3")]
+    /// Computes a byte mask at given offset in the byte buffer. Its first 16 (least significant)
+    /// bits correspond to whether there is an HTML special byte (&, <, ", >) at the 16 bytes
+    /// `bytes[offset..]`. For example, the mask `(1 << 3)` states that there is an HTML byte
+    /// at `offset + 3`. It is only safe to call this function when
```
raphlinus: It might be clearer to put the `debug_assert` on length here, because this function is the one for which the safety condition is critical, and it would also get coverage for the final segment. But it's ok either way.

Issue #82, opened 2017-08-06
Quadratic time consumption on pathological input


DemiMarie:
See commonmark/cmark#218.  Pulldown-cmark is also vulnerable:
```
$ python -c 'print("[a](<b"*50000)' | time pulldown-cmark >/dev/null
pulldown-cmark > /dev/null  9.80s user 0.00s system 99% cpu 9.831 total
```

andersk:
Are you interested in further reports of this nature? The `new_algo` branch has a complexity regression compared to `master` where `"[](" * 50000` takes quadratic time instead of linear.

raphlinus:
Yes, we will want to start addressing these. I think the new_algo branch is a good basis to start cleaning this up; some of these would be very hard to fix in master.

PR #237, opened 2019-03-23
Guard against pathological link input

marcusklaas:
Fixes https://github.com/raphlinus/pulldown-cmark/issues/82.

### DIFF
--- a/benches/lib.rs
+++ b/benches/lib.rs
@@ -18,9 +18,8 @@ mod to_html {
     fn crdt_empty_options(b: &mut test::Bencher) {
         let input_bytes = include_bytes!("crdt.md");
         let input = from_utf8(input_bytes).unwrap();
-        let mut opts = Options::empty();
 
-        b.iter(|| render_html(&input, opts));
+        b.iter(|| render_html(&input, Options::empty()));
     }
 
     #[bench]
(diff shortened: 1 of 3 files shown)


benches/lib.rs
```diff
@@ -33,8 +32,30 @@ mod to_html {
...
+
+    #[bench]
+    fn pathological_links(b: &mut test::Bencher) {
+        let input = std::iter::repeat("[a](<").take(1000).collect::<String>();
+
+        b.iter(|| render_html(&input, Options::empty()));
+    }
+
+    #[bench]
+    fn pathological_codeblocks(b: &mut test::Bencher) {
```
raphlinus: This is not addressed directly in this PR, but maybe is one of the next things to work on?

marcusklaas: I argue in the final commit message that we don't need to: https://github.com/raphlinus/pulldown-cmark/pull/237/commits/f756258c8fdd438451df94646f8dce0617768449

marcusklaas:
Mixing the PRs was an oversight on my part! Apologies!

marcusklaas:
Mixing the PRs was an oversight on my part! Apologies!

raphlinus:
No worries :)

benches/lib.rs
raphlinus: Ah. There's still super-linear growth if your input is the code block sequence (O(n) AST nodes, O(n^2) bytes), followed by a sequence of O(n^2) AST nodes (i.e. `"*a* "` repeated n^2 times). Then time is O(n^3), so O(n^1.5) relative to input size.

marcusklaas: I believe that you are correct. Nice example! Then indeed we will need to do some extra work :-)

What would be a good way of dealing with this? We could keep a list of non-matching code blocks delimiters whenever we are scanning for a matching one. If it fails, it will contain all code block delimiters of the paragraph. Subsequent code block scans in the same paragraph could scan that vector instead of the AST. I think it should guard against your example.

PR #238, opened 2019-03-24
SIMD parse line

marcusklaas:
Use SIMD instructions to speed up scanning for special inline characters on x64. Since it involves some unsafe code, it is behind a feature flag.

Provides a solid overall 10% speedup on my machine.

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -30,7 +30,6 @@ use crate::strings::{CowStr, InlineStr};
 use crate::scanners::*;
 use crate::tree::{TreePointer, TreeIndex, Tree};
 use crate::linklabel::{scan_link_label, scan_link_label_rest, LinkLabel, ReferenceLabel};
-use crate::simd::{iterate_special_bytes, LoopInstruction};
 
 // Allowing arbitrary depth nested parentheses inside link destinations
 // can create denial of service vulnerabilities if we're not careful.
@@ -2634,6 +2633,39 @@ impl<'a> Parser<'a> {
             inner: self,
         }
     }
+}
+
+pub(crate) enum LoopInstruction<T> {
+    /// Continue looking for more special bytes, but skip next few bytes.
+    ContinueAndSkip(usize),
+    /// Break looping immediately, returning with the given index and value.
+    BreakAtWith(usize, T)
+}
+
+fn iterate_special_bytes<F, T>(bytes: &[u8], ix: usize, callback: F) -> (usize, Option<T>)
+    where F: FnMut(usize, u8) -> LoopInstruction<Option<T>> 
+{
+    #[cfg(all(target_arch = "x86_64", feature="simd"))]
+    { crate::simd::iterate_special_bytes(bytes, ix, callback) }
+    #[cfg(not(all(target_arch = "x86_64", feature="simd")))]
+    { scalar_iterate_special_bytes(bytes, ix, callback) }
+}
+
+pub(crate) fn scalar_iterate_special_bytes<F, T>(bytes: &[u8], mut ix: usize, mut callback: F) -> (usize, Option<T>)
+    where F: FnMut(usize, u8) -> LoopInstruction<Option<T>> 
+{
+    while ix < bytes.len() {
+        match callback(ix, bytes[ix]) {
+            LoopInstruction::ContinueAndSkip(skip) => {
+                ix += skip + 1;
+            }
+            LoopInstruction::BreakAtWith(ix, val) => {
+                return (ix, val);
+            }
+        }
+    }
+
+    (ix, None)
 }
 
 pub struct OffsetIter<'a> {
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -2642,6 +2642,16 @@ pub(crate) enum LoopInstruction<T> {
     BreakAtWith(usize, T)
 }
 
+/// This function walks the byte slices from the given index and
+/// calls the callback function on all byte (and their indices) that are in the following set:
+/// `` ` ``, `\`, `&`, `*`, `_`, `~`, `!`, `<`, `[`, `]`, `|`, `\r`, `\n`
+/// It may also call the callback on other bytes, but it is not guaranteed.
+/// Whenever `callback(ix, byte)` returns a `ContinueAndSkip(n)` value, the callback
+/// will not be called with an index that is lesser than `ix + n + 1`.
+/// When the callback returns a `BreakAtWith(end_ix, opt+val)`, no more callbacks will be
+/// called and the function returns immediately with the return value `(end_ix, opt_val)`.
+/// If `BreakAtWith(..)` is never returned, this function will return the first
+/// index that is outside the byteslice bound and a `None` value.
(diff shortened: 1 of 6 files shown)


src/simd.rs
```diff
@@ -0,0 +1,198 @@
...
+mod simd_test {
+    use super::{iterate_special_bytes, LoopInstruction};
+
+    fn check_expected_indices(bytes: &[u8], expected: &[usize], skip: usize) {
+        let mut indices = vec![];
+
+        iterate_special_bytes(bytes, 0, |ix| {
+            indices.push(ix);
+
+            // for type inference
```
raphlinus: I wouldn't be surprised if there were a better way to pin down the type, but ok.

src/simd.rs
```diff
@@ -0,0 +1,198 @@
...
+            LoopInstruction::BreakAtWith(ix, val) => {
+                return (ix, val);
+            }
+        }
+    }
+
+    (ix, None)
+}
+
+#[cfg(all(target_arch = "x86_64", feature="simd", test))]
```
raphlinus: I'm slightly uneasy the behavior is different in simd and non-simd cases, and this doesn't test the latter. But I think I see what you're doing.

src/simd.rs
```diff
@@ -0,0 +1,198 @@
...
+    );
+    let bitmask_lookup = _mm_setr_epi8(
+        1, 2, 4, 8, 16, 32, 64, -128,
+        1, 2, 4, 8, 16, 32, 64, -128
+    );
+
+    // actual computation
+    let raw_ptr = bytes.as_ptr().offset(ix) as *const _;
+    let input = _mm_loadu_si128(raw_ptr);
+    let bitset = _mm_shuffle_epi8(bitmap, input);
```
raphlinus: One of the problems we have here is that this is only available when the CPU has "ssse3" feature ([docs](https://doc.rust-lang.org/beta/core/arch/x86_64/fn._mm_shuffle_epi8.html)), and this is not guaranteed. For this code to be safe, there needs to be a runtime check. I haven't audited all such uses.

src/simd.rs
```diff
@@ -0,0 +1,198 @@
...
+            }
+        }
+
+        ix = std::cmp::max(ix, 16 + block_start);
+    }
+
+    let alignment = (bytes.as_ptr() as usize) % 16;
+    let remaining_bytes = bytes.len().saturating_sub(ix);
+
+    if remaining_bytes > 0 {
```
raphlinus: Similar concerns as #235; I'd prefer this not to ever read past the slice given. On the other hand, if we can get a determination from an unsafe Rust expert that this is ok, I might soften my stance.

src/simd.rs
marcusklaas: Agreed, this isn't great. I can try aligning the two without doing a double check on the non-simd case.

src/simd.rs
marcusklaas: There is, not sure why I didn't just typehint here. Will fix.

src/simd.rs
marcusklaas: Woops - hadn't realized that. Will fix, thanks!

marcusklaas:
Ok this looks to be pretty criminally fast when the simd flag is set now, and that's without out-of-bound reads!

src/parse.rs
```diff
@@ -2628,6 +2635,49 @@ impl<'a> Parser<'a> {
...
 
+pub(crate) enum LoopInstruction<T> {
+    /// Continue looking for more special bytes, but skip next few bytes.
+    ContinueAndSkip(usize),
+    /// Break looping immediately, returning with the given index and value.
+    BreakAtWith(usize, T)
+}
+
+/// This function walks the byte slices from the given index and
+/// calls the callback function on all byte (and their indices) that are in the following set:
```
raphlinus: Tiny typo: "all bytes"

src/parse.rs
```diff
@@ -2628,6 +2635,49 @@ impl<'a> Parser<'a> {
...
+/// `` ` ``, `\`, `&`, `*`, `_`, `~`, `!`, `<`, `[`, `]`, `|`, `\r`, `\n`
+/// It may also call the callback on other bytes, but it is not guaranteed.
+/// Whenever `callback(ix, byte)` returns a `ContinueAndSkip(n)` value, the callback
+/// will not be called with an index that is lesser than `ix + n + 1`.
```
raphlinus: "less than"

src/simd.rs
```diff
@@ -0,0 +1,215 @@
...
+//! The byteset is defined in `compute_lookup`.
+//!
+//! The idea is to load in a chunk of 16 bytes and perform a lookup into a set of
+//! bytes on all the bytes in this chunk simultaneously. We produce a 16 bit bitmask
+//! from this and call the callback on every index corresponding to a 1 in this mask
+//! before moving on to the next chunk. This allows us to move quickly when there
+//! are no or few matches.
+//!
+//! The table lookup is inspired by this [great overview]. However, since all of the
+//! bytes we're interested in are ASCII, we don't quite need the fully generality of
```
raphlinus: "full generality"

src/simd.rs
```diff
@@ -0,0 +1,215 @@
...
+use core::arch::x86_64::*;
+use crate::parse::LoopInstruction;
+
+const VECTOR_SIZE: usize = std::mem::size_of::<__m128i>();
+
+/// Generates a lookup table containing the bitmaps for our
+/// special marker bytes. This is effectively a 128 element 2d bitvector,
+/// that can be indexed by a four bit row index (the lower nibble)
+/// and a three bit column index (upper nibble).
+const fn compute_lookup() -> [u8; 16] {
```
raphlinus: Just wanna say, using a const fn for this kind of thing is great, much better than having random packed hex values in the code - in particular, easy to see how to update in case of more markup characters.

src/simd.rs
```diff
@@ -0,0 +1,215 @@
...
+    lookup
+}
+
+/// Computes a bit mask for the given byteslice starting from the given index,
+/// where the 16 least significant bits indicate (by value of 1) whether or not
+/// there is a special character at that byte position. The least significant bit
+/// corresponds to `bytes[ix]` and the most significant bit corresponds to
+/// `bytes[ix + 15]`.
+/// It is only safe to call this function when `bytes.len() >= ix + VECTOR_SIZE`.
+unsafe fn compute_mask(bytes: &[u8], ix: usize) -> i32 {
```
raphlinus: This needs a `#[target_feature(enable = "ssse3")]`, otherwise you might get less ideal code generation. It might also be worth adding an `#inline` - you'd need to look at generated asm to be very confident this is being done right. (If the generated asm is already good, it's because the compiler is automatically determining that inlining is appropriate, and the `target_feature_enable` on `simd_iterate_special_bytes` is in effect. But that's a bit fragile, for example if it's compiled optimized for size the inlining decision might go another way).

marcusklaas:
Thanks a 1E6 for the reviews Raph! I really appreciate it. :-)

I'll patch up the PR's where necessary and get them merged. Very happy about to start working towards a 0.5!

PR #240, opened 2019-03-26
Codeblock dos guard

marcusklaas:
This should guard against super-linear behaviour in examples like this: https://github.com/raphlinus/pulldown-cmark/pull/237#discussion_r268781036.


src/parse.rs
```diff
@@ -2284,6 +2284,8 @@ impl<'a> Parser<'a> {
     /// precedence, because the URL of links must not be processed.
     fn handle_inline_pass1(&mut self) {
         let mut link_stack = Vec::new();
+        let mut codeblock_delims: Vec<(usize, TreeIndex)> = Vec::new();
```
marcusklaas: I guess this could also be a `HashMap<usize, TreeIndex>` where we store the index of the last occurrence of each length delimiter. And then a scan for a match would simply be a lookup in the table plus a check if any match is the same as the opening delimiter.

marcusklaas: I've thought about this more, and just keeping track of the last occurrence of each delimiter size won't work. This would get parsed incorrectly:
```
` first occ of double backtick: `` snd: `` and third: ``
```
As the first occurrence of the double back tick would incorrectly match the last instead of the second.

raphlinus: I've thought about this a little and it's tricky. I think the general approach is to retain the results of the forward scan so it doesn't need to get redone.

An appropriate data structure is probably just Vec<Option<TreeIndex>>, where the index is the count. In real life, indices are tiny. In asymptotic analysis, this is O(n) which is acceptable, even if it *feels* like a sparser data structure should be better.

marcusklaas: I'm not sure if that that is much better than this PR's solution though (which also retains the forward scan results). It should be just as fast for common scenarios and also has O(n) worst-time complexity.

marcusklaas:
Ok, [this issue](https://github.com/raphlinus/pulldown-cmark/issues/247) convinced me that the previous approach probably wouldn't work. I implemented your HashMap idea and it seems to do well on all pathological codeblock inputs we've come up with so far :-)

andersk:
This one’s still a problem:

```console
$ python -c 'print("`a`" * 100000 + "`")' | time pulldown-cmark > /dev/null
0.97user 0.00system 0:01.00elapsed 97%CPU (0avgtext+0avgdata 20864maxresident)k
0inputs+0outputs (0major+4713minor)pagefaults 0swaps
$ python -c 'print("`a`" * 200000 + "`")' | time pulldown-cmark > /dev/null
3.92user 0.01system 0:03.99elapsed 98%CPU (0avgtext+0avgdata 39076maxresident)k
0inputs+0outputs (0major+9279minor)pagefaults 0swaps
$ python -c 'print("`a`" * 400000 + "`")' | time pulldown-cmark > /dev/null
20.63user 0.04system 0:20.87elapsed 99%CPU (0avgtext+0avgdata 75648maxresident)k
0inputs+0outputs (0major+18410minor)pagefaults 0swaps
```

marcusklaas:
Interesting, thanks! How do you find these examples anyway? Are you crafting them by hand or do you have some automated process generating them?

andersk:
For the most part, I’ve just been using a loop that searches through all possible repeating patterns from a set of interesting characters.

marcusklaas:
That should do it. Guess that `TODO` wasn't just a nice-to-have.. Thanks again for the great test cases @andersk!

Cargo.toml
```diff
@@ -12,6 +12,9 @@ readme = "README.md"
 
 build = "build.rs"
 
+[profile.release]
+debug = true
```
raphlinus: This'll possibly bloat code size a bit, so if this is just an artifact from your testing it should probably be removed.

src/parse.rs
```diff
@@ -2228,6 +2230,48 @@ struct LinkDef<'a> {
     title: Option<CowStr<'a>>,
 }
 
+struct CodeDelims {
+    inner: HashMap<usize, VecDeque<TreeIndex>>,
```
raphlinus: I think a comment explaining exactly what this represents would be useful.

Another thing to keep in mind is that an empty HashMap still has an allocation (one reason I sometimes prefer BTreeMap for similar purposes). But since this seems to be once per parse, shouldn't be a problem.

Cargo.toml
marcusklaas: It is an artifact of testing - will remove!

src/parse.rs
marcusklaas: TIL - I did not know about empty hashmaps allocating. When hashbrown is merged to the standard library this should no longer be the case I believe.

raphlinus: Good to know. Looks like hashbrown is going to be a major improvement!

PR #243, opened 2019-03-26
Buffer stdout for huge speedup

marcusklaas:
This provides a 2.61x speedup on my machine when using the binary.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -252,7 +252,7 @@ pub fn main() {
         }
     } else {
         let mut input = String::new();
-        if let Err(why) = io::stdin().read_to_string(&mut input) {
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -252,7 +252,7 @@ pub fn main() {
         }
     } else {
         let mut input = String::new();
-        if let Err(why) = io::stdin().read_to_string(&mut input) {
+        if let Err(why) = io::stdin().lock().read_to_string(&mut input) {
```
oberien: This line doesn't change anything, because that's already the internal implementation of `<Stdin as Read>::read_to_string`: <https://doc.rust-lang.org/src/std/io/stdio.rs.html#288-290>

PR #245, opened 2019-03-27
Fix panic on unterminated autolink (new_algo)

andersk:


marcusklaas:
This looks like a duplicate of https://github.com/raphlinus/pulldown-cmark/pull/244, closing.. but thanks again for the fix!

PR #250, opened 2019-03-27
Write html output to Vec before writing to output

oberien:
This provides a 5% speedup when tested on 10_000 × crdt.md. Unsure if it's worth it considering the highly increased RAM usage.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -262,8 +262,9 @@ pub fn main() {
         } else {
             let p = Parser::new_ext(&input, opts);
             let stdio = io::stdout();
-            let buffer = std::io::BufWriter::new(stdio.lock());
-            html::write_html(buffer, p).unwrap();
+            let mut buffer = Vec::with_capacity(input.len());
+            html::write_html(&mut buffer, p).unwrap();
+            std::io::BufWriter::new(stdio.lock()).write_all(&buffer).unwrap();
         }
     }
 }


marcusklaas:
Thanks for the PR, but this seems a bit excessive. The input can be arbitrary large (gigabytes), which would quickly generate memory problems. Perhaps we can achieve a similar speedup with a large but fixed size buffer. Have you tried `BufWriter::with_capacity`?

oberien:
Tested on 10000×crdt.md.  
`DEFAULT_BUF_SIZE` (`BufWriter::new()`): 8KB (<https://github.com/rust-lang/rust/blob/190feb65290d39d7ab6d44e994bd99188d339f16/src/libstd/sys_common/io.rs#L1>)

Capacity | Time 1 | Time 2 | Time 3 | Avg | Time rel. to Default
--- | :-- | :-- | :-- | :-- | --:
8KB | 5.941 | 5.906 | 5.930 | 5.926 | 100%
128KB | 5.889 | 5.836 | 5.745 | 5.824 | 98.3%
1MB | 5.703 | 5.664 | 5.699 | 5.689 | 96.0%
2MB | 5.721 | 5.792 | 5.727 | 5.747 | 97.0%
4MB | 5.740 | 5.715 | 5.737 | 5.726 | 96.7%
8MB | 5.727 | 5.754 | 5.767 | 5.749 | 97.0%

I can't really interpret this data, and I think considering the minimal performance improvements, more samples should be used. These tests are on a rather old machine (i7-4790, Samsung SSD 860 EVO). On my high-end machine (i7-6700K, Samsung 960 PRO NVMe M.2 SSD) I had 95% with 1MB and 92% at 8MB. With 128MB buffer I was back down to 100% compared to the default 8KB.

I guess it's safe to say that using 1MB is the safest bet and should give 4-5% performance improvement (which interestingly is the same as this PR, which buffers everything).

oberien:
Closing in favour of #252 .

Issue #248, opened 2019-03-27
Parsing a***a***a***a***… takes quadratic time (master, new_algo)


andersk:
```console
$ python -c 'print("a***" * 20000)' | time pulldown-cmark > /dev/null
0.48user 0.00system 0:00.52elapsed 93%CPU (0avgtext+0avgdata 10048maxresident)k
968inputs+0outputs (1major+2009minor)pagefaults 0swaps
$ python -c 'print("a***" * 40000)' | time pulldown-cmark > /dev/null
1.90user 0.00system 0:01.93elapsed 99%CPU (0avgtext+0avgdata 17196maxresident)k
0inputs+0outputs (0major+3864minor)pagefaults 0swaps
$ python -c 'print("a***" * 80000)' | time pulldown-cmark > /dev/null
9.40user 0.02system 0:09.56elapsed 98%CPU (0avgtext+0avgdata 32408maxresident)k
0inputs+0outputs (0major+7580minor)pagefaults 0swaps
```

See also mity/md4c#63.

marcusklaas:
These are very useful, thank you!

PR #254, opened 2019-03-27
Guard against pathological emphasis

marcusklaas:
This was a tough (but fun) one. Scanning the entire `InlineStack` for matching emphasis delimiters was a linear operation. This PR sets upperbounds on offsets in the stack to where we can limit the search. These upperbounds don't grow unless an element is added to the stack that will definitely match, thereby guaranteeing bounded search times.

Fixes https://github.com/raphlinus/pulldown-cmark/issues/248.

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -22,6 +22,7 @@
 
 use std::collections::{VecDeque, HashMap};
 use std::ops::Range;
+use std::cmp::min;
 
 use unicase::UniCase;
 use memchr::memchr;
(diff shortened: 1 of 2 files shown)


marcusklaas:
Ok, that seems completely reasonable to me. I shall try to prepare an argument in the meantime :-)

marcusklaas:
I think I was able to do it without extra vectors by using this approach: https://github.com/mity/md4c/issues/63#issuecomment-471767398. Keeping track of lower bounds was the trick apparently, not keeping upper bounds.

Now also fixes https://github.com/raphlinus/pulldown-cmark/issues/255.

src/parse.rs
```diff
@@ -1520,41 +1521,89 @@ struct InlineEl {
 #[derive(Debug)]
 struct InlineStack {
     stack: Vec<InlineEl>,
+    // lower bounds for
+    // _, non_both, * (mod 3), ** (mod 3), *** (mod 3)
+    // for example an underscore empasis will never match
+    // with any element in the stack with index lowerbounds[0] or below
```
raphlinus: I *think* it can match at the lowerbound (the code seems do to so), so this should probably read "will never match with any element in the stack below index lowerbounds[0]".

src/parse.rs
```diff
@@ -1520,41 +1521,89 @@ struct InlineEl {
...
+                let el = self.stack[i];
+                self.set_lowerbound(el.c, el.count, el.both, matching_ix.saturating_sub(1));
+                for i in 0..el.count {
+                    tree[el.start + i].item.body = ItemBody::Text;
+                }
+            }
+            self.stack.truncate(matching_ix);
+            Some(matching_el)
+        } else {
+            self.set_lowerbound(c, count, both, self.stack.len().saturating_sub(1));
```
raphlinus: Given the above comment, would this still work without the saturating_sub(1)? I could test it, but am lazy :)

src/parse.rs
marcusklaas: Yup, I believe you are right here.

src/parse.rs
marcusklaas: It doesn't look like it. The stack can be empty, in which case we'd overflow..

PR #259, opened 2019-04-05
Don't bother dropping parser in the binary

marcusklaas:
This saves us about 5% of the total time we'd otherwise spent deallocating
the various structures inside the parser, most notably the AST.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -258,6 +258,11 @@ pub fn main() {
             let stdio = io::stdout();
             let buffer = std::io::BufWriter::with_capacity(1*1024*1024, stdio.lock());
             html::write_html(buffer, &mut p).unwrap();
+            // Since the program will now terminate and the memory will be returned
+            // to the operating system anyway, there is no point in tidely cleaning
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -260,10 +254,11 @@ pub fn main() {
         } else if matches.opt_present("dry-run") {
             dry_run(&input, opts);
         } else {
-            let p = Parser::new_ext(&input, opts);
+            let mut p = Parser::new_ext(&input, opts);
             let stdio = io::stdout();
             let buffer = std::io::BufWriter::with_capacity(1*1024*1024, stdio.lock());
-            html::write_html(buffer, p).unwrap();
+            html::write_html(buffer, &mut p).unwrap();
+            mem::forget(p);
```
raphlinus: I'd add a comment here why we're doing this, because it's unexpected. People copying the code *should* do it if it's a simple commandline, but *should not* if it's processing lots of content or remaining online.

marcusklaas: Good point, will do!

PR #260, opened 2019-04-05
Make code spans into leaf nodes

marcusklaas:
Makes code blocks into leaf nodes, i.e., nodes without children. This simplifies parsing a bit and should signal better to the user what to expect: there is nothing but text inside a code block. This means that there is just one event per code block instead of a minimum of three.

As a trade-off, this code may now allocate where the previous approach did not. However, we do everything to avoid this and it should be rare. The reduced number of events should offset this potential extra cost and then some.

crdt benchmark, baseline
```
running 1 test
test to_html::crdt_empty_options               ... bench:     599,347 ns/iter (+/- 10,162)
```

crdt benchmark, with this PR
```
running 1 test
test to_html::crdt_empty_options               ... bench:     519,990 ns/iter (+/- 10,503)
```

During the making of this PR, I found that the test normalization code was too aggressive; it replaced breaks by whitespaces inside code blocks. This could allow us to pass tests [like this](https://spec.commonmark.org/0.28/#example-317) without actually doing the right thing. This has now been fixed.

**Note**: this is a breaking change.


src/html.rs
```diff
@@ -100,10 +100,17 @@ where
...
+                    self.write(b"<code>", false)?;
+                    escape_html(&mut self.writer, &text, false)?;
+                    self.end_newline = text.ends_with('\n');
+                    self.write(b"</code>", false)?;
+                }
                 Html(html) | InlineHtml(html) => {
                     self.write(html.as_bytes(), false)?;
                 }
                 SoftBreak => {
+                    dbg!("got here");
```
oberien: Should this still be here?

marcusklaas:
No! Good catch, thanks! :-)

oberien:
Would it make sense to move `Tag::CodeBlock(lang)` to `Event::CodeBlock(lang, code)` as well? It should also just be a single `Cow::Borrowed` containing the content of the whole code block, shouldn't it?

marcusklaas:
That's an excellent question. It may make sense. One thing to consider for code blocks is that they are likely to contain newlines. Whenever the code block is inside a container, like this,
```md
> ```rust
> check this
> sick codeblock
> ```
```
it would have to allocate and copy the code text, since the blockquote markers must be removed. I haven't checked whether the reduced number of events compensate fully for the allocation, performance wise. Even if not, it may still be worth it for the (common?) case where the code block is not inside a container. It would probably also be good for ergonomics/ consistency.

Hopefully we can get some additional input on this when this PR gets reviewed.

oberien:
Good point, I didn't think about this. Also for indented code blocks the leading spaces need to be removed. So even if a code block isn't in another container, it might be needed to be converted into a `String` and concatenated.

src/html.rs
```diff
@@ -100,6 +100,12 @@ where
                     escape_html(&mut self.writer, &text, false)?;
                     self.end_newline = text.ends_with('\n');
                 }
+                Code(text) => {
+                    self.write(b"<code>", false)?;
+                    escape_html(&mut self.writer, &text, false)?;
+                    self.end_newline = text.ends_with('\n');
```
raphlinus: This is maybe a minor point, but since there's a `</code>` closing tag, maybe `end_newline` should be `false`?

src/parse.rs
```diff
@@ -2580,14 +2577,15 @@ impl<'a> Parser<'a> {
 
     /// Returns the next event in a pre-order AST walk, along with its
     /// start and end offset in the source.
+    #[inline(always)]
```
raphlinus: Would just `#[inline]` be a serious performance regression? Generally `always` is frowned on, and this is a pretty big function to inline.

src/html.rs
marcusklaas: Good catch!

src/parse.rs
marcusklaas: It's not serious, but it is measurable: around 2%. This call is very hot. The reason for forcing inlining it was that the regular (non source-mapped) iterator has to strip source information otherwise. If it is inlined, the source references need not even be formed.

I suspect it's not inlined because it's fairly large as you say and it's used by two distinct iterators. But these are mutually exclusive in use for any given parser. So while forcing inlining here may increase code size a bit, it shouldn't hurt instruction caches.

marcusklaas: I have left a comment on it with the current considerations. We can revisit it later if when we find a better solution.

Issue #261, opened 2019-04-08
byte index 19 is not a char boundary; it is inside '»' 


Geobert:
Using 0.4
When writing:
```
[« Blurry Eyes »][blurry_eyes] est assez représentatif de
ce peut faire le bassiste (cliquez sur [le lien][blurry_eyes], elle n’est pas intégrable).
```

I got this error:
```
thread 'main' panicked at 'byte index 19 is not a char boundary; it is inside '»' (bytes 18..20) of `[« Blurry Eyes »][blurry_eyes] est assez représentatif de
ce peut faire le bassiste (cliquez sur [le lien][blurry_eyes], elle n’est pas intégrable).

« Pieces », Une jolie balade :

<div class="embed-video"><iframe src="https://www.youtube.com`[...]', src\libcore\str\mod.rs:2102:5
note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

After some try, it seems that using ` »` inside a `[]` is causing the error.
Note that the space before `»` is a non-breakable space.
I've tried: 
* `space»][link_ref]` -> working
* `][link_ref]nbsp»` -> working
* `nbsp"][link_ref]` -> working
* `nbsp»][link_ref]` -> broken

marcusklaas:
Thank you for filing this detailed report!

I will look into this bug immediately.

marcusklaas:
The bug appeared to be multi-byte spaces inside link labels. We normalize all spaces to regular spaces, but made some error in calculating the right index (and would end up in the middle of another multi-byte character). If it wasn't for `»`, we wouldn't crash but fail to recognize the link.

There's [a fix](https://github.com/raphlinus/pulldown-cmark/pull/262) for this, but review capacity seems to be very limited at the moment. I hope we can get it merged soon.

Thanks again for your report :-) These things really help make the project more robust.

Geobert:
Thank you for your analysis!

> We normalize all spaces to regular spaces

Only for internal use I presume? The nbsp is still output?

marcusklaas:
Correct!

PR #262, opened 2019-04-08
Fix byte offset bug in `scan_link_label_rest`

marcusklaas:
Fixes https://github.com/raphlinus/pulldown-cmark/issues/261.

### DIFF
--- a/src/linklabel.rs
+++ b/src/linklabel.rs
@@ -84,13 +84,12 @@ pub(crate) fn scan_link_label_rest(text: &str) -> Option<(usize, CowStr<'_>)> {
                     }
                 }
                 if whitespaces > 1 || c != ' ' {
-                    byte_index -= c.len_utf8();
                     if still_borrowed {
-                        label.push_str(&text[start_byte..byte_index]);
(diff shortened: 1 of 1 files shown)


src/linklabel.rs
```diff
@@ -84,13 +84,12 @@ pub(crate) fn scan_link_label_rest(text: &str) -> Option<(usize, CowStr<'_>)> {
...
                 if whitespaces > 1 || c != ' ' {
-                    byte_index -= c.len_utf8();
                     if still_borrowed {
-                        label.push_str(&text[start_byte..byte_index]);
+                        let end_ix = byte_index - c.len_utf8();
+                        label.push_str(&text[start_byte..end_ix]);
                         still_borrowed = false;
                     }
                     c = ' ';
-                    byte_index += c.len_utf8();
```
raphlinus: I agree this change is better (less mutation), but swapping lines 92 and 93 would also have worked, yes?

marcusklaas: Correct, that was my initial fix ^^

PR #272, opened 2019-04-15
Fix backslash unescaping logic

### LINKED ISSUE #270: `unescape` removes escaped backslashes
`unescape(…)` | Expected | Actual
--- | --- | ---
`\` | `\` | `\` ✓
`\\` | `\` | `\` ✓
`\\\\` | `\\` | `\` ✗
`\\\*` | `\*` | `*` ✗

raphlinus:
Fixes #270


raphlinus:
This simplifies some existing code, so it makes sense to check that I've got it right. In particular, the title in a refdef scan doesn't get changed, so it can be borrowed from the source string rather than being stored as a cow.

src/parse.rs
```diff
@@ -1319,31 +1311,12 @@ impl<'a> FirstPass<'a> {
                 }
                 '\\' => {
```
marcusklaas: I think we can skip this branch altogether now, no?

marcusklaas:
Nice cleanup! Just one request: since this wasn't caught by the spec, can we add a regression test for this ourselves?

src/parse.rs
raphlinus: Pretty sure not, you have to resolve the difference between `(title\))` and `(title))`. The former is a valid link title, the second isn't. As you say, tests will help settle the question, I'll add those.

PR #274, opened 2019-04-17
Smaller tree

marcusklaas:
These are some small optimizations to further speed up parsing. Most importantly, this PR reduces the size of `ItemBody` from 56 bytes to 16 bytes and makes it `Copy`, so it doesn't need to be dropped anymore. It also reserves space for the tree nodes in advance and increases the maximum inline string size for `CowStr`.

Some benchmarks using criterion.rs, which has proven to be much more reliable than the built-in bencher:

baseline:
```
crdt empty options      time:   [395.11 us 395.36 us 395.63 us]
```

pr:
```
crdt empty options      time:   [377.91 us 378.63 us 379.72 us]
```

which is a ~4.5% increase in throughput.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9,6 +9,24 @@ dependencies = [
 ]
 
 [[package]]
+name = "arrayvec"
+version = "0.4.10"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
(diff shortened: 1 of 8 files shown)


marcusklaas:
Thanks for the review!

PR #275, opened 2019-04-24
Move to CommonMark 0.29

marcusklaas:
This was way easier than expected! Most issues could be fixed with a few line changes.

### DIFF
--- a/src/scanners.rs
+++ b/src/scanners.rs
@@ -590,12 +590,12 @@ pub fn scan_listitem(data: &str) -> (usize, u8, usize, usize) {
     (w + postn, c, start, w + postindent)
 }
 
-fn char_from_codepoint_str(s: &str, radix: u32) -> char {
-    let mut codepoint = u32::from_str_radix(s, radix).unwrap();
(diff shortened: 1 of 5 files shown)


src/scanners.rs
```diff
@@ -604,16 +604,25 @@ pub fn scan_entity(data: &str) -> (usize, Option<CowStr<'static>>) {
     let mut end = 1;
     if scan_ch(&data[end..], b'#') == 1 {
         end += 1;
-        if end < size && (data.as_bytes()[end] == b'x' || data.as_bytes()[end] == b'X') {
+        let next_byte = data.as_bytes()[end];
+        if end < size && (next_byte == b'x' || next_byte == b'X') {
```
marcusklaas: Woops, this will probably panic now when `end >= size`. I'll revert this change real quick.

marcusklaas:
That sounds like a good plan to me! 

There's two documentation things that may be worthwhile to update: let people know about the existence of the simd feature and a warning to use the html renderer with a buffered writer to avoid performance bottlenecks. If we include that in the readme before releasing it will also be visible on crates.rs which may be useful?

PR #56, opened 2017-03-21
Remove unwanted conditional compilation

GuillaumeGomez:
Sorry for forgetting to remove it. :-/

PR #276, opened 2019-04-24
0.5

marcusklaas:
Prepares for a 0.5 release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,7 +131,7 @@ SIMD accelerated scanners are available for the x64 platform from version 0.5 on
 enable them, build with simd feature:
 
 ```bash
-> cargo build --features simd
+> cargo build --release --features simd
 ```
 
 Or add the feature to your project's `Cargo.toml`:
--- a/README.md
+++ b/README.md
@@ -108,7 +108,7 @@ strings are available when they're needed. Thus, when rendering text to
 HTML, most text is copied just once, from the source document to the
 HTML buffer.
 
-## Building only the pulldown-cmark library
+## Build options
 
 By default, the binary is built as well. If you don't want/need it, then build like this:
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -119,7 +124,20 @@ By default, the binary is built as well. If you don't want/need it, then build l
...
 ```toml
-pulldown-cmark = { version = "0.4", default-features = false }
+pulldown-cmark = { version = "0.5", default-features = false }
+```
+
+SIMD accelerated scanners are available for the x64 platform from version 0.5 onwards. To
+enable them, build with simd feature:
+
+```bash
+> cargo build --features simd
```
raphlinus: I think it might make sense to add `--release` here, a debug build with SIMD isn't very helpful.

src/html.rs
```diff
@@ -409,6 +409,11 @@ where
 /// Iterate over an `Iterator` of `Event`s, generate HTML for each `Event`, and
 /// write it out to a writable stream.
 ///
+/// **Note**: using this function with an unbuffered writer like a file or socket
+/// will result in poor performance. Wrap these in a
+/// [`Bufwriter`](https://doc.rust-lang.org/std/io/struct.BufWriter.html) to
```
raphlinus: BufWriter (capitalization)

README.md
```diff
@@ -119,7 +124,20 @@ By default, the binary is built as well. If you don't want/need it, then build l
...
+```
+
+Or add the feature to your project's `Cargo.toml`:
+
+```toml
+pulldown-cmark = { version = "0.5", default-features = false, features = ["simd"] }
```
raphlinus: I wonder if we can remove the getopts default feature from library builds while leaving them in for binaries (the current state was added in #56), but I don't see an obvious way to do it, at least without having two separate crates, one for the lib and one for the bin. This is probably a pretty minor nit and shouldn't block this release.

marcusklaas: Hmmm yes, that would be nice. I don't see how we could do it either though…

Issue #266, opened 2019-04-13
Remove unsafety in html renderer


marcusklaas:
The `StrWrite` trait idea should probably work: a trait that contains a single `write_str` method which returns `io::Result<()>` which is implemented by `String` and any `Write` type.

PR #277, opened 2019-04-25
Remove html renderer unsafety

marcusklaas:
Addresses https://github.com/raphlinus/pulldown-cmark/issues/266 by introducing a `StrWrite` trait that enables writing string slices. It has a blanket implementation for types implementing `std::io::Write`, so we should remain fully backward compatible.

This blanket implementation does force some trickery since `std::io::Write` is an external trait. Rust won't let us implement `StrWrite` for types that are not defined in the crate itself since it can't know they won't implement `Write` in the future and introduce a conflict. Hence we need wrapper types, one for `String` and one for `&'_ mut W where W: Write`. This solution isn't particularly elegant, so I'm very open to alternatives!

Performance is within 1% of master.

### DIFF
--- a/src/html.rs
+++ b/src/html.rs
@@ -23,6 +23,7 @@
 use std::collections::HashMap;
 use std::io::{self, Write};
 
+
 use crate::parse::{LinkType, Event, Tag, Alignment};
 use crate::parse::Event::*;
 use crate::strings::CowStr;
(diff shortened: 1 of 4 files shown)


src/html.rs
```diff
@@ -33,10 +34,69 @@ enum TableState {
     Body,
 }
 
-struct HtmlWriter<'a, I, W>
-where
-    W: Write,
+/// Wrapper for String for which we can implement `StrWrite`.
```
raphlinus: I see where these wrappers are coming from, the blanket impl for `Write` makes them necessary, but I wonder, what would it look like the other way, where `String` and `&mut impl StrWriter` impl the trait, but `Write` needs a wrapper?

marcusklaas: It's a good question, and one I will investigate!

marcusklaas: Your suggestion not only works well, it's also much cleaner! Thanks :-)

PR #279, opened 2019-04-26
Some small optimizations

marcusklaas:
I hope you aren't getting too tired of these optimization PRs. It appears there is still some juice left to squeeze in this thing. Getting rid of the `#[inline(always)]` and implementing the two event iterators separately yielded the largest speedup. Then there's a short circuit for html scanning and some other small scanner optimizations.

Master:
```
crdt_total              time:   [384.45 us 384.82 us 385.25 us]
```
PR:
```
crdt_total              time:   [356.17 us 357.33 us 358.61 us]
```
In total, this produces a ~7.5% speedup.


marcusklaas:
```
crdt_total              time:   [350.39 us 350.70 us 351.05 us]
```

raphlinus:
Not tired of the improvements at all! But am dealing with overload atm, might not be able to get to these in detail very soon.

marcusklaas:
Yes!! Indeed, `as_bytes` is a no op! But it's the string subslicing that's expensive apparently!

marcusklaas:
All the char boundary checking.. I'm not quite done yet, but you should give this PR a spin. It *flies* dude!

marcusklaas:
Current benchmark time on my machine:
```
crdt_total              time:   [338.86 us 339.12 us 339.41 us]
```
or about 13.4% higher throughput!

raphlinus:
On mine (13" MBP 2017), with target-cpu=native and simd feature:

```
crdt_total              time:   [258.98 us 259.69 us 260.59 us]
```

![Neo whoa gif]

src/parse.rs
```diff
@@ -1523,8 +1517,8 @@ impl InlineStack {
         }
     }
 
-    fn pop_all<'a>(self, tree: &mut Tree<Item>) {
-        for el in self.stack {
+    fn pop_all<'a>(&mut self, tree: &mut Tree<Item>) {
+        for el in self.stack.drain(..) {
```
raphlinus: I'm not sure why this works, but it doesn't break any tests. For future reference, the test the original version of the code was meant to fix is now [0.29 test 516](https://spec.commonmark.org/0.29/#example-516).

src/parse.rs
```diff
@@ -2126,8 +1957,9 @@ fn scan_reference<'a, 'b>(tree: &'a Tree<Item>, text: &'b str, cur: TreePointer)
         TreePointer::Valid(cur_ix) => cur_ix,
     };
     let start = tree[cur_ix].item.start;
+    let tail = &text.as_bytes()[start..];
     
-    if text[start..].starts_with("[]") {
+    if tail.len() >= 2 && tail[0] == b'[' && tail[1] == b']' {
```
raphlinus: Did you try `starts_with(b"[]")` here (and in related places)? It seems clearer and I'd be surprised if it generated worse code.

src/parse.rs
marcusklaas: The drain isn't functionally different than consuming the vector, right? (Note that the function signature moved from `self` to `&mut self`, so the (empty) vector inside can be reused later)

src/parse.rs
marcusklaas: Good point! Godbolt confirms that the two generate similar code. Will change.

marcusklaas:
Thanks for the review!

In total these changes amount to a ~18% throughput increase on my machine.

Issue #280, opened 2019-04-29
New denial of service attack!


marcusklaas:
Was going over the inline html scanning routines when I ran into some questionable scanning methods. In particular, parsing repetitions of `"a <![CDATA["` will scale quadratically!

Haven't tested other parsers yet. It would probably be good to test them as well and let them know when they are also vulnerable.

marcusklaas:
md4c is vulnerable also

marcusklaas:
and so is cmark..

mity:
@marcusklaas Fixed in MD4C by remembering how far we have scanned for the expected HTML closer without finding it. Later attempts for re-scanning for the same closer then may fail early.

Note other raw HTML openers (e.g. HTML processing instructions or declarations) exhibited the same problem so you may check it too.

marcusklaas:
Thanks for the pointers! Good to have the parser people looking out for eachother ^^

The `<?` and `<A` vulnerabilities I hadn't caught and by looking at the code, pulldown-cmark is vulnerable to these as well.

oberien:
Is `a <![CDATA[` just a special case of `a <!A`? Or is it actually due to containing `<![CDATA[`?

mity:
The specs sees them as separate because the rules for starting/ending them is different.

See https://spec.commonmark.org/0.29/#declaration versus https://spec.commonmark.org/0.29/#cdata-section.

That said, implementation make share some code, but obviously not all of it.

PR #281, opened 2019-04-30
Rework inline html scans

marcusklaas:
This PR
 - removes `InlineScanner` in favor of simpler (existing) byte scanners
 - improves scanning throughput on inline HTML and autolinks by ~10%
 - guards against quadratic scanning behavior on pathological HTML (https://github.com/raphlinus/pulldown-cmark/issues/280)
 - fixes a quadratic scanning regression on inline links and improves their general scanning performance by ~4%
 - moves a bunch of scanning functionality from the parse module to the scanner module
 - fixes https://github.com/raphlinus/pulldown-cmark/issues/287, https://github.com/raphlinus/pulldown-cmark/issues/288 and https://github.com/raphlinus/pulldown-cmark/issues/285


marcusklaas:
I may need to take another look at the `HtmlScanGuard`s. Setting a flag may not be enough because I think it may be the case that we do not scan the entire text..

Issue #284, opened 2019-05-01
Non-linear behaviour when parsing `[[]()...`


oberien:

marcusklaas:
Erm, have you tried this on https://github.com/raphlinus/pulldown-cmark/pull/281? I think this was fixed as well :sweat_smile:

oberien:
I am in fact running my fuzzer (actually it is a fuzzer by definition) on that branch.

marcusklaas:
Excellent, good find then! Thanks :-)

PR #286, opened 2019-05-01
Find non-linear growth patterns

### LINKED ISSUE #257: Create a framework for detecting quadratic time regressions
There's a bunch of test cases provided by @andersk on which pulldown_cmark previously exhibited quadratic time behaviour. Many should be fixed now, but at the moment there is no (automated) way of ensuring that we don't regress on them.

We would ideally have a test pass which runs those examples one-by-one in release mode and make sure they don't time out (and thereby providing some assurance that we haven't regressed horribly).

Here's one way of doing it:
- create a separate test module, so it can be easily skipped or singled out
- create a single test in that module, ensuring that no two time-out tests run simultaneously
- loop through a static array of function pointers generating a testcases (`Fn() -> String`), for each we
  * create timeout thread, which returns after a fixed time
  * create an execution thread, which parses and renders to html into a `std::io::Sink`
  * wait for the first thread to return and make sure the execution thread returned first

Some care will have to be taken to pick the right testcase size and timeout time. Luckily, if we truly regressed to quadratic behaviour, execution time will blow up very quickly. Therefore, we have some margin to work with.

You could try with sizing your testcases so that they return in ~50ms on your local machine and then setting the timeout multiplier to 50. That should make sure that we don't get too many spurious failures on machines that are much weaker or under heavy load. This will likely require some tweaking.

Of course, this is just one approach. There may very well be more robust approaches. Feel free to get creative!

Edit: I think benchmarks can have assertions. It may be best to implement this as a benchmark module. That automatically guarantees release compilation and that no two tests are run simultaneously.

oberien:
Fix #257 
Depends on #281 (which is why the diff includes those commits as well). I'm working on that branch as several patterns and issues I found have already been fixed there.

For my notes about the concept used and different things I tried, see [this hackmd](https://hackmd.io/S227dq9kSli87yHgQmyP7Q).

Basically I'm writing an intelligent fuzzer, which parses the pulldown-cmark source-code, extracts all literals, then tests if combinations of those literals result in non-linear behaviour.


marcusklaas:
Very interesting — thanks for working on this! The general approach of forming test strings of different lengths and measuring the slope seems sound to me. I will have more time to closely look into this tomorrow.

For now, I have these questions:
 - it looks like a lot of code deals with parsing source code looking for literals. I could only find a reference to `../src/entities.rs`. Does it take the literals found there? What's the advantage over having a manually defined set of literals?
- is it correct that combinations of these literals are taken and are then different repetition counts are used to create test strings of different lengths?
- could this be used to do regression testing of known pathological inputs?

oberien:
> * it looks like a lot of code deals with parsing source code looking for literals. I could only find a reference to `../src/entities.rs`. Does it take the literals found there? What's the advantage over having a manually defined set of literals?

I use `WalkDir` to get all files in `../src` (which are all source files of pulldown-cmark). I realized that the literals in `entities.rs` don't add much value, but clobber the number of combinations extremely. That's why I just straight up blacklisted that file. I'm still left with 227 different literals (which would be >8k more if I included `entities.rs`), too much to exhaust the search space. That's why I decided to introduce randomness for fuzzing.

> * is it correct that combinations of these literals are taken and are then different repetition counts are used to create test strings of different lengths?

I extract all literals from the `pulldown-cmark` source code (including `"foo"`, `b"foo"`, `'x'` and `b'x'`). I put all of them in a list. Then I generate all combinations of those literals of length 4. For example given the literals `"foo", "x", "{", "<?", "!"`, one possible combination would be `"x<?!foo"`. I then repeat that combination such that I get an input of 256 KB. That length should be large enough to reduce variance.

>     * could this be used to do regression testing of known pathological inputs?

Yes, it's as simple as adding `test("a<?");` to the top of the main function (which is what I do when minifying a find).

find_exponential/src/main.rs
```diff
@@ -0,0 +1,197 @@
...
+        });
+        if count > 0 {
+            println!("{:<20} throughput: {}", count, count as u64 / start_time.elapsed().as_secs());
+        }
+    }
+}
+
+fn test(pattern: &str) {
+    let (_, _, non_linear) = test_pattern(pattern, SAMPLE_SIZE, false);
+    if non_linear {
```
marcusklaas: Is this for retesting/ validation?

find_exponential/src/main.rs
```diff
@@ -0,0 +1,197 @@
...
+    while !s.is_char_boundary(n) {
+        n += 1;
+    }
+    (time_needed(&s[..n]), n)
+}
+
+fn time_needed(s: &str) -> Duration {
+    let parser = Parser::new_ext(&s, Options::all());
+    let time = Instant::now();
+    parser.for_each(|evt| {
```
marcusklaas: Why is the black box needed? What happens to the event isn't particularly important, right? I think a `parser.count()` would be fine here.

find_exponential/src/scoring.rs
```diff
@@ -0,0 +1,42 @@
...
+use itertools::Itertools;
+
+#[allow(dead_code)]
+pub fn pearson_correlation(array: &Array2<f64>) -> (f64, bool) {
+    let corr = array.pearson_correlation()[[1, 0]];
+    (corr, corr < super::ACCEPTANCE_CORRELATION)
+}
+
+#[allow(dead_code)]
+pub fn slope_stddev(array: &Array2<f64>) -> (f64, bool) {
```
marcusklaas: Can you explain how this works?

oberien:
Thanks for the review. I took notes in the hackmd while the idea was still in flux. Now that the general concept seems to work, I can move those notes into comments in the source-code.

I'm currently rewriting outlier handling, because behaviour like in #287 currently result in an infinite loop. After that I'll get back to your comments :)

find_exponential/src/main.rs
```diff
@@ -0,0 +1,197 @@
...
+    let (_, _, non_linear) = test_pattern(pattern, SAMPLE_SIZE, false);
+    if non_linear {
+        let (array, score, non_linear) = test_pattern(pattern, SAMPLE_SIZE*2, true);
+        if non_linear {
+            print_find(pattern, &array, score);
+        }
+    }
+}
+
+fn test_pattern(pattern: &str, sample_size: usize, remove_outliers: bool) -> (Array2<f64>, f64, bool) {
```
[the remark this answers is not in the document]
oberien: Pretty much, yes. I could use a `Vec<[f64; 2]>` instead, as converting that to an `Array2` [is pretty much zero-cost](https://github.com/rust-ndarray/ndarray/blob/27c84a5eb72fc0c399cfb804eac3a5e6f2ed936d/src/free_functions.rs#L235-L262). Not sure if that is much better though.

find_exponential/src/main.rs
```diff
@@ -0,0 +1,197 @@
...
+    let mut i = 0;
+    while i < sample_size {
+        let (dur, n) = calculate_point(&s, i+1, sample_size);
+        array[[0, i]] = n as f64;
+        array[[1, i]] = dur.as_nanos() as f64;
+        if DEBUG_LEVEL >= 3 {
+            println!("duration: {}", dur.as_nanos());
+        }
+
+        if remove_outliers && i > 0 && array[[1, i-1]] > array[[1, i]] {
```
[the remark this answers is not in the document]
oberien: I try to keep the throughput as high as possible. As such, I first do a fast test, to filter out most non-linear patterns. However rescheduling / spikes result in skewed values. As such if the first test detects possible non-linear behaviour, I reperform the test, but with higher precision. The second test is the first test, but with double as many samples. Additionally, if I encounter an outlier, which is any value that is larger than its successor (because I expect a strict growth of parsing time with increasing length), I recompute that value. While this is rather expensive, it highly reduces false positives due to rescheduling.
Unfortunately it's not always true that parsing time increases strictly, as seen in #287. Therefore, I have now implemented detection of large outliers (at least 50× as large as their successor), which I try to recompute ten times. If they are a large outlier all those ten times, I mark that pattern as weird pattern (non necessarily non-linear, but at least it has weird parsing behaviour).

>  Ideally, we'd use a method that is resistant to spurious outliers.

Outliers aren't uncommon unfortunately due to rescheduling. If you have an idea how to process outliers in the scoring function, feel free to tell me and I'll implement it :) As it is right now it is able to handle some outliers pretty well (the first test can contain a few outliers as long as they aren't too drastic). If it fails however, I'll manually remove the outliers in the second test.

I'll add a comment explaining this.

find_exponential/src/main.rs
```diff
@@ -0,0 +1,197 @@
...
+        rx.recv().unwrap();
+        let mut literals = literals.clone();
+        let tx = tx.clone();
+        count += BATCH_SIZE as isize;
+        pool.execute(move || {
+            literals.shuffle(&mut rand::thread_rng());
+            let combs = literals.iter()
+                .combinations(COMBINATIONS)
+                .take(BATCH_SIZE);
+            for comb in combs {
```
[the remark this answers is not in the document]
oberien: `combinations()` itself is predictable. That's why for every `BATCH_SIZE`, I'm first randomly shuffling the literals: `literals.shuffle(&mut rand::thread_rng());`. That results in the generated combinations being random for every `BATCH_SIZE`.

find_exponential/src/main.rs
```diff
@@ -0,0 +1,197 @@
...
+                let res = panic::catch_unwind(|| {
```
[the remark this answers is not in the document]
oberien: I don't know how exactly the `threadpool` handled panicking in threads. `rayon` afaict removes the thread from the thread pool, reducing the number of concurrent threads. To prevent that behaviour, and also not depend on any implementation detail of the implementation of the used thread pool, I'm catching the panic.
The `process::exit` is probably wrong though and instead it should just be logged.

find_exponential/src/main.rs
```diff
@@ -0,0 +1,197 @@
...
+        }
+        return;
+    } else {
+        fuzz(num_cpus);
+    }
+}
+
+fn fuzz(num_cpus: usize) {
+    // get all literals from pulldown-cmark's source code
+    let walkdir = WalkDir::new("../src")
```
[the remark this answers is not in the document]
oberien: Currently I just extract all literals. I definitely agree that that's probably not the best, especially since there are a lot of printed / debug string literals. I don't think that a fully manual list is the correct answer though. While it might seem nice at first, changes to pulldown-cmark won't be reflected in the token list automatically and we'll need to manually keep it updated.

I may try just extracting literals from if-conditions and match arms. I'll compare those to a manually trimmed down list to see if all relevant tokens are included. Additionally, we probably want to add some literals manually after the automatic scan.

find_exponential/src/main.rs
```diff
@@ -0,0 +1,197 @@
...
+
+    let literals: Vec<_> = literals.into_iter()
+        .filter(|lit| !lit.contains(&b'\n'))
+        .filter(|lit| !lit.is_empty())
+        .collect();
+
+    let pool = ThreadPool::new(num_cpus);
+    let mut count = -(num_cpus as isize) * BATCH_SIZE as isize;
+    let start_time = Instant::now();
+    let (tx, rx) = mpsc::channel();
```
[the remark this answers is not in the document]
oberien: The way `threadpool` works is that you execute `pool.execute` for all tasks you want finished. However, those tasks are collected into a queue. In our case we can't just add all tasks into the queue. Instead, I only add a task to the queue when a thread is finished. I detect that a thread is finished by that thread sending a message through that channel.

I'll implement a buffer of 5 or so elements, which may improve throughput.

find_exponential/src/main.rs
oberien: As explained above, I perform two passes to verify if something is non-linear. The first one is the simple test, the second one the second test including outlier recalculation.

find_exponential/src/main.rs
oberien: I need to use the produced events of the parser. Otherwise the optimizer will realize that the events are unused, meaning the whole parser is unused, removing the whole parsing process. By introducing the black_box, it is guaranteed that the compiler won't optimize out the parser and event generation.

If I use `black_box(parser.count())`, due to `-C lto = fat`, the optimizer *might* be intelligent enough to eliminate `Event` generation and possibly other parts of the parser. That's why instead I opted into black-boxing the events directly, just as any normal user would.

find_exponential/src/scoring.rs
oberien: I'll add a comment.

Imagine the `(num_bytes, duration)`-pairs being an `(x, y)`-coordinate. This function calculates the slope between all possible pairs of coordinates, keeping only positive slopes (to keep A—B, but remove B­—A) (I should rewrite this as nested for-loop). Then I calculate the standard deviation of those slopes. In a perfect linear world, all slopes are equal and the standard deviation is zero. With non-linear growth, the standard deviation will be much higher as the slopes increase quadratically.

Taking into account all combinations of coordinates (A—B, A—C, B—C) instead of only adjacent ones (A—B, B—C) I reduce the impact of outliers as their impact will be smaller in the mean.

find_exponential/src/main.rs
```diff
@@ -0,0 +1,270 @@
...
+        literals::extract_literals_from_items(&mut literals, parsed.items);
+    }
+
+    let literals: Vec<_> = literals.into_iter()
+        .filter(|lit| !lit.contains(&b'\n'))
+        .filter(|lit| !lit.is_empty())
+        .collect();
+
+    let pool = ThreadPool::new(num_cpus);
+    let mut count = -(num_cpus as isize) * BATCH_SIZE as isize;
```
oberien: count only the number of batches, do the subtraction when printing

find_exponential/src/main.rs
```diff
@@ -0,0 +1,270 @@
...
+    for _ in 0..num_cpus {
+        tx.send(()).unwrap();
+    }
+    loop {
+        rx.recv().unwrap();
+        let mut literals = literals.clone();
+        let tx = tx.clone();
+        count += BATCH_SIZE as isize;
+        pool.execute(move || {
+            literals.shuffle(&mut rand::thread_rng());
```
oberien: XorShiftRNG is good enough

find_exponential/src/main.rs
```diff
@@ -0,0 +1,270 @@
...
+                .combinations(COMBINATIONS)
+                .take(BATCH_SIZE);
+            for comb in combs {
+                let res = panic::catch_unwind(|| {
+                    let concatenated = comb.into_iter().flatten().cloned().collect::<Vec<_>>();
+                    let s = String::from_utf8(concatenated).unwrap();
+                    test(&s);
+                });
+                if res.is_err() {
+                    ::std::process::exit(1);
```
oberien: print inputs instead of exiting.

find_exponential/src/main.rs
```diff
@@ -0,0 +1,270 @@
...
+            .expect(&format!("unable to parse file {:?}", file.path()));
+        literals::extract_literals_from_items(&mut literals, parsed.items);
+    }
+
+    let literals: Vec<_> = literals.into_iter()
+        .filter(|lit| !lit.contains(&b'\n'))
+        .filter(|lit| !lit.is_empty())
+        .collect();
+
+    let pool = ThreadPool::new(num_cpus);
```
oberien: No need for a thread-pool. I can just use crossbeam's mpmc channel and start `num_cpu` threads with a bounded channel of `num_cpu` elements.

find_exponential/src/main.rs
```diff
@@ -0,0 +1,270 @@
...
+    /// (points, number of bytes truncated from end reaching this behaviour)
+    ///
+    /// Possibly indicates weird behaviour in pulldown-cmark.
+    /// See <https://github.com/raphlinus/pulldown-cmark/issues/287> for an example.
+    HugeOutlier(Vec<[f64; 2]>, usize),
+}
+
+fn test_pattern(pattern: &str, sample_size: usize, recalculate_outliers: bool) -> PatternResult {
+    let s = pattern.repeat(NUM_BYTES / pattern.len());
+    let mut array = Vec::with_capacity(sample_size);
```
oberien: Make the caller pass `&mut [[f64; 2]; 10][..sample_size]` instead of using a `Vec`, no allocations required. Only allocate in `pearson_correlation`.

find_exponential/src/scoring.rs
```diff
@@ -0,0 +1,44 @@
...
+        println!("slopes: {}", slopes);
+    }
+
+    if slopes.is_empty() {
+        // Values were too small, we just assume linear behaviour.
+        // Otherwise we should see larger values.
+        return (0.0, false);
+    }
+
+    let stddev = slopes.central_moment(2).unwrap().sqrt();
```
oberien: Instead of using `Array1` (which allocates), we might get more throughput by manually implementing standard deviation on arrays / slices.

find_exponential/src/main.rs
oberien: Done in b42bbab

find_exponential/src/main.rs
oberien: Actually mpmc / spmc isn't needed here. The only thing sent through it are the literals, which don't change between thread iterations. A simple `std::sync::mpsc::channel` is enough to send feedback from the worker thread back to the main thread. The main thread then print the "status update" with the throughput.

I could optimize it even further by making the count-variable atomic, incrementing it from each thread and printing the status update in the worker threads. Then I can just join on all threads in the main thread, letting it sleep forever (unless there are uncaught panics in itertools / std). This is probably a premature optimization though…

oberien: I did the latter and used crossbeam scoped-threads. Done in 25dea8478cd43f6953b843523cf99abe288ed21e.

find_exponential/src/main.rs
oberien: Done in ca8e59e94b074039cf42fe36de5563f33ae5a7c5. Didn't seem to boost the throughput.

find_exponential/src/main.rs
oberien: Done in e3bb3a15b1795c1531dc31d62f86f3c37d9bdebf. I decided for `Xoshiro256+` because [it's better and faster](https://rust-random.github.io/book/guide-rngs.html#basic-pseudo-random-number-generators-prngs) than `XorShift`.

find_exponential/src/scoring.rs
oberien: I tested 3 different versions:
* `primitive`: Iterator adapters and combinators.
* `manual`: Using ndarray and `ArrayViewMut`.
* `stats`: Using `ndarray_stats::SummaryStatisticsExt::central_moment(2).unwrap().sqrt()`

```rust
test scoring::tests::bench_manual    ... bench:          44 ns/iter (+/- 6)
test scoring::tests::bench_primitive ... bench:          55 ns/iter (+/- 6)
test scoring::tests::bench_stats     ... bench:         306 ns/iter (+/- 67)
```

So I went with the `manual` approach in c99b189431bdd8848a47154b3590b9256229ce6f.

find_exponential/src/main.rs
oberien: Obsolete since 25dea84 and solved differently.

find_exponential/src/main.rs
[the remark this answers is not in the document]
oberien: I now use an array and only pass a mutable slice to all functions below. That completely avoids all allocations and should be a little clearer.

find_exponential/src/main.rs
```diff
@@ -0,0 +1,309 @@
...
+            if res.is_err() {
+                println!("Panic caught. Pattern: {:?}", pattern);
+            }
+        }
+
+        // print throughput
+
+        // TODO: I don't know which ordering we actually need, SeqCst is always correct, but may
+        // not actually be needed.
+        let batches_finished = num_batches_finished.fetch_add(1, Ordering::SeqCst);
```
marcusklaas: Relaxed ordering should be fine when you're just counting stuff.

oberien: According to [the nomicon Relaxed section](https://doc.rust-lang.org/nomicon/atomics.html#relaxed) you're right. I thought so as well, but wasn't 100% sure.

> For instance, incrementing a counter can be safely done by multiple threads using a relaxed `fetch_add` if you're not using the counter to synchronize any other accesses.

find_exponential/src/main.rs
marcusklaas: That's a good strategy!

find_exponential/src/scoring.rs
marcusklaas: Right — that is a sensible start I think. Though we may need to look at the actual estimated slope. If we estimate it is less than 1, it's probably not a performance issue.

find_exponential/src/main.rs
marcusklaas: Makes sense, good thinking!

find_exponential/src/main.rs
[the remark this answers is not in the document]
marcusklaas: This was my oversight. I thought you were catching panics caused by potential errors in forming utf-8 strings.

Catching panics is essential for reporting bugs in pulldown-cmark :+1:

marcusklaas:
This is looking a lot better. I really like that you're using `rand::distributions::Distribution`. That enables us to use different sampling strategies in the future really easily. Thanks also for setting up regression tests.

Remaining thoughts:
 - retesting needs to be super solid and should (almost) never report false positives. A new strategy that is not as sensitive to outliers and noise may be needed. I can help you with this if you want. I was thinking of bootstrapping much like criterion does;
 - it would be excellent to use your code to test for performance regressions in CI. Could we have a separate command that just runs the regression tests?
 - add random-length prefixes and suffixes. Not all scaling bugs we have seen so far consisted of simple repetitions. If we add short (0-2 tokens) prefixes and suffixes, we can search a larger space and fully capture the regression tests;
- have a more general name? I certainly hope that we find all exponential scaling bugs, but it seems your code should be capable of finding many other behaviors as well ;-)

find_exponential/src/main.rs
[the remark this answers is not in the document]
oberien: These comments are superseded by #302, which was fixed in 747a765.

marcusklaas:
Ok, one more thought on testing. Given the hypothesis that a pattern does not show superlinear behavior, the probability that a least squares linear fit will have slope greater than 1 is no greater than 0.5.

We can use this to control the false positive rate: by bootstrapping `n` times, the probability under this hypothesis that they are *all* greater than 1 is `1/2^n`, or about 1 in a billion for `n = 30`.

oberien:
Another possible idea is to perform two pearson correlations: One on `(x, y)`, and one on `(x, sqrt(y))`. If the former is larger (correlates better), then it's probably linear growth. If the latter is larger, it's probably non-linear growth.

These ways of scoring are pretty slow, so they should only be used for retesting and not during normal execution IMO.

On a different note: I seem to have found multiple cases where pulldown-cmark just enters an infinite loop or takes super long. While I'm still trying to verify those, we need some way to handle infinite loops in testing. Otherwise the thread will loop forever and not even report those cases back. We can't spawn a new thread for it, because you can't kill a thread. Starting a new process seems excessive, but we might be able to spawn one process per thread, which we feed patterns to through stdin. If the process takes too long, we kill it.

marcusklaas:
Good point on only doing these rigorous procedures during retesting. 

On your adapted scoring proposal: this just compares whether a linear or quadratic fit would be better. Ideally, we would be testing sublinear vs superlinear hypothesis. There is at least one known attack which has O(n^{3/2}) complexity, which may be caught by this. Scaling of O(n log(n)) will definitely not be caught.

The hangs are definitely a problem. I don't have a great solution for it. Spawning processes is one way around it, but having to go through stdin isn't very elegant.

oberien:
> retesting needs to be super solid and should (almost) never report false positives. A new strategy that is not as sensitive to outliers and noise may be needed. I can help you with this if you want. I was thinking of bootstrapping much like criterion does

I absolutely agree with you here. I first though Pearson's correlation coefficient would be good enough. But not only is it very bad with outliers, it's also super slow. That's why I though about the primitive slope solution. Unfortunately that's also not too good with outliers. That's why I implemented the outlier detection and recalculation.
I'm definitely not into statistics. As such any help here is highly appreciated. I could imagine adding a third pass after the current outlier removal, or replacing the current second one with a more sophisticated approach. In fact I'm not sure if we should only use a slow statistic during retesting, or if it's better if we directly do it as part of normal fuzzing, as it reduces the noise.
When designing a new statistic, you need to take all the edge-cases laid out in the documentation comments into account (like #287).

A new scoring function should also not depend on the size of the values. Currently the `slopes_stddev` function doesn't scale the input `(len, time)`-pairs. As such the resulting score changes from a fast machine to a slower one. On my PC with an i7-6700K a threshold of `20.0` is fine, while on the server where I run the fuzzer on 29 threads I need a threshold of `70.0` since the more compressed literals (it was `40.0` before). This should be fixed if we continue using that scoring function.

Re criterion, I found a paper which analyses testing / timing methods in noisy environments. I haven't read it, but it might be relevant here. They are even analysing Haskell's criterion and conclude that it's not that good: <https://arxiv.org/pdf/1608.04295.pdf>

> it would be excellent to use your code to test for performance regressions in CI. Could we have a separate command that just runs the regression tests?

I added a `--regressions` flag. As stated above the score of `slopes_stddev` is dependent on the machine's performance, so it might not make sense to perform regression testing on the CI before that is fixed.

> add random-length prefixes and suffixes. Not all scaling bugs we have seen so far consisted of simple repetitions. If we add short (0-2 tokens) prefixes and suffixes, we can search a larger space and fully capture the regression tests;

I thought about something like that already since I realized we can't test all regressions. A primitive approach would be to pass a function down the whole fuzzing stack, which is passed the final string to be tested and modifies it before being passed to pulldown-cmark.
We should create a function family consisting of possible behaviour we want to include when fuzzing. For each pattern we select a function from that family. The function must be deterministic such that we get the same behaviour for all samples in all passes.
We should have:

* Truncation of start of a random length smaller than the pattern size
* Truncation of end of a random length smaller than the pattern size
* Adding a random amount of random literals to the start of the string
* Adding a random amount of random literals to the end of the string

A refactor of the fuzzing stack into a struct with methods might help here as the function could be stored in the fuzzer struct instead of being required as argument in every single function.

> have a more general name? I certainly hope that we find all exponential scaling bugs, but it seems your code should be capable of finding many other superlinear behaviors as well ;-)

I renamed it to `fuzzer`, which is what it is: It's a fuzzer finding non-linear behaviour and even some panics. (I can't rename the branch because that would make this PR invalid)

---

So we're left with the following TODOs:

- [ ] Better scoring function / outlier detection / (first pass?) / second pass / third pass?
- [ ] Normalize the score of `slope_stddev` to not be dependent on the machine's performance
- [ ] Input modification function / random prefixes / suffixes / truncations
- [ ] Handle infinite loops / very long parsing times (I'll get back to that tomorrow)

Unfortunately I won't be able to continue working on this for much longer. I think the fuzzer is in a good state code- and documentation-wise such that someone else should be able to take over easily. I should have fixed all of your annotations and comments and documented and commented the important parts. This fuzzer was intended to be a short and simple one-evening project for me, and it ended up consuming the better of a week now :) Don't get me wrong, I'd like to continue working on this and I can definitely perform code review, concept discussions and one or the other small change. I can keep running it on my server on 29 cores if you'd like. I just won't have much time to actively continue working on this in the next 2 months. (tldr: this was yet another procrastination project before I continue working on my thesis ;) )

I'll definitely finish the handling of infinite loops, but I won't tackle the input modification function and statistics parts.

oberien:
For time measurement we should use a timer which measures CPU-user-time like `getitimer` (or a similar variant of that function that only takes into account the user time of the current thread instead of all threads). That should reduce the jitter caused by rescheduling. I don't know if there is a function which gives you the cpu user time of a single thread. This is not a problem if we have worker-processes instead of worker-threads and the processes report the time taken back to the parent.

For handling infinite loops, I currently have the following ideas:

1. Detect if a thread has panicked. In that case print the pattern that thread has just worked on with an according message. After that, restart the whole process from the bash script (`run` or `retest-output`). Disadvantage is that it'll be hard to print the pattern that resulted in the loop. We'd need a synchronized way for the worker threads to write the pattern to the thread that checks if a thread hangs. Introducing that thread also means more scheduling overhead.
2. Spawning worker processes from each worker thread. The fuzzer will have a new argument flag `--worker`, which indicates that the binary should be executed as worker process instead of parent / master. We then use `Command` and `Child` to spawn worker processes. We can use (2.1) stdin/stdout or (2.2) the `ipc-channel` crate (*nix only). (2.1) will have more overhead, while (2.2) is *nix specific but probably provides more throughput compared to (2.1). It uses `serde` for serialization and deserialization of the data.
3. Manually `fork`ing worker-processes. That way we can write our own shared memory mapping as an atomic queue and have less overhead compared to (2.2). Disadvantage is that we either need to `exec` following the fork, or we're limited in the function we can call (because after a fork only some syscalls are allowed). That shouldn't matter too much as pulldown-cmark is only userland code, but adding a syscall in pulldown-cmark (for whatever reason) will make the fuzzer produce UB.
4. Similar to (3), but perform the `fork` for every pattern. I did some benchmarks and it seems like I can get around 1000 fork / wait calls per second per thread. That might be fast enough considering we currently have less than 50 patterns per second per thread. Further testing might be needed.
5. There is a way to send unix signals to a thread instead of to the whole process. We can use this to trigger a `SIGALARM` after our required timeout. We can set an alarm to be triggered after our timeout, then run pulldown-cmark. If pulldown-cmark finishes first, we delete the timer. Otherwise the timer will trigger a `SIGALARM` which is sent to only our thread. In the signal handler we panic. That panic unwinds the stack until we reach the thread's function. We can wrap the pulldown-cmark execution in a `catch_unwind` to recover from the sigalarm handler panic. [Proof of concept code](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=4474e5c316b2e4af532aac770d4009ff).  
  We do unwind through C code, in particular the signal producing functions. That's fine on all platforms that use dwarf-unwinding information (which are all unix platforms). This is because if dwarf-unwinding information is used, each function's personality function unwinds its function's stack frame. All is well as long as the C implementation is well-behaved wrt stack unwinding (which we can expect from linux and freebsd, which are the only two platforms the fuzzer is probably going to be run on).  
  This solution obviously only works on linux and freebsd, as those are the two special cases I implemented.

If the fuzzer should be run on windows and we're using linux-specific features, we might be able to run it in WSL 1 or soon on WSL 2.

marcusklaas:
Given that you have a thesis to write, I would try for a solution that is easily implemented. Managing processes is almost always lot of work as your options show.

Why not consider this: construct a shared array of `(pattern, timestamps)` with an entry for every worker. Once a worker starts investigating a new pattern, have it set that pattern and the current timestamp in its slot. Then, have a separate thread walk through the array in set intervals to check that no timestamp is older than a certain threshold. If it does find one, have it print the pattern and kill the entire program. We will find at most 1 hang per program run, but those shouldn't be too common hopefully and every one we find is a big win in itself.

marcusklaas:
Yeah, synchronization will be needed, but since every thread writes to its own value and that value is only checked for reading infrequently, it will generally be uncontested and fast. I would be very surprised if it'd show up in the profile above 0.1%.

That's mostly the quick and dirty solution though. All of your options seem reasonable. I am neither experienced enough or knowledgeable enough about threads/ processes to make an informed decision between them. Dropping support for non-unix/ non-posix systems should be ok, since this will only be run by a small set of people/ services. Go for the one that makes sense to you :-)

oberien:
Regarding scaling / normalization in `slope_stddev` I tested a bit today. My results:

We should scale the data such that slopes won't depend on the CPU's performance. We can scale the x and y coordinates by dividing them by the first x and y coordinate. However, lines with larger slopes will have larger slope values. Those will result in a larger standard deviation for outliers / inconsistencies. If we scale the slopes by dividing them by the first slope, outliers have a much higher impact. In fact a single outlier has more impact than a perfect quadratic line in the standard deviation. As such no scaling is performed so far and the ACCEPTANCE_STDDEV must be tweaked manually.

oberien:
Some stats. I let it run for 18 hours on 29 cores, with the following results:

* 406 panics encountered
* 7 thread timeouts were triggered
* 25868 false positives
* 1870 true positives / huge outliers
* 93% false positive rate, which is not good, but acceptable
  * First `./retest-output` run filtered out all false positives except for ≈150. Took about 1h.
  * Second run filtered out ≈100. Took about 45min.
  * Third run filtered out ≈50. Took 37min.
* Some patterns recognized as false positives were actually true positives. Don't have a number on this one, but should be little and redundant enough to not affect us too much.
  * `./retest-output` is a fixpoint iteration, approaching the final number of panics, timeouts and true positives.

Considering the amount of panics, I should write another short (python) script, which analyzes the backtraces of the panics to group them into groups producing the same backtrace.
Considering the huge amount of true positives, most (if not all) of them are variations of the currently known ones. Maybe I should stop fuzzing until those are fixed.

marcusklaas:
Woah, that is a large number of hits. It sounds likely that many of them are simply triggering the same behaviour though. It'll be much easier to identify new issues once we fix the existing ones. Unfortunately, there aren't many contributors to review PRs, so progress is slow. Maybe you could have look at https://github.com/raphlinus/pulldown-cmark/pull/290 and https://github.com/raphlinus/pulldown-cmark/pull/281? Those should fix a bunch, and then we'd be able to rebase this PR as well :-)

oberien:
I automated the retesting process some more. I wanted to perform the "fixpoint iteration" until the true positives don't change anymore. But apparently the more retesting I perform, the more true positives are removed. In fact even after 6 rounds of retesting there were still patterns removed (with one round taking over an hour). Thus, I've changed it to only test 3 rounds, which should be good enough here.

48 more hours of fuzzing on the server:
* Total panics: 725
* Total thread timeouts: 15
* Total joined threads: 0
* Total UTF8 Errors: 0
* Total false positives: 39961
* Total true positives: 3282

Panics:
* 19 unreproducible
* Known ones:
    * 8 × #289
    * 18 × #313 
    * 116 × #305 
    * 127 × #304 
    * 132 × #306 
    * 300 × #303 
* Previously unknown ones:
    * 3 × #319
    * 1 × #320

Timeouts:
* all false positive timeouts, but most were true positives wrt weird parsing behaviour / non-linear growth.

True Positives:
* 2026 × #307
* 701 × #284
* 635 × #312 
* 3 × #287 

With this I think this PR is ready for review and conclusion. I'll stop fuzzing until the above issues are fixed, because it doesn't make sense to only verify known issues :)

marcusklaas:
Thanks for rebasing! I have not forgotten about this. I will try to do a comprehensive review before the week is out.

fuzzer/src/main.rs
```diff
@@ -0,0 +1,443 @@
+//! This is a fuzzer for pulldown-cmark, which tries to detect super-linear parsing behaviour.
+//! Additionally it can detect panicks of pulldown-cmark.
+//!
+//! The general idea is to extract relevant literals from the pulldown-cmark source code
+//! (see the `literals` module).
+//! Short random combinations of literals are generated (like concatenating 6 random literals),
+//! which are called patterns.
+//! Those patterns are repeated until they reach a byte-length of `NUM_BYTES`.
+//! We iterate over subslices of the pattern to get `SAMPLE_SIZE` uniformly distributed
+//! (len, time)-pairs.
```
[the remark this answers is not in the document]
oberien: You're right that I used the wrong word here. I meant evenly.

fuzzer/src/main.rs
```diff
@@ -0,0 +1,443 @@
...
+//! In the second pass, we double the `SAMPLE_SIZE`, because more points make for higher precision
+//! in the scoring functions.
+//! Additionally, we assume that the (len, time)-pairs interpreted as coordinates on a graph  are
+//! strictly monotonously increasing.
+//! If we encounter a time-value, which is larger than the next time value, that'll probably be an
+//! outlier.
+//! As such, if we encounter such a value, we'll re-time it.
+//! Unfortunately, again, we have an edge-case.
+//! When testing substrings of the input, we might not always end up after the whole pattern,
+//! but instead in the middle of it.
```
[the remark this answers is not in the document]
oberien: I guess we *might* avoid that problem this way, but at the same time I think that it is valuable to find these kinds of weird parsing behaviour like #287 and #312. If you think that they add no / not enough value, it'd also be possible to change it to align to pattern-length instead of character-length.

fuzzer/src/main.rs
```diff
@@ -0,0 +1,443 @@
...
+const BATCH_SIZE: usize = 10_000;
+/// Length of number of literals per combination.
+///
+/// For example given the literals `a` and `b`, with a value of 4, `aaab` is a possible pattern.
+const COMBINATIONS: usize = 6;
+/// If parsing takes longer than this many milliseconds, it is assumed to be non-linear and
+/// parsing is aborted.
+const MAX_MILLIS: u128 = 500;
+/// Byte-length of maximum pattern repetitions used as input to pulldown-cmark.
+const NUM_BYTES: usize = 256*1024;
```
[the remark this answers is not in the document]
oberien: The problem with using smaller numbers are cases where parsing is in the range of nanoseconds. With those inputs, the whole parsing takes so little time, that in some cases the measured times appear non-related. With larger inputs the required time is large enough to not be affected that much by jitter.

However, the last time I verified that behaviour was before we implemented #302, so the behaviour might have changed since then due to the more "hardcore" literals used. It might also make sense to add 3 passes with the first one only working on 8kB, the second one filtering out the jitter, and the third one verifying the non-linear growth with more data points. But then again a different scoring function might be able to achieve better results with a single pass.

Anyway, this is something that is tested the easiest when all currently known superlinear growth is fixed.

fuzzer/src/main.rs
```diff
@@ -0,0 +1,443 @@
...
+/// Number of samples per pattern tested. Higher number increases precision, but reduces the throughput.
+const SAMPLE_SIZE: usize = 5;
```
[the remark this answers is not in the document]
oberien: The higher the sample size, the longer the analysis takes. It takes O(SAMPLE_SIZE²) due to the even distribution of measurement input regarding the input length. 5 for the first pass and 2*5 for the second pass seemed ok, especially because we verify all outputs multiple times afterwards with the `./retest-output` script. Obviously better tweaking of those numbers will probably achieve better results and a lower false-positive rate, but the current numbers seem good enough for my machine.

There are probably lots of easy ways to improve the mechanics used for superlinear growth detection. But I didn't want to write a paper about this (yet) ;)

fuzzer/src/main.rs
```diff
@@ -0,0 +1,443 @@
...
+        return score;
+    }
+}
+
+#[derive(Debug)]
+enum PatternResult {
+    /// Probably linear growth (score)
+    Linear(f64),
+    /// Probably non-linear growth (score)
+    NonLinear(f64),
```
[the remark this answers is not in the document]
oberien: It is actually `NonLinear`. The stddev-slopes test will probably also detect sub-linear growth. (Even though I don't think that it's possible to write a parser with o(n) :) ).

fuzzer/src/main.rs
```diff
@@ -0,0 +1,443 @@
...
+
+/// Start fuzzing on given number of threads in parallel.
+fn fuzz(num_cpus: usize) {
+    let literals = literals::get();
+
+    let num_batches_finished = AtomicU64::new(0);
+    let num_batches_finished = &num_batches_finished;
+    let mut rng = Xoshiro256Plus::from_rng(&mut rand::thread_rng()).unwrap();
+    let start_time = Instant::now();
+    let mut pattern_times = Vec::with_capacity(num_cpus);
```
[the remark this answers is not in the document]
oberien: Unfortunately ``the trait bound `std::sync::Mutex<(std::string::String, std::time::Instant)>: std::clone::Clone` is not satisfied`` :( That's why I wrote it this way.

fuzzer/src/main.rs
```diff
@@ -0,0 +1,443 @@
...
+            })
+        }).collect();
+
+        // timeout thread
+        threads.push(s.spawn(|_| {
+            loop {
+                std::thread::sleep(Duration::from_millis(MAX_MILLIS as u64 * 10));
+                for pattern_time in &pattern_times {
+                    let (pattern, time) = &*pattern_time.lock().unwrap();
+                    if time.elapsed().as_millis() > MAX_MILLIS * 10 {
```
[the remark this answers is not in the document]
oberien: I added a section in the module documentation to explain this.

fuzzer/src/main.rs
```diff
@@ -0,0 +1,443 @@
...
+                        // We print the pattern and exec ourselves again to not infinite-loop a thread.
+                        // See <https://github.com/raphlinus/pulldown-cmark/pull/286#issuecomment-490315582>
+                        // for more information.
+                        println!(
+                            "Thread timeout triggered (thread took too long) for Pattern: {:?}\n\
+                            Restarting process...",
+                            pattern,
+                        );
+                        let args: Vec<_> = env::args().collect();
+                        Command::new(&args[0])
```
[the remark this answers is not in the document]
oberien: > A call to any exec function from a process with more than one thread shall result in all threads being terminated and the new executable image being loaded and executed. No destructor functions or cleanup handlers shall be called.
> -- [Posix Specification](https://pubs.opengroup.org/onlinepubs/9699919799/functions/exec.html)

fuzzer/src/main.rs
```diff
@@ -0,0 +1,443 @@
...
+    let mut pattern_times = Vec::with_capacity(num_cpus);
+    for _ in 0..num_cpus {
+        pattern_times.push(Mutex::new((String::new(), Instant::now())));
+    }
+
+    // start threads
+    thread::scope(|s| {
+        // worker threads
+        let mut threads: Vec<_> = (0..num_cpus).map(|i| {
+            let literals = literals.clone();
```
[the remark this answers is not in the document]
oberien: Good catch. This was still from a time where I shuffled the vector and used the combinations iterator adapter to iterate over one batch of combinations per thread. Mutating (shuffling) the vector required mutable access.

fuzzer/src/main.rs
```diff
@@ -0,0 +1,443 @@
...
+        let batches_finished = num_batches_finished.fetch_add(1, Ordering::Relaxed);
+        // fetch_add returns the old value
+        let batches_finished = batches_finished + 1;
+        let patterns_finished = batches_finished * BATCH_SIZE as u64;
+        let elapsed_secs = start_time.elapsed().as_secs();
+
+        // if for some reason we are super-fast, we must avoid div-by-zero
+        if elapsed_secs > 0 {
+            println!(
+                "{:<20} throughput: {} / s",
```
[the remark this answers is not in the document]
oberien: The throughput is only printed once every few minutes, so a difference of a few patters per second doesn't really matter. The zero-check is because sometimes one or two threads would end up finishing the first batch pretty much instantaneously (never figured out why). The same could happen with millis. Not panicking here is very important as any panic in a thread that's not within `test_catch_unwind` just straight up kills the whole thread. Any chance of that function panicking might result in way lower throughput when running the fuzzer on the server for several days.

fuzzer/src/scoring.rs
```diff
@@ -0,0 +1,67 @@
...
+/// If the time increase is perfectly linear, the slopes of the line running through each two
+/// adjacent points are all equal.
+/// In fact given any pair of points from the array, the slope will be the same.
+/// If all slopes are the same, the standard deviation of all of those slopes will be zero.
+///
+/// However, in the real world there are inconsistencies, deviations, rounding and outliers.
+/// As such, the points won't be on a perfect line.
+/// Nevertheless, if it's linear, the slopes will be close enough to each other, that small outliers
+/// won't increase the standard deviation too much.
+/// The more points we add and the more slopes we add, the more resilient to outliers the result
```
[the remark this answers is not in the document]
oberien: I'm not 100% sure as well, but with increasing the sample size constant I was experiencing less false positives (at the cost of less throughput). I explained this with the following observation. There are `n * (n-1) / 2` slopes in total. If there is one outlier, `n-1` of those are "bad". That means that we have `(n * (n-1) / 2)  -  (n-1)` "good" slopes. The number of "bad" slopes is in O(n) while the number of "good" slopes is in O(n²). As such increasing `n` will increase the number of "good" slopes more than the number of "bad" slopes, reducing the impact of a single outlier.

oberien:
> Job well done! +1

Thanks :)

> I think it'd be very useful to have a short readme file.

Done. I described all scripts in the project with their respective functionality, how to run the fuzzer and how to retest the output correctly.

> I'm not 100% sold on using slope standard deviations as a scoring function, particularly because the thresholds need to be tweaked manually.

There are a lot of things that can still be improved, I'll open issues for them once this PR is merged. We already discussed some earlier in https://github.com/raphlinus/pulldown-cmark/pull/286#issuecomment-489880949. Additionally there are the following possible improvements (which I'll open issues for as well, these are more notes to myself so that I don't forget about them):
* Joined threads doesn't actually work if the thread panicking is not the first thread of the threads-vector.
* structopt for argument parsing
* core pinning

> I'm mostly curious how the program restart works when a timeout situation is encountered. If you could elaborate your thinking a bit here, I'd be very grateful.

I `exec` the current binary another time with the same environment and arguments. Usually when you `fork` you follow with an `exec`. However I can just `exec` directly without a fork to semantically replace the current process with the `exec`'d one.  
As stated in an earlier comment (https://github.com/raphlinus/pulldown-cmark/pull/286#discussion_r285323666), the posix API says that all threads should be terminated before the "new" process is spawned.

> Oh, and let's add the regression run to the CI script. That way we should never regress on known adversarial input.

I changed the regression testing code to report any regression back as exit code 1. If no non-linear behaviour could be found, it'll exit with exit code 0. I hope that azure handles this like the CI services I've used previously and abort on a non-zero exit code. I've also added the regression test to the azure config file.

> Maybe it's good if raphlinus has a quick look as well?

Having more reviews should never hurt :)

marcusklaas:
Thanks for your update and comments. I did not know about the posix behavior for spawning new processes without forks. TIL!

You have convinced me of most issues that they are either ok or can wait until later PRs. The one remaining change I really believe we should do is only testing proper multiples of patterns. Indeed, we may not spot some panics that we may otherwise see, but I think that shouldn't be the primary focus of the fuzzer anyway. By only testing proper multiples, we won't have to do the outlier detection stuff and it should make the results much more interpretable. I strongly suspect behavior like in https://github.com/raphlinus/pulldown-cmark/issues/312 is caused by strings that are not exact multiples of a pattern.

We can always use a 'classic' fuzzer like afl to trace panics.

r+ from me with that change!

oberien:
> By only testing proper multiples, we won't have to do the outlier detection stuff and it should make the results much more interpretable.

This is somewhat true. We can simplify the outlier detection again to just always retest from the last value. We probably can't completely remove outlier detection from the fuzzer due to variations in time measurement. Consistent outliers will then be "detected" by the infinite loop detection and result in a restart of the whole process.

> By only testing proper multiples, we won't have to do the outlier detection stuff and it should make the results much more interpretable.

Yes and no. While the timings might improve slightly, we lose all "weird parsing behaviour". I'm not sure if it's fine to just ignore weird parsing behaviour, or if there is something that could be done to improve cases where leaving out one glyph of the input will dramatically increase parsing time.  
Additionally, if we implement truncation / adding literals of random lengths as discussed in https://github.com/raphlinus/pulldown-cmark/pull/286#issuecomment-489880949, we might require proper outlier detection somewhat similar to the one currently implemented.

However, if you feel like it's not needed right now, I'll gladly remove it, which will be easy enough (delete ~20 lines of code and remove the documentation regarding outlier detection) :)

marcusklaas:
Yeah, I believe it's better to not also try to catch unusual parsing behavior. Whenever we detect unusual timings, it won't be as clear what is going on, superlinear behavior or bad parsing behavior. It's probably to have this tool focus on the former.

oberien:
I changed it :)

marcusklaas:
Excellent. Thanks for your hard work!

PR #297, opened 2019-05-05
Update html5ever to 0.23

ignatenkobrain:
Signed-off-by: Igor Gnatenko <i.gnatenko.brain@gmail.com>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -184,12 +184,12 @@ dependencies = [
 
 [[package]]
 name = "html5ever"
-version = "0.22.5"
+version = "0.23.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "log 0.4.6 (registry+https://github.com/rust-lang/crates.io-index)",
  "mac 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
- "markup5ever 0.7.5 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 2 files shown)


marcusklaas:
There are some test failures with this update. It appears some behavior changed. We may need to adjust our usage to accomodate. Is this something you could look into?

ignatenkobrain:
Oh, I did simple `cargo test` on my laptop and it worked without any problems...

I'm not expert in this html/markup5ever things.. I would appreciate if you could help me with that.

ignatenkobrain:
So I checked and it happens due to this commit: https://github.com/servo/html5ever/commit/8bf65ef61daf612a3b4d7e6007fde7297d6210eb

marcusklaas:
Did you identify that commit using a git-bisect? It's not immediately clear to me how that influences its behavior...

ignatenkobrain:
Yes.

marcusklaas:
Ok, so the commit you linked actually had all the answers in its comments :grimacing:

Apparently `RcDom` drops all its nodes now when it is dropped itself. The solution is trivial: only drop it when we are done. Making `normalize_dom` accept a reference [here](https://github.com/raphlinus/pulldown-cmark/blob/master/tests/lib.rs#L73) does the trick.

marcusklaas:
html5ever was upgrading in the 0.5.1 release. Thanks for the PR though :-)

Issue #308, opened 2019-05-06
Panic "index out of bounds"


Draphar:
I recently published my crate [test-exec](https://github.com/Draphar/test-exec) and all CI builds were successful. So, I was pretty surprised when docs.rs failed to build it, and it turns out when that pulldown_cmark panics when trying to document it, but only on Nightly. Steps to reproduce (only the compiler error):

```sh
git clone https://github.com/Draphar/test-exec
cd test-exec
cargo +nightly doc
```

Error:
```
thread 'rustc' panicked at 'index out of bounds: the len is 2 but the index is 2', /cargo/registry/src/github.com-1ecc6299db9ec823/pulldown-cmark-0.4.1/src/parse.rs:1105:17
note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.

error: internal compiler error: unexpected panic

note: the compiler unexpectedly panicked. this is a bug.

note: we would appreciate a bug report: https://github.com/rust-lang/rust/blob/master/CONTRIBUTING.md#bug-reports

note: rustc 1.36.0-nightly (e305df184 2019-04-24) running on x86_64-unknown-linux-gnu

error: Could not document `test-exec`.

Caused by:
  process didn't exit successfully: `rustdoc --edition=2018 --crate-name test_exec src/lib.rs --color always -o /home/joshua/Rust/test-exec/target/doc -L dependency=/home/joshua/Rust/test-exec/target/debug/deps --extern wait_timeout=/home/joshua/Rust/test-exec/target/debug/deps/libwait_timeout-e992bdb7635958bf.rmeta` (exit code: 1)
```

[docs.rs error](https://docs.rs/crate/test-exec/0.1.0/builds/158337)

marcusklaas:
Woops, that's no good. Thanks for reporting this. We'll have a look.

Draphar:
I just [updated](https://github.com/Draphar/test-exec/commit/a33e5cd41439c01171d839ff87d8f507bfd4b70e) the repository "fixed" the error. Here is the code to reproduce the error now:
```sh
git clone https://github.com/Draphar/test-exec
cd test-exec
git reset --hard e6865dfba602446c6761fde7c72436a204daa27c
cargo +nightly doc
```

dimbleby:
Suspect that the issue is a lone `#` somewhere, or similar.

At v0.5.0, the following testcase fails:

```rust
    #[test]
    fn single_octothorpe() {
        // dont crash
[...]

PR #310, opened 2019-05-09
Add a testcase for #308

dimbleby:
As noted in #308, this seems already to be fixed in master.  Still, might as well have the testcase to prevent regression.


marcusklaas:
Thanks!

Issue #319, opened 2019-05-12
Panic on '|\r-]([^' * 2 (attempt to subtract with overflow)


oberien:
```
python -c "print('|\r-]([^' * 2)" | RUST_BACKTRACE=1 cargo run -- -TFSL
python -c "print('|\r:][^;' * 2)" | RUST_BACKTRACE=1 cargo run -- -TFSL
python -c "print('|\r\r=][^' * 2)" | RUST_BACKTRACE=1 cargo run -- -TFSL
```

```
thread 'main' panicked at 'attempt to subtract with overflow', src/parse.rs:1300:9
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
             at src/libstd/sys/unix/backtrace/tracing/gcc_s.rs:39
   1: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:71
   2: std::panicking::default_hook::{{closure}}
             at src/libstd/sys_common/backtrace.rs:59
             at src/libstd/panicking.rs:197
   3: std::panicking::default_hook
             at src/libstd/panicking.rs:211
   4: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:474
   5: std::panicking::continue_panic_fmt
             at src/libstd/panicking.rs:381
   6: rust_begin_unwind
             at src/libstd/panicking.rs:308
   7: core::panicking::panic_fmt
             at src/libcore/panicking.rs:85
   8: core::panicking::panic
             at src/libcore/panicking.rs:49
   9: pulldown_cmark::parse::count_header_cols
             at src/parse.rs:1300
  10: pulldown_cmark::parse::FirstPass::parse_line::{{closure}}
             at src/parse.rs:593
  11: pulldown_cmark::parse::scalar_iterate_special_bytes
             at src/parse.rs:2237
  12: pulldown_cmark::parse::iterate_special_bytes
             at src/parse.rs:2230
  13: pulldown_cmark::parse::FirstPass::parse_line
             at src/parse.rs:551
  14: pulldown_cmark::parse::FirstPass::parse_paragraph
             at src/parse.rs:494
  15: pulldown_cmark::parse::FirstPass::parse_block
             at src/parse.rs:393
  16: pulldown_cmark::parse::FirstPass::run
             at src/parse.rs:238
  17: pulldown_cmark::parse::Parser::new_with_broken_link_callback
             at src/parse.rs:1763
[...]

Issue #318, opened 2019-05-10
tables versus hard breaks


mity:
When tables extensions are enabled, any extra trailing spaces on any line in the following input should not cause that we treat it as a paragraph with a hard-break on that line.

Yet, if you place two spaces at the end of the 1st line, it does happen.

So this is ok:
```
$ printf 'foo|bar\n---|---\nfoo|bar' | ./pulldown-cmark -T
<table><thead><tr><th>foo</th><th>bar</th></tr></thead><tbody>
<tr><td>foo</td><td>bar</td></tr>
</tbody></table>
```

but the following is not:
```
$ printf 'foo|bar  \n---|---\nfoo|bar' | ./pulldown-cmark -T
<p>foo|bar<br />
---|---
foo|bar</p>
```
(I would expect the same output as before.)

Ditto if the hard-break is denoted with a backslash:
```
$ printf 'foo|bar\\\n---|---\nfoo|bar' | ./pulldown-cmark -T
<p>foo|bar<br />
---|---
foo|bar</p>
```
(I would expect the backslash is seen as a literal backslash, which becomes part of the respective table cell.)

**Rationale:** Hard-break is an inline, while table is a block. See https://spec.commonmark.org/0.29/#precedence

Issue #307, opened 2019-05-06
Non-linear behaviour when parsing "*~~\u{a0}"


oberien:
Also produced by `~~_a`.

[plot](https://www.wolframalpha.com/input/?i=plot+%5B(26215.0,+14999087.0),+(52428.0,+59359636.0),+(78642.0,+136215463.0),+(104856.0,+244850386.0),+(131070.0,+394244799.0),+(157285.0,+556634061.0),+(183498.0,+761634573.0),+(209712.0,1007746879.0)%5D)

marcusklaas:
I can't reproduce this behavior with either example. What branch did you produce this on?

oberien:
I fuzzed on #281. Whenever strikethrough (must be enabled) occurs next to inline-formatting (like italic), non-linear behaviour can be observed. Did you test without strikethrough enabled? In that case it doesn't happen.

marcusklaas:
Aaah, indeed I did not have the extension enabled! I'm getting quadratic behavior now as well. My bad!  :man_facepalming:

Issue #296, opened 2019-05-04
Non-linear behavior when parsing "[](<..."


mity:
```
$ time python -c 'print("[](<" * 10000)' | ./pulldown-cmark >/dev/null
real    0m1.600s
user    0m1.547s
sys     0m0.031s

$ time python -c 'print("[](<" * 20000)' | ./pulldown-cmark >/dev/null
real    0m6.134s
user    0m6.063s
sys     0m0.078s

$ time python -c 'print("[](<" * 40000)' | ./pulldown-cmark >/dev/null
real    0m24.085s
user    0m24.016s
sys     0m0.047s
```

mity:
Likely caused by the fact that the parser accepts unescaped `<` in the `<>`-type of link destination:

```
$ printf '[](<<>)' | ./pulldown-cmark
<p><a href="%3C"></a></p>
```

That is against the spec:

> A link destination consists of either
> * a sequence of zero or more characters between an opening < and a closing > **that contains no line breaks or unescaped < or > characters**, or
>
> * a nonempty sequence of characters that does not start with <, does not include ASCII space or control characters, and includes parentheses only if (a) they are backslash-escaped or (b) they are part of a balanced pair of unescaped parentheses. (Implementations may impose limits on parentheses nesting to avoid performance issues, but at least three levels of nesting should be supported.)

(Emphasis mine)

oberien:
I can't verify that behaviour on #281. There I get [this graph](https://www.wolframalpha.com/input/?i=plot+%5B(26214.0,+1607501.0),+(52428.0,+2992827.0),+(78642.0,+4133549.0),+(104856.0,+5571305.0),+(131070.0,+7016048.0),+(157284.0,+8316662.0),+(183498.0,+9883032.0),+(209712.0,+11065276.0)%5D) for 200KB and [this graph](https://www.wolframalpha.com/input/?i=plot+%5B(1677721.0,+89575815.0),+(3355442.0,+177112216.0),+(5033163.0,+269324236.0),+(6710884.0,+357378271.0),+(8388605.0,+478089843.0),+(10066326.0,+533949863.0),+(11744047.0,+633455160.0)%5D) for 11MB of input data consisting only of that pattern. Both look pretty linear to me :)

oberien:
This might be related to #282, which opens an autolink instead of the inline link used here.

PR #325, opened 2019-05-15
Numerous parsing fixes

### LINKED ISSUE #314: Backticks and escaping
Backslash escaping should be suppressed **inside** code span. Backtick, which would otherwise be an opener of a code span, still may be escaped.

Input:
```
\`x`
```

Expected output:
```
<p>`x`</p>
```

Actual output:
```
<p><code>x</code></p>
```

marcusklaas:
Fixes #314, #315 and #317.

Many thanks to @mity for reporting these issues with test cases. Such work makes fixing these things much easier! :bowing_man:


marcusklaas:
Also fixes https://github.com/raphlinus/pulldown-cmark/issues/294 now.

marcusklaas:
Also closes #296 and #304.

src/parse.rs
```diff
@@ -651,10 +647,11 @@ impl<'a> FirstPass<'a> {
                 b'`' => {
                     self.tree.append_text(begin_text, ix);
                     let count = 1 + scan_ch_repeat(&bytes[ix+1..], b'`');
+                    let preceeded_by_backslash = ix > 0 && bytes[ix - 1] == b'\\';
                     self.tree.append(Item {
                         start: ix,
                         end: ix + count,
-                        body: ItemBody::MaybeCode(count),
+                        body: ItemBody::MaybeCode(count, preceeded_by_backslash),
```
marcusklaas: Hmmmm, this is probably not correct. `` \\`foo` `` may now incorrectly render to ``<p>\`foo`</p>`` where it should create a proper span in this case.

mity: Also be aware of funny case as this one:
```
\``foo``bar`
```

Note the 1st backtick is escaped, but the 2nd still makes beginning of a code span - of length 1 in this case so the closer is after the `bar`, not before it.

marcusklaas:
Yup, it's very tricky. I *think* we managed such situations ok, but after seeing what other (trivial) tests we failed, I'm not sure about anything any more ;-)

Thanks again for your valuable input. I will check your example as soon as I am back on my machine.

src/parse.rs
marcusklaas: It produces `<p><code>foo``bar</code></p>` now. So the code span is alright, but there should be a text node with a single backtick preceding it, right?

marcusklaas:
` ``` ` renders to `<code></code>` in both pulldown-cmark as the reference implementation. That makes no sense to me :thinking:

marcusklaas:
Ok, the backtick issues should be fixed now. Also https://github.com/raphlinus/pulldown-cmark/issues/300.

mity:
> `` ``` `` renders to `<code></code>` in both pulldown-cmark as the reference implementation. That makes no sense to me 🤔

It's required by the specs. See https://spec.commonmark.org/0.29/#example-98. 

If the enclosing block ends, it enforces the end of the unfinished fenced code block. Here the enclosing block is the document as a whole.

marcusklaas:
Oooh wait those are code *blocks*, not inline code. It makes sense now! :sweat_smile: 

Thanks for the reference!

src/parse.rs
mity: Yes.

marcusklaas:
Fixes https://github.com/raphlinus/pulldown-cmark/issues/293 now.

marcusklaas:
Addresses https://github.com/raphlinus/pulldown-cmark/issues/320.

marcusklaas:
Fixes https://github.com/raphlinus/pulldown-cmark/issues/318.

marcusklaas:
Addresses https://github.com/raphlinus/pulldown-cmark/issues/283.

marcusklaas:
Hmmm, performance does seem to slip a bit with these changes (2-3%).. Will have to dig a bit to see what is causing this..

marcusklaas:
Closes https://github.com/raphlinus/pulldown-cmark/issues/307.

marcusklaas:
Fixes https://github.com/raphlinus/pulldown-cmark/issues/328.

marcusklaas:
Fixes https://github.com/raphlinus/pulldown-cmark/issues/303 and #313.

marcusklaas:
Fixes https://github.com/raphlinus/pulldown-cmark/issues/319.

marcusklaas:
Addresses https://github.com/raphlinus/pulldown-cmark/issues/289.

marcusklaas:
Fixes https://github.com/raphlinus/pulldown-cmark/issues/330.

marcusklaas:
Closes https://github.com/raphlinus/pulldown-cmark/issues/292.

marcusklaas:
Fixes https://github.com/raphlinus/pulldown-cmark/issues/331.

marcusklaas:
Closes https://github.com/raphlinus/pulldown-cmark/issues/326.

marcusklaas:
Fixes https://github.com/raphlinus/pulldown-cmark/issues/327.

marcusklaas:
Addresses https://github.com/raphlinus/pulldown-cmark/issues/329.

marcusklaas:
Closes https://github.com/raphlinus/pulldown-cmark/issues/295.

src/parse.rs
```diff
@@ -1482,18 +1486,24 @@ impl InlineStack {
                 min(mod3_lower, self.lower_bounds[1])
             }
         } else {
-            0
+            self.lower_bounds[5]
```
raphlinus: Might be a good idea to define constants for these magic numbers, now that there are a few of them.

marcusklaas: :+1: done!

marcusklaas:
Thanks for the review. And especially deep mega thanks to @oberien and @mity for opening so many hiqh quality issues with reproducible test cases! That really helped a lot! :bowing_man:

PR #335, opened 2019-05-27
Some fuzzer tweaks

marcusklaas:
This PR makes a few small changes:
 - do throughput reporting on main thread
 - reduce size of inputs for increased throughput
 - don't throw away outliers, but retest more often to confirm problematic patterns
 - round up when computing number of threads or else it won't run on single core systems at all :sweat_smile:

Throughput is significantly higher with fewer false positives on my machine. 

Could you review @oberien?

### DIFF
--- a/fuzzer/src/main.rs
+++ b/fuzzer/src/main.rs
@@ -126,39 +126,39 @@ fn main() {
 /// Returns the exit code. 0 if all tests passed and 1 otherwise.
 fn regression_test() -> i32 {
     let mut exit_code = 0;
-    let mut check_result = |res| match res {
+    let mut check_result = |pat| match test(pat) {
         PatternResult::Linear(_) => (),
         _ => exit_code = 1,
     };
     // https://github.com/raphlinus/pulldown-cmark/issues/246
-    check_result(test("[]("));
+    check_pattern("[](");
     // https://github.com/raphlinus/pulldown-cmark/issues/247
-    check_result(test("``\\"));
+    check_pattern("``\\");
     // https://github.com/raphlinus/pulldown-cmark/issues/248
-    check_result(test("a***"));
+    check_pattern("a***");
     // https://github.com/raphlinus/pulldown-cmark/issues/249
     // TODO: we can't perform tests like this yet
-//    check_result(test("* * * ...a"));
+//    check_pattern("* * * ...a");
     // https://github.com/raphlinus/pulldown-cmark/issues/251
-    check_result(test("[ (]("));
+    check_pattern("[ (](");
     // https://github.com/raphlinus/pulldown-cmark/issues/255
-    check_result(test("[*_a"));
+    check_pattern("[*_a");
     // https://github.com/raphlinus/pulldown-cmark/issues/280
-    check_result(test("a <![CDATA["));
+    check_pattern("a <![CDATA[");
     // https://github.com/mity/md4c/issues/73#issuecomment-487640366
-    check_result(test("a <!A"));
+    check_pattern("a <!A");
     // https://github.com/raphlinus/pulldown-cmark/issues/282
-    check_result(test("a<?"));
+    check_pattern("a<?");
     // https://github.com/raphlinus/pulldown-cmark/issues/284
-    check_result(test("[[]()"));
+    check_pattern("[[]()");
     // https://github.com/raphlinus/pulldown-cmark/issues/287
     // TODO: we can't perform tests like this reliably yet
-//    check_result(test("[{}]:\\a"));
+//    check_pattern("[{}]:\\a");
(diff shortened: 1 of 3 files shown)


fuzzer/src/main.rs
```diff
@@ -278,50 +282,46 @@ fn test_catch_unwind(pattern: &str) -> Result<PatternResult, ()> {
...
-    if let PatternResult::Linear(_) = res1 {
-        return res1;
-    }
-    // If the first pass indicated possible non-linear behaviour, retest it with a larger
-    // sample size and handle outliers.
-    let res2 = test_pattern(pattern, &mut time_samples, true);
-    if let PatternResult::Linear(_) = res2 {
-        return res2;
+    let mut time_samples = [(0.0, 0.0); SAMPLE_SIZE];
+    let mut res = PatternResult::TooLong;
```
oberien: I don't think you need to initialize this variable, because it's always initialized before its first use.

marcusklaas: So I thought as well, but apparently the compiler can't infer that the loop will execute at least once..

marcusklaas:
The reason for printing throughput in the main thread is twofold:
- it makes the updates more accurate. The first thread to finish the batch would underestimate the rate since it would not see the other batches that are about to finish
- updates are more predictable because they don't scale with the number of threads

The PR drops outlier elimination not for speed, but because to make simplify the procedure. I feel it's better to make the routine resistant to statistical noise instead of trying to reduce it manually.

Good point about updating the docs, totally forgot about that!

marcusklaas:
Yay, many of the bugs the fuzzer found have now been fixed. Your work is already starting to pay off :-)

The CI task seems to work well too. Although [spurious false positives](https://raphlinus.visualstudio.com/pulldown-cmark/_build/results?buildId=401) do seem to happen. We can fix that luckily :-)

fuzzer/src/main.rs
oberien: Oh right, the problem here is the if-condition after the for-loop, because the compiler doesn't (yet) know if the loop will ever execute, which may result in an uninitialized variable being used in the if-condition.

oberien:
> * it makes the updates more accurate. The first thread to finish the batch would underestimate the rate since it would not see the other batches that are about to finish

The approach in this PR only helps marginally, because you also underestimate throughput if there are threads which take more than 5 seconds longer than others. If 2 out of 4 threads finish within the 5s sleeping period of the timeout thread, then the timeout thread will only report those two threads as finished and underestimate the throughput by a half.

> * updates are more predictable because they don't scale with the number of threads

It still depends on the number of threads, because the throughput is only reported whenever at least one batch (thread) is finished. As such it depends on the number of threads just like before. The only advantage is thay you sometimes process batches of threads by grouping multiple threads together if they finish their batches within the 5s sleep period.

> The PR drops outlier elimination not for speed, but because to make simplify the procedure. I feel it's better to make the routine resistant to statistical noise instead of trying to reduce it manually.

You're absolutely right here. The scoring function should be responsible for being resistant to noise / outliers. If you think the current noise is low enough for the current slopes_stddev function to work like this, I approve of this change. I just never had the guts to do it until we find a better scoring function :D

> [CI:] Although spurious false positives do seem to happen. We can fix that luckily :-)

I think the long-term solution here is the same as before, to find a proper scoring function.

> Yay, many of the bugs the fuzzer found have now been fixed. Your work is already starting to pay off :-)

:+1: Looks like I can fire the fuzzer back up again on the server :)

marcusklaas:
> The approach in this PR only helps marginally, because you also underestimate throughput if there are threads which take more than 5 seconds longer than others. If 2 out of 4 threads finish within the 5s sleeping period of the timeout thread, then the timeout thread will only report those two threads as finished and underestimate the throughput by a half.

True. It's still not very good, but slightly better I believe. Not sure how important doing super accurate throughput estimates is, but we can probably do much better.

> It still depends on the number of threads, because the throughput is only reported whenever at least one batch (thread) is finished.

I meant that the rate at which messages are printed scales with the number of threads in the old approach.

> I think the long-term solution here is the same as before, to find a proper scoring function.

For sure. I've started doing some experiments [here](https://github.com/marcusklaas/linearity-testing-experiments). That'll be a later PR though.

I'll try to fix this PR later today or tomorrow at the latest.

oberien:
Could you please also update the main-module doc comment? After that I think this PR can be merged. (To this day I don't know if no comments or redundant comments are worse…)

On the Server I get about 18000 patterns per second on this branch. I set the acceptance for the slopes_stddev to 100 (was 70 before this PR) and I seem to get a few false positives per second (guessing >2 and <10, but don't have actual data so far), which I guess is fine for now. With your research regarding a proper scoring functions I'm highly positive that the number of false positives will go down.

Issue #336, opened 2019-05-27
Raw HTML block versus HTML inline.


mity:
Input:
```
<foo bar>
```

Expected output:
```
<foo bar>
```
(Rationale: https://spec.commonmark.org/0.29/#html-blocks, start condition 7)

Actual output:
```
<p><foo bar></p>
```
(So, the tag is good enough for inline HTML, but not for HTML block? Strange.)

mity:
Imho the block start condition 7 and inline HTML (the open and close tag cases) should ideally share code as much code as possible in order to guarantee the consistency.

marcusklaas:
Indeed, this is [long overdue](https://github.com/raphlinus/pulldown-cmark/blob/master/src/scanners.rs#L1233) ;-)

marcusklaas:
Blocks of type 7 don't seem to allow multi-line attribute values though, whereas this is fine for inline html. So I guess they're not completely equivalent.

This is inline, apparently.
```
<foo bar="hello
world">
```

mity:
That's true, and not only in the attributes. The complete tag must be on the single line for the block case.
```
<tag
attr="this cannot be block too"
attr2="but still it can be inline">
```

But I believe the EOLs are the only difference. The parser function for that may be parametrized to handle both cases, but not with just a bool value as the inline may be prefixed with some block decorations, e.g. in a blockquote:

```
> This paragraph inside a blockquote can contain an <inline_html
> attr="spanning over two lines">.
```

MD4C works that way. See [`md_is_html_tag()`](https://github.com/mity/md4c/blob/master/md4c/md4c.c#L951). In [the case of block](https://github.com/mity/md4c/blob/master/md4c/md4c.c#L5170), the caller has also additionally test there is nothing more on the line.

marcusklaas:
Aah yes, that's a great point. We're currently missing some logic to properly deal with multiline leaf elements in containers like blockquotes. Will need to think a bit about how to best solve this.

Thanks once more for your excellent issues!

marcusklaas:
Hmmm wait, your example doesn't render as inline html in the reference implementation :thinking: 

https://spec.commonmark.org/dingus/?text=%3E%20This%20paragraph%20inside%20a%20blockquote%20can%20contain%20an%20%3Cinline_html%0A%3E%20attr%3D%22spanning%20over%20two%20lines%22%3E.

mity:
Because I shouldn't use the underscore.

```
[...]

PR #338, opened 2019-05-28
Share more code between inline html and html7 blocks

marcusklaas:
Fixes https://github.com/raphlinus/pulldown-cmark/issues/336.


marcusklaas:
@raphlinus this could be a good candidate for 0.5.3

marcusklaas:
I am fairly confident this shouldn't introduce any new problems, so I believe we can go ahead. Thanks for having a look anyway! :+1:

PR #341, opened 2019-05-30
Add some documentation for exported types

marcusklaas:
This fills some gaps in the documentation of exported types.

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -31,27 +31,38 @@ use crate::scanners::*;
 use crate::strings::CowStr;
 use crate::tree::{Tree, TreeIndex, TreePointer};
 
+/// Tags for elements that can contain other elements.
 #[derive(Clone, Debug, PartialEq)]
 pub enum Tag<'a> {
-    // block-level tags
+    /// A paragraph of text and other inline elements.
     Paragraph,
-    Rule,
+    /// A horizontal ruler.
+    Rule, // FIXME: shouldn't this just be an event?
 
     /// A heading. The field indicates the level of the heading.
     Header(i32),
 
     BlockQuote,
+    /// A code block. The value contained in the tag describes the language of the code,
+    /// which may be empty.
     CodeBlock(CowStr<'a>),
 
     /// A list. If the list is ordered the field indicates the number of the first item.
+    /// Contains only list items.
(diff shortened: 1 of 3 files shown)


src/parse.rs
```diff
@@ -100,30 +112,50 @@ impl LinkType {
...
     End(Tag<'a>),
+    /// A text node.
     Text(CowStr<'a>),
+    /// An inline code node.
     Code(CowStr<'a>),
+    /// An HTML node.
     Html(CowStr<'a>),
-    InlineHtml(CowStr<'a>),
+    /// An inline HTML node.
+    InlineHtml(CowStr<'a>), // TODO: do we need a distinction between Html blocks/ Html/ InlineHtml?
```
oberien: Yes please. I'm still wondering what the difference between `Html` and `InlineHtml` is and I guess most people won't read into the CommonMark spec to find out.

src/parse.rs
```diff
@@ -31,27 +31,38 @@ use crate::scanners::*;
...
 
+/// Tags for elements that can contain other elements.
 #[derive(Clone, Debug, PartialEq)]
 pub enum Tag<'a> {
-    // block-level tags
+    /// A paragraph of text and other inline elements.
     Paragraph,
-    Rule,
+    /// A horizontal ruler.
+    Rule, // FIXME: shouldn't this just be an event?
```
raphlinus: I'm trying to remember the rationale, there might not be a good one. This does feel more event-like than tag-like.

src/parse.rs
```diff
@@ -100,30 +112,50 @@ impl LinkType {
...
     Left,
     Center,
     Right,
 }
 
 bitflags! {
+    /// Option struct containing flags for enabling extra features
+    /// that are not part of the CommonMark spec.
     pub struct Options: u32 {
         const FIRST_PASS = 1 << 0;
```
raphlinus: This should be removed next major release, as it's a source of confusion. But taking it out now would be a semver break. It might be worth adding a comment now saying, "this is unused and will be removed."

marcusklaas: Both of these are addressed in https://github.com/raphlinus/pulldown-cmark/pull/343 :-)

Will add that comment for now.

PR #343, opened 2019-05-31
Simplify Html and Options

marcusklaas:
Removes some unused and unnecessary code. This is a breaking change.


Keats:
While you're doing breaking changes, you might consider renaming Header -> Heading?
See the comment by @chris-morgan https://github.com/getzola/zola/pull/685#issuecomment-493488879 and http://altformat.org/index.asp?id=5&pid=389&ipname=GB

marcusklaas:
I am all for it!

marcusklaas:
The zola people were the ones who suggested this ;-) But it would definitely be helpful to get more feedback.

GuillaumeGomez:
Looks good to me as well! :+1:

Issue #342, opened 2019-05-31
Document minimum version of Rust required


Keats:
It looks like the latest version has bumped the minimum version to 1.33? I only found that when CI failed for a project with 1.31.

The pattern I follow on my projects is to have a CI task with the known lowest allowed version so I know if a dependency or some new features are requiring a newer Rust: https://github.com/Keats/tera/blob/master/.travis.yml

marcusklaas:
I must admit that I don't know. We haven't tracked this so far. It's a good idea.

Keats:
It's 1.34

PR #349, opened 2019-06-03
Document MSRV

marcusklaas:
Closes https://github.com/raphlinus/pulldown-cmark/issues/342.


raphlinus:
I'm slightly confused by this commit because it shows a lot of changes. I think that's because of rebasing, and the only substantive commit is https://github.com/raphlinus/pulldown-cmark/pull/349/commits/f5e4f2f98e7c5808c1a2d153b0e2522339644518. If that's the case, then I'm happy for the commit to go in. If there's any confusion, maybe it's better to rebase on top of current master?

marcusklaas:
Good point, will rebase!

Issue #346, opened 2019-05-31
Configure link attributes as options


Keats:
I have seen 2 requests for absolute links attributes that are present on Blackfriday (a md renderer in Go):

## target=blank

Initial feature request from Zola with more details: https://github.com/getzola/zola/issues/681

## rel="noopener/noreferrer/nofollow"

Initial feature request from Zola with more details: https://github.com/getzola/zola/issues/695

Those are three separate options in Blackfriday (https://gohugo.io/getting-started/configuration/#blackfriday-options)

cc @lukeshiao and @MichaelByrneAU in case they have other things they want to mention.

This is easily implementable in Zola itself but it sounds like something like can/should be built-in.

Keats:
@marcusklaas would that be accepted as a PR? I say that without having looked at the pulldown-cmark code so hopefully it's not too hard to implement.

Keats:
ping @marcusklaas

marcusklaas:
Apologies for the long silence here. It's a tricky issue. It's currently still uncertain whether we want to keep the standard renderer mean and lean or provide something more complete out of the box.

I've opened a discussion on this subject on the xi zulip. I hope to share a decision soon!

marcusklaas:
This was discussed briefly with @raphlinus and it seems best to leave more advanced rendering features to a separate crate. We'll maintain a lean implementation within this repository. It should be fairly straightforward to take this implementation and extend it in another crate. If a good community effort emerges, we might bless that. If not, I may start one myself.

Hope that this answers your question.

marcusklaas:
Since this won't be addressed at least in the near future, I will close this issue for now.

Issue #142, opened 2018-08-05
Footnote back-references


lnicola:
It would be nice if the footnotes linked back to the source, like [here](https://en.wikipedia.org/wiki/Huguenot-Walloon_half_dollar#cite_note-FOOTNOTESwiatek_&_Breen106-17).

2mol:
I came here to open the same request, since I would really like this in Zola.

Has anything like this been implemented, or do you need help? I'm happy to offer to do it, maybe with some pointers to the right place in the code, since I'm not a Rust programmer.

marcusklaas:
This is a tricky one. On the one hand, the value of back-references are immediately apparent.

On the other, implementing does not play well with the two-phase design of pulldown. One would have to go through all the events and capture the footnote references before you could emit a footnote itself. Usually, when footnotes are where they are supposed to be --at the bottom-- that's ok. But it's also valid to have footnotes at the top, like this:

```md
[note]: we'll be referring to this later!

See the [^note] above.
```

Adding support for this not only brings some extra user complexity, but also a big hit to performance. The pulldown html parser is meant to provide a lean and performant reference implementation, which does not lend itself well to this feature.

The reality is that many users of the library do not write their own renderers and there seems to be an appetite for more features like this. We may have to re-evaluate whether we want to keep a lean implement or rather provide something more featureful.

alerque:
Personally I use this exclusively for scanning to see what Markdown features are used where in source files; and it was hands down the best choice for that because of the raw speed. If things like this get implemented at all I would like to see them relegated to the renderer, not the parser. Personally I would have like to have _just_ a parsing library with a separate renderer based on it, but in any even I would still appreciate being able to parse without being slowed down by features like this.

2mol:
That makes sense, thanks for the thoughtful reply @marcusklaas . I don't have a clever suggestion, since I'm not familiar with the parser. But would you necessarily need to capture strictly valid back-references? If the renderer emits a `#note` link, and I don't have a target for it, that's sort of my problem, no? But maybe I don't have the right mental model, I'm assuming the parser wants to traverse the entire document just once.

I agree with both you and @alerque that taking a performance hit for this feature is not desirable.

krtab:
Hi, I have hit the same issue and here are my thoughts on it: as @2mol said, if we don't care about generating dangling backlinks for footnotes that are never referenced, then this is simply a matter of adding the right `id` to the reference. Except that there is not guarantee that there will only be one reference to a given definition.

[...]

Issue #130, opened 2018-03-12
Add support for extending/overwriting handlers for tags


lambdadog:
I'm currently writing something that calls for syntax highlighting within markdown code blocks and I've been looking around and it doesn't seem there's any real inbuilt extensibility in pulldown-cmark like I'd hoped.

Correct me if I'm wrong, since I'm new to rust, but the only real ways to do this are to

1. Fork the project and edit the lines I need to (sort of overkill),
2. Parse the html after the conversion has finished and replace it (cludgy and inefficient),
2. Completely reimplement the portion of the code that handles converting to html (duplicating a large block of code really shouldn't be necessary), or
3. Create a "proxy" parser that contains the actual parser inside of it and takes over when a code block starts, then pushes it all through as InlineHtml when the code block closes (probably what I'll actually do, but definitely jumping through hoops to accomplish a fairly simple change)

I know this isn't necessarily the focus of the project, but it would definitely be helpful to devs using pulldown-cmark to be able to extend/modify the html conversions a little bit without jumping through hoops.

raphlinus:
The current HTML renderer is designed to be simple. I'm worried that extensibility would make a significantly more complicated API surface. That said, I'm not completely averse to such a thing, and will take another look after the new_algo branch lands.

Figments:
I'd love to see this too. Been wanting to add support for Discord-like tags for underlining and spoiler text in my project.

Weasy666:
@transitracer I have the same use case and am facing the same problem. Have you found a solution?

Martin1887:
This seems to be the implementation of HTML rendering in a different crate, like #637, #408 and #525.

Issue #116, opened 2017-11-14
Extensible rendering proposal


RadicalZephyr:
Hi there!

@reillysiemens and I have been working on [cleaning up the customized Markdown processing](https://github.com/Keats/gutenberg/pull/174) in the [Gutenberg static site generator](https://github.com/Keats/gutenberg), which uses pulldown-cmark for the base markdown processing.

For the most part, the extensions that "Gutenberg Flavored Markdown" would require can be accomplished just by manipulating the iterator of events before passing it to the renderer.  However, for a few features, notably adding ids to `h*` tags and otherwise manipulating the internal portions of how specific tags are rendered doesn't seem to be possible with the current design.

We have started exploring implementing a more extensible structure in the `html` module that would allow reuse of the current commonmark html generation code, while also allowing the user to provide custom rendering behavior and even possibly their own event types.  I've submitted [a PR](#115) with our experimentation so far.

The basic gist of our approach is to introduce a new trait `IntoHtml` for rendering `Event`s into HTML, move the current commonmark rendering to be done through the new trait, and then also expose enough of the current rendering implementation in a public API so that is available for reuse.

Hopefully this idea sounds interesting to you all, and if you have any feedback about what you think the end API should look like, that would be much appreciated.  Also if I'm missing a way to customize how events are rendered please let me know!

Thanks!

marcusklaas:
This seems to be the most highly requested feature, seeing as how this could solve these issues:

- footnote back-references: https://github.com/raphlinus/pulldown-cmark/issues/142
- tag overwrites: https://github.com/raphlinus/pulldown-cmark/issues/130
- custom link attributes: https://github.com/raphlinus/pulldown-cmark/issues/346

Unfortunately, it's not as easy as implementing `IntoHtml` for `Event` since some elements (tables and image alts are examples) need access to the event iterator and other state.

As I see it, there's three ways we could go about this:

1. Document how to write a custom renderer. It's not hard and it's not a lot of code. The default one included is around 400 LoC. Much of the functionality will likely be duplicated with the default renderer however.
2. Introduce a new, extensible HTML renderer. Maybe one where you'd supply a `FnMut(Event) -> Option<String>`, and the default behavior applying on `None`s.
3. Extend the current HTML renderer with similar functionality, but providing access to internal `write_str` methods for lower overhead. Hopefully we can do this without affecting default performance.

marcusklaas:
[...]

PR #350, opened 2019-06-03
Extensible rendering prototype

marcusklaas:
This implements a prototype of option 3 listed in https://github.com/raphlinus/pulldown-cmark/issues/116. It allows for users to define their own custom rendering for a selection of events and/ or tags. This should add the required flexibility to address many use cases without having to account for them all in pulldown-cmark itself, keeping the implementation lean and fast.

An example inline HTML stripper use case could look like this:
```rust
use pulldown_cmark::{html, Parser, Event, Tag};

let markdown_str = "No html!<foo>";
let mut html_buf = String::new();
let parser = Parser::new(markdown_str);

html::push_html_with_extension(&mut html_buf, parser, |state, event| {
    if let Event::InlineHtml(..) = event {
        Ok(state.write("<REDACTED>"))
    } else {
        // default rendering
        Err(event)
    }
});

assert_eq!(&html_buf[..], "<p>No html!<REDACTED></p>\n");
```

The overhead for adding custom rendering would be fairly small and none if it's not used.

Paging issues https://github.com/raphlinus/pulldown-cmark/pull/103, https://github.com/raphlinus/pulldown-cmark/issues/116, https://github.com/raphlinus/pulldown-cmark/issues/142, https://github.com/raphlinus/pulldown-cmark/issues/130 and https://github.com/raphlinus/pulldown-cmark/issues/346.

Paging users @RadicalZephyr, @maghoff, @Inicola, @Keats, @transitracer and @Figments.

Would this cover your use cases? And would such an interface work for you?

### DIFF
--- a/src/html.rs
+++ b/src/html.rs
@@ -109,7 +109,7 @@ struct HtmlWriter<'a, I, W> {
 impl<'a, I, W> HtmlWriter<'a, I, W>
 where
     I: Iterator<Item = Event<'a>>,
-    W: StrWrite,
+    W: StrWrite    
 {
     fn new(iter: I, writer: W) -> Self {
         Self {
@@ -140,8 +140,14 @@ where
         Ok(())
     }
 
-    pub fn run(mut self) -> io::Result<()> {
+    pub fn run<F>(mut self, mut custom_renderer: F) -> io::Result<()>
+        where F: FnMut(&mut HtmlWriter<'a, I, W>, Event<'a>) -> Result<io::Result<()>, Event<'a>>
+    {
         while let Some(event) = self.iter.next() {
+            let event = match custom_renderer(&mut self, event) {
+                Ok(res) => return res,
+                Err(event) => event,
+            };
             match event {
                 Start(tag) => {
                     self.start_tag(tag)?;
@@ -478,7 +484,7 @@ pub fn push_html<'a, I>(s: &mut String, iter: I)
 where
     I: Iterator<Item = Event<'a>>,
 {
-    HtmlWriter::new(iter, s).run().unwrap();
+    HtmlWriter::new(iter, s).run(|_, event| Err(event)).unwrap();
 }
 
 /// Iterate over an `Iterator` of `Event`s, generate HTML for each `Event`, and
@@ -519,5 +525,5 @@ where
     I: Iterator<Item = Event<'a>>,
     W: Write,
 {
-    HtmlWriter::new(iter, WriteWrapper(writer)).run()
+    HtmlWriter::new(iter, WriteWrapper(writer)).run(|_, event| Err(event))
 }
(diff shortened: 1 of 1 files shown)


marcusklaas:
OTOH, this may not be better than simply filtering or mapping events of interest to HTML events, since those will be written to the buffer verbatim. :thinking:

maghoff:
Appreciate the initiative :) This looks like the first implementation I had in my branch for HTML-less markup (#103), but it only made me realize that it cannot work.

I'm not looking for a way to secure the input, but rather to change the input language such that HTML tags aren't special at all. The problem is that HTML tags trigger a state change in the parser, so it parses the tags _as HTML_, where I would need it to be parsed as Markdown.

For example, I would want `<a href=\"this *should* be parsed\">Not a link</a>` to cause `should` to be rendered emphasized: `&lt;a href=&quot;this <em>should</em> be parsed&quot;&gt;Not a link&lt;/a&gt;`. See [my test suite](https://github.com/maghoff/pulldown-cmark/blob/master/tests/disable_html.rs) for more examples.

marcusklaas:
Ah yes, it seems that would have to be tackled in the parser indeed. Thanks for elaborating!

Keats:
For #346 it might work but ideally it would be done in pulldown_cmark directly as that seems something common enough that it wouldn't make sense for every user to re-implement it.

Figments:
Yeah, this is something I'm really excited to use. Though, what about modifying existing Markdown syntax, such as `_this_` being an underline rather than emphasis? [Discord does this for multiple tags, on top of extensions.](https://support.discordapp.com/hc/en-us/articles/210298617-Markdown-Text-101-Chat-Formatting-Bold-Italic-Underline-) Would something like this be possible?

marcusklaas:
Yes, that would be possible with this API extension.

Figments:
That's perfect, then. 😄

gauravtiwari:
Hey, @marcusklaas Any plans when this is gonna be in?

marcusklaas:
It's not clear that this is a proper solution as it covers only slightly more use cases than an adapter around the event iterator. Uses like the one described by @maghoff can't be dealt with using this extension. The additional complexity may not be worth it.

I think we need to further others options first before we consider merging this. More input would be useful!

Figments:
Has any progress been made on this front? Sorry if asking this comes across as rude.

marcusklaas:
@Figments don't worry, it's a fair question. There's no recent progress on this, and there is unlikely to be much progress on this in the foreseeable future. This design adds complexity while still not covering all use cases. I think we should stick to the current philosophy of only providing a simple renderer for basic use cases until a design is found that strikes a good balance between complexity and utility. I will close this PR as I think it's a dead end.

Fortunately, it is really easy to either inject HTML events into the stream to do simple rendering extensions or even write a custom renderer. The default renderer is a good example and very succinct!

PR #354, opened 2019-06-08
Deprecations and other minor issue fixes

marcusklaas:


marcusklaas:
@raphlinus it may be good to include this in a release as well, as it prevents warnings (and later possibly errors!) in future releases of Rust

marcusklaas:
I do not, I will check that we don't regress here. Good catch!

marcusklaas:
Just checked, and it compiles fine on 1.34, which is our MSRV. When we get around to merging https://github.com/raphlinus/pulldown-cmark/pull/349, this should be guarded by the build pipeline. That's not a requirement for this release though.

PR #371, opened 2019-07-11
Always interrupt paragraph on blank line

### LINKED ISSUE #370: blank line with whitespace only is treated as a soft break
The following string (in Rust syntax) treats the line with whitespace as a soft break.

```
"Paragraph 1\n    \nParagraph 2\n"
```

Outputs the following events:

```
Start(Paragraph)
Text(Borrowed("Paragraph 1"))
SoftBreak
HardBreak
Text(Borrowed("Paragraph 2"))
End(Paragraph)
```

I would expect it to be treated as a "blank line" as defined in the spec.

> A line containing no characters, or a line containing only spaces (U+0020) or tabs (U+0009), is called a blank line.

and should emit the following events:

```
Start(Paragraph)
Text(Borrowed("Paragraph 1"))
End(Paragraph)
Start(Paragraph)
Text(Borrowed("Paragraph 2"))
End(Paragraph)
```

Using pulldown-cmark 0.5.2 (42296ca).

https://spec.commonmark.org/dingus/?text=Paragraph%201%0A%20%20%20%20%0AParagraph%202%0A

raphlinus:
Previously it would fail to catch a line with four or more spaces.

Fixes #370


specs/regression.txt
```diff
@@ -424,3 +424,13 @@ __a__
 <p>~~*_**__</p>
 <p><strong>a</strong></p>
 ````````````````````````````````
+
+ISSUE 370
+Paragraph 1
+
```
marcusklaas: It looks like this blank line has no spaces on it? Maybe your editor trimmed the line? The generated test looks good, however.

marcusklaas: Woops, it looks like it was just the review pane that stripped the spaces. My bad!

Issue #375, opened 2019-07-18
Escape immediately after inline link doesn't work.


ehuss:
The following markdown: `\[[link text](https://www.google.com/)\]`

Results in: `<p>[<a href="https://www.google.com/">link text</a>\]</p>`
(Notice the extra backslash on the last bracket.)

Expected: `<p>[<a href="https://www.google.com/">link text</a>]</p>`

This used to work. Bisected to 0cfd39da66766034a275ebfcbc22e49b37611fa5 which changed behavior.

Tested on latest master (f6212d83f66944b8f5c3d4c36347cbc3c410f934).

PR #376, opened 2019-07-18
Fix bad span start after inline link

marcusklaas:
Fixes https://github.com/raphlinus/pulldown-cmark/issues/375.


marcusklaas:
Sure, but maybe we can sneak in some of the open PRs as well? There's a few ones that shouldn't be controversial and fix open issues.

Issue #382, opened 2019-08-03
Autolinks accept whitespace in the URI


mity:
The Commonmark specs forbids ASCII whitespace, control characters, `<`, and `>` inside autolinks but
```
$ printf '<http:// >' | ./pulldown-cmark
<p><a href="http://%20">http:// </a></p>
```

(Notice the space after the scheme and before the closing `>`.)

Issue #381, opened 2019-08-03
Autolinks not respecting block boundaries


mity:
Input:
```
foo | bar
--- | ---
<http://|>
```

Actual output (with `-T`):
```
<table><thead><tr><th>foo</th><th>bar</th></tr></thead><tbody>
<tr><td><a href="http://%7C">http://|</a></td><td>&gt;</td></tr>
</tbody></table>
```

I would expect there is no autolink because what looks as one is actually broken into multiple table cells.

(Might be the same cause as the one for #351. Perhaps scanners for other inline elements should be reviewed too to verify the issue is not more widely spread.)

marcusklaas:
Thanks for these!

PR #383, opened 2019-08-04
Autolink fixes

marcusklaas:
Addresses https://github.com/raphlinus/pulldown-cmark/issues/382 and https://github.com/raphlinus/pulldown-cmark/issues/381.

This PR should also fix any scans beyond block ends of other inline items.

Many thanks to @mity for the excellent bug reports!


mity:
The commit message "Don't allow ascii whitspace at end of autolink" makes me wonder whether it is good enough. CommonMark specs bans whitespace (as well as control chars) anywhere in the URI, so `<http://foo bar>` shouldn't be autolink too.

Even if you want to allow URIs with spaces in them, it should imho still at least also ban the control chars at the end of it.

marcusklaas:
Oh, sorry for the confusion. You're absolutely right that ascii whitespace is not allowed anywhere in the autolink. That was already mostly working, with only whitespace at the end being (incorrectly!) accepted by pulldown. That commit fixed that last case. Hope this clears it up!

mity:
Ok. Sorry for false alarm then. 

And the control char rule as e.g. for `<http://\a>` (in C string notation)?

marcusklaas:
Any [whitespace character](https://spec.commonmark.org/0.29/#whitespace-character) will cause the autolink to be rejected, including `U+000A`. Is that what you meant?

mity:
No. C esq. seq. '`\a`' is `U+0007` (known also as alert or bell char, used in ancient times to make a terminal beep if it supported it), not `U+000A`. It is ASCII conttol char (as any with a codepoint lower then 32) but it is not a whitespace.

marcusklaas:
Aaaah, I may have incorrectly removed that bit :fearful: 

Excellent catch! Thank you for your vigilance ;-) I'll correct this immediately.

Issue #298, opened 2019-05-05
Link ref. def. in a block quote 


mity:
For the input
```
> [foo
> bar]: /url
>
> [foo bar]
```

I would expect something like 
```
<blockquote>
<p><a href="/url">foo bar</a></p>
</blockquote>
```

but I get 

```
<blockquote>
</blockquote>
<blockquote>
<p>[foo bar]</p>
</blockquote>
```

which makes no sense at all.

(tricky one from https://github.com/mity/md4c/blob/master/test/coverage.txt#L53)

marcusklaas:
This is a hard one. Good catch!

PR #386, opened 2019-08-11
Correctly parse reference labels within containers

marcusklaas:
Fixes https://github.com/raphlinus/pulldown-cmark/issues/298.

### DIFF
--- a/specs/regression.txt
+++ b/specs/regression.txt
@@ -602,3 +602,61 @@ ISSUE 301
 bar">link</a></p>
 </blockquote>
 ````````````````````````````````
+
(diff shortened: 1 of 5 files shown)


specs/regression.txt
```diff
@@ -525,6 +525,64 @@ ISSUE 351
 <p>&quot;)</p>
 ````````````````````````````````
 
+ISSUE 285
```
mity: Seems as a bad issue ID to me. (Ditto for the other tests below in this patch.)

marcusklaas: Great catch!

PR #387, opened 2019-08-12
Allow for more flexible fuzzer patterns

marcusklaas:
This PR allows for more flexible patterns, repeating patterns with a variable but fixed prefix and postfix. As a result, almost all of the existing scaling regression tests could be moved into the fuzzer regression suite, which means that they'll be included in the CI pipeline.

@oberien: the fuzzer seems to panic immediately nowadays when left to fuzz, even on the master branch. I get the following trace:
```bash
marcusklaas@localhost ~/p/fuzzer> env RUST_BACKTRACE=1 cargo run --release
    Finished release [optimized + debuginfo] target(s) in 0.05s
     Running `target/release/fuzzer`
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error("expected `struct`")', src/libcore/result.rs:1051:5
stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.29/src/backtrace/libunwind.rs:88
   1: backtrace::backtrace::trace_unsynchronized
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.29/src/backtrace/mod.rs:66
   2: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:47
   3: std::sys_common::backtrace::print
             at src/libstd/sys_common/backtrace.rs:36
   4: std::panicking::default_hook::{{closure}}
             at src/libstd/panicking.rs:200
   5: std::panicking::default_hook
             at src/libstd/panicking.rs:214
   6: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:481
   7: std::panicking::continue_panic_fmt
             at src/libstd/panicking.rs:384
   8: rust_begin_unwind
             at src/libstd/panicking.rs:311
   9: core::panicking::panic_fmt
             at src/libcore/panicking.rs:85
  10: core::result::unwrap_failed
             at src/libcore/result.rs:1051
  11: core::result::Result<T,E>::unwrap
             at /rustc/bc2e84ca0939b73fcf1768209044432f6a15c2e5/src/libcore/result.rs:852
  12: <fuzzer::literals::LiteralParser::extract_literals_from_macro::Bitflags as syn::parse::Parse>::parse
             at src/literals.rs:218
  13: core::ops::function::FnOnce::call_once
             at /rustc/bc2e84ca0939b73fcf1768209044432f6a15c2e5/src/libcore/ops/function.rs:231
  14: <F as syn::parse::Parser>::parse2
             at /home/marcusklaas/.cargo/registry/src/github.com-1ecc6299db9ec823/syn-0.15.34/src/parse.rs:1103
  15: syn::parse2
             at /home/marcusklaas/.cargo/registry/src/github.com-1ecc6299db9ec823/syn-0.15.34/src/lib.rs:633
  16: fuzzer::literals::LiteralParser::extract_literals_from_macro
             at src/literals.rs:236
  17: fuzzer::literals::LiteralParser::extract_literals_from_item
             at src/literals.rs:134
  18: fuzzer::literals::LiteralParser::extract_literals_from_items
             at src/literals.rs:100
  19: fuzzer::literals::LiteralParser::extract_literals_from_file
             at src/literals.rs:95
  20: fuzzer::literals::get
             at src/literals.rs:47
  21: fuzzer::fuzz
             at src/main.rs:211
  22: fuzzer::main
             at src/main.rs:148
  23: std::rt::lang_start::{{closure}}
             at /rustc/bc2e84ca0939b73fcf1768209044432f6a15c2e5/src/libstd/rt.rs:64
  24: std::rt::lang_start_internal::{{closure}}
             at src/libstd/rt.rs:49
  25: std::panicking::try::do_call
             at src/libstd/panicking.rs:296
  26: __rust_maybe_catch_panic
             at src/libpanic_unwind/lib.rs:82
  27: std::panicking::try
             at src/libstd/panicking.rs:275
  28: std::panic::catch_unwind
             at src/libstd/panic.rs:394
  29: std::rt::lang_start_internal
             at src/libstd/rt.rs:48
  30: main
  31: __libc_start_main
  32: _start
```
I haven't dug too deep, but it seems related to the token parsing. Which would be strange, since neither the fuzzer's dependencies nor the project's structure has really changed. Do you happen to know what's going on here?

### DIFF
--- a/fuzzer/retest-output
+++ b/fuzzer/retest-output
@@ -8,7 +8,7 @@ UTF8PATTERN="Couldn't convert pattern to UTF-8. Err: "
 JOINPATTERN="Joined thread: "
 TIMEOUTPATTERN="Thread timeout triggered (thread took too long) for Pattern: "
 # pattern to extract actual findings
-FINDPATTERN='^pattern: "'
+FINDPATTERN='^pattern: Pattern"'
 # output file of this program
 OUTFILE="output-retested"
 # temporary output file while performing a retesting run
--- a/fuzzer/retest-output
+++ b/fuzzer/retest-output
@@ -8,7 +8,7 @@ UTF8PATTERN="Couldn't convert pattern to UTF-8. Err: "
 JOINPATTERN="Joined thread: "
 TIMEOUTPATTERN="Thread timeout triggered (thread took too long) for Pattern: "
 # pattern to extract actual findings
-FINDPATTERN='^pattern: Pattern"'
+FINDPATTERN='^pattern: '
 # output file of this program
 OUTFILE="output-retested"
 # temporary output file while performing a retesting run
@@ -25,7 +25,7 @@ retestround() {
   grep "$JOINPATTERN" "$OUTFILE2" >> "$OUTFILE" && :
   grep "$TIMEOUTPATTERN" "$OUTFILE2" >> "$OUTFILE" && :
   cargo rustc --release -- -C target-cpu=native > /dev/null 2>&1
-  grep "$FINDPATTERN" < "$OUTFILE2" | sed 's/pattern: "\(.*\)"/\1/' | RUST_BACKTRACE=1 ./target/release/fuzzer --retest >> "$OUTFILE" 2>&1
+  grep "$FINDPATTERN" < "$OUTFILE2" | sed 's/pattern: \({.*}\)/\1/' | RUST_BACKTRACE=1 ./target/release/fuzzer --retest >> "$OUTFILE" 2>&1
   rm "$OUTFILE2"
 }
 
--- a/fuzzer/retest-output
+++ b/fuzzer/retest-output
@@ -25,7 +25,7 @@ retestround() {
   grep "$JOINPATTERN" "$OUTFILE2" >> "$OUTFILE" && :
   grep "$TIMEOUTPATTERN" "$OUTFILE2" >> "$OUTFILE" && :
   cargo rustc --release -- -C target-cpu=native > /dev/null 2>&1
-  grep "$FINDPATTERN" < "$OUTFILE2" | sed 's/pattern: \({.*}\)/\1/' | RUST_BACKTRACE=1 ./target/release/fuzzer --retest >> "$OUTFILE" 2>&1
+  grep "$FINDPATTERN" < "$OUTFILE2" | sed 's/pattern: \(.*\)/\1/' | RUST_BACKTRACE=1 ./target/release/fuzzer --retest >> "$OUTFILE" 2>&1
   rm "$OUTFILE2"
 }
 
@@ -41,8 +41,8 @@ for i in `seq 3`; do
 
   finds_before=$finds_after
   retestround
-  finds_after=$(grep "$FINDPATTERN" "$OUTFILE" | wc -l)
 
+  finds_after=$(grep "$FINDPATTERN" "$OUTFILE" | wc -l)
   false_positives_this_round=$(($finds_before - $finds_after))
   false_positives=$(($false_positives + $false_positives_this_round))
 
--- a/fuzzer/src/literals.rs
+++ b/fuzzer/src/literals.rs
@@ -350,12 +350,12 @@ impl LiteralParser {
             // handling lo is enough as we can trigger that pattern with the lo element already
             Pat::Range(pat) => self.extract_literals_from_expr(*pat.lo),
             Pat::Slice(pat) => {
-                for pat in pat
+                let pats_iter = pat
                     .front
                     .into_iter()
                     .chain(pat.middle.map(|pat| *pat))
-                    .chain(pat.back)
-                {
+                    .chain(pat.back);
+                for pat in pats_iter {
                     self.extract_literals_from_pat(pat);
                 }
             }
--- a/fuzzer/src/literals.rs
+++ b/fuzzer/src/literals.rs
@@ -2,9 +2,7 @@ use std::collections::HashSet;
 use std::fs;
 use std::path::Path;
 
-use syn::{
-    Expr, ExprArray, ExprCall, ExprMethodCall, ExprTuple, ImplItem, Item, Lit, Pat, Stmt,
-};
+use syn::{Expr, ExprArray, ExprCall, ExprMethodCall, ExprTuple, ImplItem, Item, Lit, Pat, Stmt};
 use walkdir::WalkDir;
 
 /// Get all relevant literals from pulldown-cmark to generate fuzzing input from.
--- a/fuzzer/src/literals.rs
+++ b/fuzzer/src/literals.rs
@@ -2,11 +2,16 @@ use std::collections::HashSet;
 use std::fs;
 use std::path::Path;
 
-use walkdir::WalkDir;
-use syn::{Item, Expr, ImplItem, Stmt, ExprArray, ExprCall, ExprMethodCall, Visibility, Ident, Type,
-          ExprTuple, Lit, Pat, punctuated::Punctuated, token::{Comma, Semi, Struct, Colon, Eq, Const, Brace},
-          parse::{Parse, ParseStream, Error}, braced};
 use proc_macro2::TokenStream;
+use syn::{
+    braced,
+    parse::{Error, Parse, ParseStream},
+    punctuated::Punctuated,
+    token::{Brace, Colon, Comma, Const, Eq, Semi, Struct},
+    Expr, ExprArray, ExprCall, ExprMethodCall, ExprTuple, Ident, ImplItem, Item, Lit, Pat, Stmt,
+    Type, Visibility,
+};
+use walkdir::WalkDir;
 
 /// Get all relevant literals from pulldown-cmark to generate fuzzing input from.
 ///
@@ -24,7 +29,13 @@ pub fn get() -> Vec<Vec<u8>> {
         .into_iter()
         .filter_map(|e| e.ok())
         .filter(|e| e.file_type().is_file())
-        .filter(|e| if let Some(ext) = e.path().extension() { ext == "rs" } else { false });
+        .filter(|e| {
+            if let Some(ext) = e.path().extension() {
+                ext == "rs"
+            } else {
+                false
+            }
+        });
 
     let mut literal_parser = LiteralParser::new();
 
@@ -79,10 +90,8 @@ impl LiteralParser {
 
     pub fn extract_literals_from_file<P: AsRef<Path>>(&mut self, path: P) {
         let path = path.as_ref();
-        let content = fs::read_to_string(path)
-            .expect(&format!("unable to read file {:?}", path));
-        let parsed = syn::parse_file(&content)
-            .expect(&format!("unable to parse file {:?}", path));
+        let content = fs::read_to_string(path).expect(&format!("unable to read file {:?}", path));
+        let parsed = syn::parse_file(&content).expect(&format!("unable to parse file {:?}", path));
         self.extract_literals_from_items(parsed.items);
     }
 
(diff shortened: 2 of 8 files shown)


oberien:
I'll review this commit tomorrow.

Regarding the panic from the looks it seems to occur in the macro parsing. Macros can be any arbitrary rust tokens. Therefore, we can't just parse them arbitrarily. Instead, I implemented the structure of all macros that had been used in pulldown-cmark (which afair have only been bitflags). I assume you added a new macro, which can't currently be parsed with the rules I defined.

marcusklaas:
Thanks Jaro, I'll look into it deeper.

marcusklaas:
I think some doc comments were added to the bitflags earlier; that may have tripped up the macro parser. Since that seems to be the only use of macros in the crate and it doesn't contain any interesting tokens, would it make sense to forego macro parsing? That's what I've done for now.

It does run again, but the time measurements are all over the place and it's hence generating lots of, presumably false, positives… Will need to dig even deeper for this.

fuzzer/src/main.rs
```diff
@@ -374,27 +418,35 @@ fn test_pattern(pattern: &str, time_samples: &mut [(f64, f64)]) -> PatternResult
...
+    buf.clear();
+    buf.push_str(&pattern.prefix);
+
+    let target_byte_count = sample_size * NUM_BYTES / sample_count;
+    let target_repeat_bytes = target_byte_count - pattern.prefix.len() - pattern.postfix.len();
+    let num_repeats = target_repeat_bytes / pattern.repeating_pattern.len();
+
+    for _ in 0..num_repeats {
+        buf.push_str(&pattern.repeating_pattern);
     }
```
oberien: The algorithm used by `str::repeat` (and `slice::repeat`) has an important optimization. It always memcopies the whole already repeated buffer as often as possible to provide exponential expansion and thus logarithmic time instead of linear time like here with the for-loop (<https://doc.rust-lang.org/src/alloc/slice.rs.html#434-487>).

fuzzer/src/literals.rs
```diff
@@ -322,87 +259,109 @@ impl LiteralParser {
...
-            Pat::Slice(pat) => for pat in pat.front.into_iter().chain(pat.middle.map(|pat| *pat)).chain(pat.back) {
-                self.extract_literals_from_pat(pat);
-            },
-            Pat::Macro(pat) => self.extract_literals_from_macro(pat.mac.tts),
+            Pat::Slice(pat) => {
+                for pat in pat
+                    .front
+                    .into_iter()
+                    .chain(pat.middle.map(|pat| *pat))
+                    .chain(pat.back)
```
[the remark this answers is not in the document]
marcusklaas: Agreed. Do note that this PR only formatted this code -- it didn't actually do any changes ;-)

fuzzer/src/main.rs
```diff
@@ -84,27 +85,66 @@ const TEST_COUNT: usize = 5;
...
+    fn sample<R: Rng + ?Sized>(&self, rng: &mut R) -> Pattern {
+        fn random_seq<'b, I: Iterator<Item = Option<&'b Vec<u8>>>>(iter: &mut I) -> String {
+            String::from_utf8(
+                iter.take(COMBINATIONS)
+                    .flatten()
+                    .flatten()
+                    .cloned()
+                    .collect(),
+            )
+            .expect("Bytes to String failed!")
```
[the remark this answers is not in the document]
marcusklaas: Good point, but that would indicate a bug in the fuzzer, right? Under our current assumptions, it *should* never panic. It may be better to fail and abort so that we can learn?

fuzzer/src/main.rs
```diff
@@ -84,27 +85,66 @@ const TEST_COUNT: usize = 5;
...
-        Some("--retest") => {
-            for pattern in io::stdin().lock().lines() {
-                let pattern = pattern.unwrap();
-                println!("retesting {:?}", pattern);
-                match test_catch_unwind(&pattern) {
-                    Ok(res) => println!("score: {}", res.score()),
-                    Err(()) => ()
-                }
-            }
-        },
```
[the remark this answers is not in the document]
marcusklaas: Right, good catch! I'll revert this change. Will probably need to alter this routine to read the new patterns instead of simple strings..

fuzzer/src/main.rs
marcusklaas: That's very interesting! Thanks for pointing it out. I'll try to rewrite this routine using those methods.

fuzzer/src/main.rs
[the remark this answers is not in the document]
oberien: I was very torn on the exact same thing. I opted for more safety than security here, because when I wrote this, the automatic restart mechanism wasn't in place yet. With the restart algorithm I don't see the need to be so strict about it anymore.

fuzzer/src/scoring.rs
```diff
@@ -40,12 +40,15 @@ pub fn slope_stddev(time_samples: &[(f64, f64)]) -> (f64, bool) {
...
+    let mean_time: f64 =
+        time_samples.iter().map(|tup| tup.1).sum::<f64>() / time_samples.len() as f64;
+
     for (point1, point2) in (0..time_samples.len()).tuple_combinations() {
         let (x1, y1) = time_samples[point1];
         let (x2, y2) = time_samples[point2];
         let dx = x2 - x1;
         let dy = y2 - y1;
-        let slope = dy / dx;
+        let slope = dy / dx / mean_time;
```
oberien: I tried ideas like this, but never reached a sufficient signal-to-noise-ratio. What is your experience so far with scaling them like this?

marcusklaas: It shouldn't really alter the signal to noise ratio. It just rescales things so that the scores aren't so dependent on the length of the patterns for example. Picking a good threshold value for the score should be easier now. I'm still getting *a lot* of noise ^^

fuzzer/src/main.rs
marcusklaas: It looks like `str::repeat` and `slice::repeat` don't quite work here since they themselves create owned data (`String`s) which is slow (O(n)) to prepend with the prefix.. These functions are also restricted to nightly for now.

I was able to optimize the pattern sampling function to continue building where the previous round left off. So it doesn't completely clear the buffer every time. This should make the string construction quite a bit faster.

marcusklaas:
This should be ready for another look, @oberien. I increased the sample size from 8KiB to 32KiB in order to get the noise levels to at least somewhat manageable levels. So far, no big new vulnerabilities have some up from my initial testing, but I wonder what your supercomputer is able to uncover after a night of intensive fuzzing ;-)

fuzzer/src/main.rs
oberien: This sounds like a nice optimization, which should have a pretty huge impact. :+1:

fuzzer/src/main.rs
```diff
@@ -445,8 +457,6 @@ fn time_needed(sample: &str) -> Duration {
     // perform actual time measurement
     let clock = Clock::<ThreadCpuTime>::now();
     let parser = Parser::new_ext(sample, Options::all());
-    parser.for_each(|evt| {
-        black_box::black_box(evt);
-    });
+    black_box::black_box(parser.count());
```
oberien: https://github.com/raphlinus/pulldown-cmark/pull/286#discussion_r280590293

oberien:
Apart from the possibly invalid black-boxing this lgtm.

> supercomputer

It's just a single machine with several cores. Very far from a supercomputer :) I'll let it run and let's see if it finds anything.

oberien:
In `retest-output` you need to change the sed replace pattern to catch the new Pattern structure. Currently it assumes that `pattern: ` is followed directly by a double-quoted string.

Also first short tests on the server seem to produce a massive amount of false positives. I think the score threshold is too large and may still be platform dependent (???). Also I get scores larger than `1.0`, which seems weird.

fuzzer/src/main.rs
marcusklaas: Forgot about that -- good catch!

marcusklaas:
Yeah, unfortunately the threshold is still platform dependent. The normalization change only made the scale of standard deviations independent of the configuration (`NUM_BYTES` and `SAMPLE_SIZE`). To create a true platform independent threshold, we likely need to alter the methodology altogether.

Thanks for the thorough review @oberien! Without it, many mistakes would have slipped through!

fuzzer/retest-output
```diff
@@ -25,7 +25,7 @@ retestround() {
   grep "$JOINPATTERN" "$OUTFILE2" >> "$OUTFILE" && :
   grep "$TIMEOUTPATTERN" "$OUTFILE2" >> "$OUTFILE" && :
   cargo rustc --release -- -C target-cpu=native > /dev/null 2>&1
-  grep "$FINDPATTERN" < "$OUTFILE2" | sed 's/pattern: "\(.*\)"/\1/' | RUST_BACKTRACE=1 ./target/release/fuzzer --retest >> "$OUTFILE" 2>&1
+  grep "$FINDPATTERN" < "$OUTFILE2" | sed 's/pattern: \({.*}\)/\1/' | RUST_BACKTRACE=1 ./target/release/fuzzer --retest >> "$OUTFILE" 2>&1
```
oberien: Either `pattern: \(Pattern {.*}\)` or `pattern: \(.*\)`, but the above doesn't match anything.

marcusklaas: Fixed!

oberien: Wait, that doesn't make any sense… It should work now with the serde serialization. Let me test a bit more, as things seem weird…

oberien: Nevermind, the problem was that my `output` file still had old patterns. Now it works as expected.

oberien:
I started it on the server and am going to let it run for a few days. With an stddev_acceptance of 0.7 it seems to only produce very few false positives. However, I can't verify if it doesn't also produce false negatives. Currently we don't know of any exponential behaviour I could use to find the acceptance produced by an exponential pattern to compare the stddev_acceptance to.

I get around 7000 patterns per second, which seems incredibly fast compared to earlier versions iirc.

fuzzer/retest-output
marcusklaas: Can happen, no worries!

marcusklaas:
Whaddya think, shall we merge this thing?

oberien:
I'm unsure if the scoring function still detects exponential behaviour. Take the following test (which should be added to the scoring module): <https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=f790e279d955731175c4c861b7b83604>

The passed values are highly exponential (2^x) ([graph](https://www.wolframalpha.com/input/?i=plot+2%5Ex+*+50000+%2B+50000+*+%28x-1%29+for+x+%3D+1+to+5)) and should definitely be detected, but are not.

marcusklaas:
Wow, that is a problem. It should definitely catch exponential behavior with any base that is not essentially 1. In fact, the big one is quadratic behavior, which is much more common. (I've never observed exponential behavior in pulldown before)

I guess the threshold should be much, much lower. The problem is that this will introduce immense numbers of false positives. It's unclear to me why, but it should obviously be resolved before we can merge.

oberien:
I think it was introduced by `dy / dx / mean_time`. You are dividing two derivations by an original value. Don't ask me for the mathematical reasons, but when I remove that from the playground, it produces quite a large stddev for the above example (25000). But it also produces quite a high value for smallish variations of a linear dataset (5000) while only having a low value for 1.4^x (1500).

marcusklaas:
Ok, I've reverted the normalization and added some unit tests for the stddev scoring function. Its score is very dependent again on the x scale and platform, which is obviously not ideal but the normalization didn't work out either as we've seen...

I'll try to finally make work of a more statistically sound test soon, which can hopefully overcome these issues.

PR #388, opened 2019-08-13
Simplify main program a bit

marcusklaas:
This PR removes the spec option, since these were made redundant
by the test suite and did not perform normalization, causing many
tests to fail.

It also removes the bench option, which was replaced by proper
benchmarks implemented with criterion.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -20,26 +20,11 @@
 
 //! Command line tool to exercise pulldown-cmark.
 
-extern crate getopts;
-
-extern crate pulldown_cmark;
-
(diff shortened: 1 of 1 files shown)


ScottAbbey:
> This commit removes the **bench** option...

_spec_

marcusklaas:
Thanks ^^

marcusklaas:
Yup, it looked really useful. Unfortunately, it has fallen behind the automated test suite in that it doesn't do proper normalization and subsequently doesn't pass the spec..

PR #392, opened 2019-08-31
remove the getopts dependency from the lib

andjo403:
to avoid that crates that uses this lib needs to compile getopts

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,6 +35,6 @@ tendril = "0.4"
 criterion = "0.2"
 
 [features]
-default = ["getopts"]
+default = []
 gen-tests = []
 simd = []


andjo403:
hmm I see now that there have been problems with publishing something like this before #56
but at least `cargo +1.34.0 publish --dry-run` works for me

marcusklaas:
You can avoid pulling in the getopts dependency by disabling the default features in your cargo.toml file like so:
```
pulldown-cmark = { version = "0.5", default-features = false }
```
I agree it's not the most elegant solution, but it allows the binary to work without explicitly enabling features.

andjo403:
think this PR solves both the problems,
I did not have to specify and feature when compiling the binary and did not have to set the default-features= false to avoid it in the lib that used this crate.
but maybe I did somthing wrong

marcusklaas:
I'm getting this error when trying to run on your branch:
```bash
marcusklaas@localhost ~/pulldown-cmark> cargo run
error: target `pulldown-cmark` in package `pulldown-cmark` requires the features: `getopts`
Consider enabling them by passing, e.g., `--features="getopts"`
```

andjo403:
taught that I hade tested all cases but apperently made some thing strange will close then sorry for the noise

PR #403, opened 2019-09-13
Excluding unneeded folders from the package

claudiocherubino:
The "tests" and "third_party" folders are not needed for the published package, and this change will simplify licenses and reduce the crate size.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,6 +9,7 @@ keywords = ["markdown", "commonmark"]
 categories = ["text-processing"]
 edition = "2018"
 readme = "README.md"
+exclude = ["/third_party/**/*", "/tools/**/*", "/specs/**/*", "/fuzzer/**/*", "/benches/**/*", "/azure-pipelines.yml"]
 
 build = "build.rs"
 


marcusklaas:
This is great, thanks! In addition, I believe that furthermore the benches, examples, fuzzer, specs and tools folders are not required in the publication. Could you add these to the exclusion list?

marcusklaas:
Out of curiosity I've been looking at some other crates, and it seems it's not common to exclude tests in Cargo.toml. Are we sure we want to exclude them?

I must admit I don't really understand what the exclude rule does, other than the little bit of information the Cargo manifest describes. Does it manage what files are downloaded when a crate is pulled in as a dependency? In that case it may make sense to exclude everything that isn't strictly required to compile the code, but it may be useful to include tests, although I wouldn't even know how to run them for dependencies.

polyzen:
Please include the tests so downstream could be sure tests pass on their systems before building.

claudiocherubino:
I'm also a little confused by the exclude rule, but my understanding is that it controls what files become part of the crate, and the default is to include everything.
Happy to add benches, examples, fuzzer, specs, and tools to the exclude list, as they are also not needed by those who depend on the crate.

As for the tests, I expected those who want to test the behavior of this library to checkout the source and run the tests, especially given that we are potentially stripping out other dependencies. That's why I suggested removing the tests folder too.
Does that make sense?

polyzen:
If the tests won't work and/or it saves a fair amount of bytes, then absolutely. I prefer to use published offerings as they are what the authors intend to be used. I grab the crate and make sure the tests pass as part of packaging the binary for Arch Linux (not (yet?) in the repos).

claudiocherubino:
Fair enough, I'm not one of the authors so I'll defer to them.
Please let me know if you want me to update this PR or you want to take care of the changes yourself.

marcusklaas:
Since the implications of exclusions are not exactly clear, I propose we take a conservative approach and do not exclude tests or examples, which may be of interest to crate users. Stuff that's of interest only to crate contributors can be excluded. I'm thinking `/third_party/**`, `/tools/**`, `/specs/**`, `/fuzzer/**`, `/benches/**` and `azure-pipelines.yml` for now.

If you could update the PR, I'd be happy to merge it!

claudiocherubino:
Thanks, updated the PR.

marcusklaas:
Thank you!

Issue #119, opened 2017-12-15
smart punctuation option?


sodiumjoe:
Is there a way to parse with smart punctuation?

ErichDonGubler:
Let me offer some more detail -- seems like this issue doesn't have much detail to act on.

One of the features I thought would be interesting from using Jekyll is "smart" typography, which comes from the SmartyPants spec implemented as an [extension in the RedCarpet markdown parser](https://github.com/vmg/redcarpet#also-now-our-pants-are-much-smarter), as noted in [Jekyll docs](https://jekyllrb.com/docs/configuration/markdown/#redcarpet).

To give some examples, that would mean turning text like this:

```
"some curly quotes" 'like this' -- they're pretty awesome --- yo
```

...into this:

> &ldquo;some curly quotes&rdquo; &lsquo;like this&rsquo; &ndash; they're pretty awesome &mdash; yo

By no means do I think this should be a default, but it would be interesting to add -- I love {en,em}-dashes in particular. :)

Keats:
@marcusklaas it looks like smart punctuation exists as an extension in the C reference parser, along with tests: https://github.com/commonmark/cmark/blob/master/test/smart_punct.txt

marcusklaas:
That's very good to know! This is a very good candidate for addition to pulldown as an extension.

The tricky part is the implementation. Smart punctuation would add a number of special tokens, quotes and dashes, that we'd need to scan for. That could have a non-trivial performance impact, since the main character scan routine is the costliest part of the parsing. This cost would even be apparent when the option is disabled, since many optimizations rely on this character set being static.

Do you have ideas on how we could solve this efficiently, @raphlinus? I recall you mentioned this problem when we added table support and its pipe character to the "new" pulldown.

cormacrelf:
Ok, I think I've solved the perf issue.

First, I added `'`, `"` and `-` to the special char lookup tables and `crdt_parse` was ~2-3% slower. About what I expected, and if I'm not mistaken you don't pay for much more if you condition the character match arm on `if self.options.contains(ENABLE_SMART) => {...}`.

However, I expected to see no change if `crdt.md` was stripped of these three characters before running the baseline and subsequent benchmarks, but there was still a 1.3% slowdown. Here's the complete diff of `cargo asm pulldown_cmark::parse::FirstPass::parse_line`:

```diff
diff --git a/parse_line_baseline b/parse_line_master
index 0b6d3c9..e4cc529 100644
[...]

PR #409, opened 2019-11-10
Change the special character lookup tables to statics

cormacrelf:
See https://github.com/raphlinus/pulldown-cmark/issues/119#issuecomment-552158266

In addition, this PR also implements LUT switching so that the performance of plain CommonMark is not affected by adding extensions. The full commit message explains the performance impact of that change.

Here's the overall effect:

```
⌁  g/pulldown-cmark ╍ (static-scanner-lut) cargo bench --bench html_rendering -- --baseline baseline crdt_parse
    Finished bench [optimized] target(s) in 0.16s
     Running target/release/deps/html_rendering-09cbd52ae1f5c0d2
crdt_parse              time:   [337.56 us 338.25 us 338.96 us]
                        change: [-2.0951% -1.8117% -1.5011%] (p = 0.00 < 0.05)
                        Performance has improved.

```

(I would also note these benchmarks were running on a pretty noisy machine, and the baseline fluctuated quite a lot later (still centred on where it was). You can at the very least say that plain-spec perf is no longer dependent on the number of extensions & extra special chars that need to be scanned to support them.)

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -2570,7 +2570,7 @@ where
     }
 }
 
-const fn special_bytes() -> [bool; 256] {
+static SPECIAL_BYTES: [bool; 256] = {
     let mut bytes = [false; 256];
     bytes[b'<' as usize] = true;
     bytes[b'!' as usize] = true;
@@ -2586,7 +2586,7 @@ const fn special_bytes() -> [bool; 256] {
     bytes[b']' as usize] = true;
     bytes[b'&' as usize] = true;
     bytes
-}
+};
 
 pub(crate) fn scalar_iterate_special_bytes<F, T>(
     bytes: &[u8],
@@ -2596,11 +2596,10 @@ pub(crate) fn scalar_iterate_special_bytes<F, T>(
 where
     F: FnMut(usize, u8) -> LoopInstruction<Option<T>>,
 {
-    let special_bytes = special_bytes();
 
     while ix < bytes.len() {
         let b = bytes[ix];
-        if special_bytes[b as usize] {
+        if SPECIAL_BYTES[b as usize] {
             match callback(ix, b) {
                 LoopInstruction::ContinueAndSkip(skip) => {
                     ix += skip;
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -2563,7 +2563,12 @@ pub(crate) enum LoopInstruction<T> {
 /// called and the function returns immediately with the return value `(end_ix, opt_val)`.
 /// If `BreakAtWith(..)` is never returned, this function will return the first
 /// index that is outside the byteslice bound and a `None` value.
-fn iterate_special_bytes<F, T>(lut: LookupTable, bytes: &[u8], ix: usize, callback: F) -> (usize, Option<T>)
+fn iterate_special_bytes<F, T>(
+    lut: LookupTable,
+    bytes: &[u8],
+    ix: usize,
+    callback: F,
+) -> (usize, Option<T>)
 where
     F: FnMut(usize, u8) -> LoopInstruction<Option<T>>,
 {
@@ -2628,7 +2633,6 @@ pub(crate) fn scalar_iterate_special_bytes<F, T>(
 where
     F: FnMut(usize, u8) -> LoopInstruction<Option<T>>,
 {
-
     while ix < bytes.len() {
         let b = bytes[ix];
         if lut[b as usize] {
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -266,6 +266,7 @@ struct FirstPass<'a> {
     allocs: Allocations<'a>,
     options: Options,
     list_nesting: usize,
+    lookup_table: LookupTable,
 }
 
 impl<'a> FirstPass<'a> {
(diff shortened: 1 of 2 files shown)


marcusklaas:
This is very interesting, and I applaud you for putting in the work. My machine produces very similar performance deltas as the ones you reported, which is encouraging.

I do need a bit more time to fully digest this. In particular, the following bit that you wrote [here](https://github.com/raphlinus/pulldown-cmark/issues/119#issuecomment-552158266),

> SIMD wasn't impacted by the static change, probably because the _mm_loadu_si128 call was itself inlined into a bunch of setup instructions based on the const.

is exactly in line with my understanding of how it gets to perform well. But I don't see how that can be preserved when the lookup table isn't fixed (without polymorphism).

The const to static change looks like a clean win and the rest seems sensible, I just need a bit more time to fully digest it!

src/parse.rs
```diff
@@ -277,6 +278,11 @@ impl<'a> FirstPass<'a> {
         let begin_list_item = false;
         let last_line_blank = false;
         let allocs = Allocations::new();
+        let lookup_table = if options == Options::empty() {
```
marcusklaas: Only the table and strikethrough options add new tokens for now, so we could just check those, right?

cormacrelf:
> I don’t see how that can be preserved when the lookup table isn’t fixed

@marcusklaas I did think about this, and I would guess it’s still inlined, but with lookups in each instruction rather than constants. I would have to study the asm more, but it’s comforting that there isn’t an overall regression. Are there actually realistic input patterns where SIMD is a lot faster rather than just slightly faster? For me the rest of the parse seems to dominate with the crdt_parses being 320 vs 350us. Maybe big tables.

src/parse.rs
cormacrelf: Yeah I’ll change that

cormacrelf:
(Just thinking - it might be slightly quicker if MaybeSimdLut was a pointer to a pair rather than a pair of pointers, as it would be 8 fewer bytes on the FirstPass struct, whose size seems to influence things.)

cormacrelf:
Oh, I remember what I meant — simd cannot reuse the setup work in the way the scalar version can — so it was only using the compile-time computation of what the setup should be called with not the setup itself, as the setup is done in registers. So we’re still reusing the precomputed shifts, nothing has changed. Hence no delta.

marcusklaas:
Okay, I've finally had another go at this. The performance boost should be greatest in the scenario where no options are set, right? I'm having trouble getting a consistent delta on `crdt_parse`, because the benchmark results are kind of noisy.

Thank you for your explanation on the SIMD stuff, that makes sense now.

It seems we should try to create a benchmark with a lot of optional tokens like `|` or `~` in them, to get a feel of how much we stand to gain.

cormacrelf:
Looking for a performance boost is not the best way to think about it. The layout changes in Parser will probably negate the construction or at least make them hard to measure. Maybe you'd see it with lots of optional tokens. But these are the two improvements:

1. **Removes some O(LUT size) setup code constructing the LUT in scalar parse_line.**
2. **Allows unextended CommonMark mode to avoid the cost of scanning characters it isn't interested in.**

Yes, (2) offers perf improvements in theory, but we don't scan that many extra special chars at present and they're not very common characters, so it's measurable in theory but I doubt anyone actually cares. So the benefit is mostly realised when adding features like smart punctuation. Because doing that increases LUT size AND requires scanning more characters. So a better comparison is between {a PR adding a feature like smart quotes, against where master is now} and {the same PR rebased against this PR}.

(I have such a PR cooking at https://github.com/raphlinus/pulldown-cmark/compare/master...cormacrelf:smartypants.)

cormacrelf:
A few pings on the issue brought me back here, and thinking about this anew, maybe varying the lookup table is best done with compile-time cargo features. That is, to enable a feature (and autocomplete that option in your editor) you must enable a cargo feature. Features == conditional LUT additions, plus some extra code enabled. Why?

1. I get a sense that while there’s a use case for runtime options, the scanner receiving loop can just ignore disabled options, rather than dynamically selecting a slower LUT and erasing some of the win with the LUT-switching. If you want full speed, compile with no features enabled. This would be the default. Pay for what you use by enabling it at compile time.
2. Markdown features tend to be additive/somewhat well-defined when enabled together. So this fits well enough as cargo features. Dep crates + your own Markdown needs should turn out okay, you’ll get the “slowest common denominator”.
2. I also get a sense that having knowledge of the options at compile time would be generally useful outside the LUT. Some checks and codepaths could be eliminated, and you might be able to tune pure CommonMark even faster. Does that sound accurate?

marcusklaas:
Hi Cormac, so sorry for the long radio silence here.

I am very keen on merging this soon. Smart punctuation is the last remaining major feature that pulldown has not implemented, and you've done excellent work here. Even if you've (understandably!) moved on, I'd like to take what you've done in this PR and the one you've linked to above and complete them to full smart punctuation support.

On the distinction between compile-time vs dynamic options: it's tricky. From a dogmatic perspective, I am in the compile-time camp like you. It seems that the use case for dynamically changing options is narrow, and we'd get optimal performance with compile-time options. Pragmatically however, it seems a bit problematic. It'd be a major breaking change, and going by your benchmarks, for very limited speed gains. I'm certainly open to it, but such a change would require broad support from other users and maintainers.

marcusklaas:
I took your ideas and used them to implement smart punctuation. Too bad this code didn't make it into master, but the effort wasn't wasted. Thanks for your PR!

cormacrelf:
I saw that, great job and I'm glad I could help!

PR #411, opened 2019-11-11
Ignore the benchmarks

dns2utf8:
So it can be published on crates.io

cc: @GuillaumeGomez

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,9 +18,9 @@ name = "pulldown-cmark"
 required-features = ["getopts"]
 doc = false
 
-[[bench]]
-name = "html_rendering"
-harness = false
+#[[bench]]
+#name = "html_rendering"
+#harness = false
 
 [dependencies]
 bitflags = "1.2"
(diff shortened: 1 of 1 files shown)


dns2utf8:
Enables https://github.com/rust-lang/rust/pull/65894

marcusklaas:
It seems we can keep the benchmarks if we remove it from the Cargo exclusion list, which seems like a nicer solution. Could you make that change in this PR, then I'll merge it and publish to crates.io.

dns2utf8:
I can do that.
On the other hand, why does the released version need the benchmark code?

marcusklaas:
Because the code commented out refers to benchmarks, but when they are excluded from the crate they cannot be found and it blocks the publication.

dns2utf8:
It seems that it is currently not possible to release a package without the benchmark code included.
Even if that code is not touched during build later.

marcusklaas:
Yup, seems that way. The current 0.6.1 release was done by including the benchmark code.

Thanks for your PR. If we find a way to exclude the benchmarks in the future, then let's revisit the subject.

dns2utf8:
There seems to be a discussion going on about maybe wanting to use included benchmarking code for compiler testing, but I am not sire that is the best idea.
Closing this PR for now. Thank you

PR #412, opened 2019-11-11
Remove benchmarks from Cargo exclusion list

marcusklaas:
When the benchmarks are excluded from the release, Cargo is not able
to locate the html_rendering benchmark and publication will fail.

There is something to be said for benchmark exclusion from releases,
but I don't see a way to exclude them from releases without removing
benchmarks altogether.


marcusklaas:
The dry run seems to work, so I have good hopes.

Issue #413, opened 2019-11-16
Footnote below unordered list wraps list item contents in paragraph elements


HanKruiger:
# Step to reproduce

- Have a footnote directly below an unordered list:

```markdown
Lorem ipsum.[^a]

An unordered list before the footnotes:
* Ipsum
* Lorem

[^a]: Cool.
```

# Expected output

```html
<p>Lorem ipsum.<sup class="footnote-reference"><a href="#a">1</a></sup></p>
<p>An unordered list before the footnotes:</p>
<ul>
    <li>Ipsum</li>
    <li>Lorem</li>
</ul>
<div class="footnote-definition" id="a"><sup class="footnote-definition-label">1</sup>
    <p>Cool.</p>
</div>
```

# Actual output

```html
<p>Lorem ipsum.<sup class="footnote-reference"><a href="#a">1</a></sup></p>
<p>An unordered list before the footnotes:</p>
<ul>
    <li><p>Ipsum</p></li>
    <li><p>Lorem</p></li>
    <div class="footnote-definition" id="a"><sup class="footnote-definition-label">1</sup>
        <p>Cool.</p>
    </div>
</ul>
```

I have added this (failing) test in my fork. See the comparison [here](https://github.com/HanKruiger/pulldown-cmark/compare/master...HanKruiger:add-footnote-test-after-ul), and feel free to use it or ignore it.

Cheers!

marcusklaas:
Thanks for filing this issue! This looks like a bug to me as well, but not for the same reason. The paragraphs within the list items is a natural consequence of the list no longer being [tight](https://spec.commonmark.org/0.29/#tight) because the footnote is considered part of the list, which seems to be the underlying problem.

I would expect this output:
```html
<p>Lorem ipsum.<sup class="footnote-reference"><a href="#a">1</a></sup></p>
<p>An unordered list before the footnotes:</p>
<ul>
    <li>Ipsum</li>
    <li>Lorem</li>
</ul>
<div class="footnote-definition" id="a"><sup class="footnote-definition-label">1</sup>
    <p>Cool.</p>
</div>
```
I suspect this is an ordering issue in `parse_block`, which has proven tricky to get right when footnotes were first added to the redesign. I'll look into it.

HanKruiger:
Oh yes, you're right. That's also the output I expected, but I didn't notice that the footnote was also inside the `<ul>`! I'll edit the post to make it clearer.

Thanks for taking a look!

[...]

PR #414, opened 2019-11-17
Properly close lists before opening a footnote

marcusklaas:
Closes GitHub issue #413.

### DIFF
--- a/specs/regression.txt
+++ b/specs/regression.txt
@@ -708,3 +708,25 @@ ISSUE 399
 the size of <code>usize</code> and have the same alignment.</p>
 </blockquote>
 ````````````````````````````````
+
+ISSUE 413
+
+```````````````````````````````` example
+Lorem ipsum.[^a]
+
+An unordered list before the footnotes:
+* Ipsum
+* Lorem
+
+[^a]: Cool.
+.
+<p>Lorem ipsum.<sup class="footnote-reference"><a href="#a">1</a></sup></p>
+<p>An unordered list before the footnotes:</p>
+<ul>
+    <li>Ipsum</li>
+    <li>Lorem</li>
+</ul>
+<div class="footnote-definition" id="a"><sup class="footnote-definition-label">1</sup>
+    <p>Cool.</p>
+</div>
+````````````````````````````````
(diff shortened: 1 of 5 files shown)


tests/suite/spec.rs
```diff
@@ -6889,8 +6889,7 @@ fn spec_test_511() {
 fn spec_test_512() {
     let original = r##"[link *foo **bar** `#`*](/uri)
 "##;
-    let expected =
-        r##"<p><a href="/uri">link <em>foo <strong>bar</strong> <code>#</code></em></a></p>
+    let expected = r##"<p><a href="/uri">link <em>foo <strong>bar</strong> <code>#</code></em></a></p>
```
ScottAbbey: I wonder why several of these lines changed...are there different formatting rules now?

specs/regression.txt
```diff
@@ -708,3 +708,25 @@ ISSUE 399
...
+<p>Lorem ipsum.<sup class="footnote-reference"><a href="#a">1</a></sup></p>
+<p>An unordered list before the footnotes:</p>
+<ul>
+    <li>Ipsum</li>
+    <li>Lorem</li>
+</ul>
+<div class="footnote-definition" id="a"><sup class="footnote-definition-label">1</sup>
+    <p>Cool.</p>
+</div>
+````````````````````````````````
```
ScottAbbey: When we have examples/tests specifically related to extra features like footnotes, is there a way we can mark them all so they can be reviewed more easily if the footnote spec is updated in the future?

tests/suite/spec.rs
marcusklaas: It seems like it. I recently moved to the most recent Rust stable distribution. It's unfortunate, because these changes are basically noise. Is there a way of dealing with this?

specs/regression.txt
marcusklaas: There currently is not, but that would be super useful. It should be easy enough to put these in separate files.

marcusklaas:
Thanks for the review @ScottAbbey! I will open an issue to split the regression tests for optional features. It seems very sensible, but is probably too large of a change to lump in with this PR.

PR #416, opened 2019-11-17
Add boolean to tell if it's an indented code block or not

### LINKED ISSUE #415: Include full codeblock start
For example, it's impossible to differentiate a codeblock indented with a codeblock between backticks and it's even worst with list. It could be fixed by providing a boolean saying if it's and indented codeblock or not.

GuillaumeGomez:
Fixes #415.

Once merged, can we have another release as quickly as possible please? I'd really love to be able to merge https://github.com/rust-lang/rust/pull/65894 (a few fixes depend on it as well).

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -2886,14 +2886,14 @@ mod test {
     #[cfg(target_pointer_width = "64")]
     fn node_size() {
         let node_size = std::mem::size_of::<Node<Item>>();
-        assert_eq!(48, node_size);
+        assert_eq!(56, node_size);
     }
 
     #[test]
     #[cfg(target_pointer_width = "64")]
     fn body_size() {
         let body_size = std::mem::size_of::<ItemBody>();
-        assert_eq!(16, body_size);
+        assert_eq!(24, body_size);
     }
 
     #[test]
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -3077,4 +3077,37 @@ mod test {
         }
         assert!(link_tag_count > 0);
     }
+
+    #[test]
+    fn code_block_kind_check_fenced() {
+        let parser = Parser::new("hello\n```test\ntadam\n```");
+        let mut found = 0;
+        for (ev, _range) in parser.into_offset_iter() {
+            match ev {
+                Event::Start(Tag::CodeBlock(CodeBlockKind::Fenced(fences), syntax)) => {
+                    assert_eq!(fences.as_ref(), "```");
+                    assert_eq!(syntax.as_ref(), "test");
+                    found += 1;
+                }
+                _ => {}
+            }
+        }
+        assert_eq!(found, 1);
+    }
+
+    #[test]
+    fn code_block_kind_check_indented() {
+        let parser = Parser::new("hello\n\n    ```test\n    tadam\nhello");
+        let mut found = 0;
+        for (ev, _range) in parser.into_offset_iter() {
+            match ev {
+                Event::Start(Tag::CodeBlock(CodeBlockKind::Indented, syntax)) => {
+                    assert_eq!(syntax.as_ref(), "");
+                    found += 1;
+                }
+                _ => {}
+            }
+        }
+        assert_eq!(found, 1);
+    }
 }
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -42,6 +42,7 @@ const LINK_MAX_NESTED_PARENS: usize = 5;
 #[derive(Clone, Debug, PartialEq)]
 pub enum CodeBlockKind<'a> {
     Indented,
+    /// The value contained in the tag describes the language of the code, which may be empty.
     Fenced(CowStr<'a>),
 }
 
@@ -59,13 +60,6 @@ impl<'a> CodeBlockKind<'a> {
             _ => false,
         }
     }
-
-    pub fn get_fences(&self) -> Option<&CowStr<'a>> {
-        match *self {
-            CodeBlockKind::Fenced(ref f) => Some(f),
-            _ => None,
-        }
-    }
 }
 
 /// Tags for elements that can contain other elements.
@@ -79,11 +73,7 @@ pub enum Tag<'a> {
 
     BlockQuote,
     /// A code block.
-    ///
-    /// The boolean is `true` is this is an indented code block (not starting with "\`\`\`").
-    ///
-    /// The value contained in the tag describes the language of the code, which may be empty.
-    CodeBlock(CodeBlockKind<'a>, CowStr<'a>),
+    CodeBlock(CodeBlockKind<'a>),
 
     /// A list. If the list is ordered the field indicates the number of the first item.
     /// Contains only list items.
@@ -237,7 +227,7 @@ enum ItemBody {
 
     Rule,
     Heading(u32), // heading level
-    FencedCodeBlock(CowIndex, CowIndex),
+    FencedCodeBlock(CowIndex),
     IndentCodeBlock,
     Html,
     BlockQuote,
@@ -1045,12 +1035,10 @@ impl<'a> FirstPass<'a> {
         let mut ix = info_start + scan_nextline(&bytes[info_start..]);
         let info_end = ix - scan_rev_while(&bytes[info_start..ix], is_ascii_whitespace);
         let info_string = unescape(&self.text[info_start..info_end]);
-        let fences = &self.text[start_ix..info_start];
         self.tree.append(Item {
             start: start_ix,
             end: 0, // will get set later
             body: ItemBody::FencedCodeBlock(
-                self.allocs.allocate_cow(fences.into()),
                 self.allocs.allocate_cow(info_string),
             ),
         });
@@ -2712,13 +2700,10 @@ fn item_to_tag<'a>(item: &Item, allocs: &Allocations<'a>) -> Tag<'a> {
             Tag::Image(*link_type, url.clone(), title.clone())
         }
         ItemBody::Heading(level) => Tag::Heading(level),
-        ItemBody::FencedCodeBlock(fences_ix, cow_ix) => {
-            Tag::CodeBlock(
-                CodeBlockKind::Fenced(allocs[fences_ix].clone()),
-                allocs[cow_ix].clone(),
-            )
+        ItemBody::FencedCodeBlock(cow_ix) => {
+            Tag::CodeBlock(CodeBlockKind::Fenced(allocs[cow_ix].clone()))
         }
-        ItemBody::IndentCodeBlock => Tag::CodeBlock(CodeBlockKind::Indented, "".into()),
+        ItemBody::IndentCodeBlock => Tag::CodeBlock(CodeBlockKind::Indented),
         ItemBody::BlockQuote => Tag::BlockQuote,
         ItemBody::List(_, c, listitem_start) => {
             if c == b'.' || c == b')' {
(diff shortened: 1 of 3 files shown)


src/parse.rs
```diff
@@ -48,9 +48,12 @@ pub enum Tag<'a> {
     Heading(u32),
 
     BlockQuote,
-    /// A code block. The value contained in the tag describes the language of the code,
-    /// which may be empty.
-    CodeBlock(CowStr<'a>),
+    /// A code block.
+    ///
+    /// The boolean is `true` is this is an indented code block (not starting with "\`\`\`").
```
oberien:
```suggestion
    /// The boolean is `true` if this is an indented code block (not starting with ```` ``` ```` or `~~~`).
```
Rendered (with GitHub Markdown): /// The boolean is `true` if this is an indented code block (not starting with ```` ``` ```` or `~~~`).

marcusklaas:
Normally I'd be okay with merging a small change like this, but since it is a compatibility break, we should be a bit more careful.

First, could you elaborate on when the distinction between indented and delimited codeblocks is relevant? I see what you wrote [here](https://github.com/rust-lang/rust/pull/65894#discussion_r347155406), but I don't quite get what rustdoc does there. Ideally, we'd only really expose semantic differences and leave users to dig into the source map if they care about constitutes a code block, even if that feels a bit hacky.

Secondly, since this is a public interface, some care should be taken with respect to the design. A boolean works for sure, but @oberien's suggestion may be more future proof if we can find a purpose for the size of the delimiters. Wrapping the `CowStr<'a>` in an option could also work.

Finally, it's a bit uncomfortable to do a major release so soon after `0.6` with only very minor changes. I'm not opposed to it, but we should at least clearly identify the reasoning for the bump. If we can do that, I'm sure we can settle on a design quickly (FLW, I know)

GuillaumeGomez:
I saw a very funny case where an indented code block started with "\`\`\`". Since we are parsing this to check if we're supposed to compile it and test it, we're in a problematic situation because we don't know if it's actually an indented code block. If we had this information, we wouldn't need to parse and just assume it's a code block.

With the small hack I wrote to know this, it's an issue in case a code block is inside a list.

So like @oberien suggested, I'll extend this PR a bit in order to also provide the string that makes this code block as a code block ("\`\`\`" for example). And yes, it'll require a medium release. :)

marcusklaas:
So rustdoc only compiles delimited codeblocks (starting with either `` ``` `` or `~~~~`) but not indented codeblocks, is that correct?

GuillaumeGomez:
No we compile both. But the difference is that you can give instructions on non-indented codeblocks.

marcusklaas:
How are these instructions encoded? Are they in the language snippet of the codeblock?

GuillaumeGomez:
Yes, looks like this for example:

"\`\`\`compile_fail,edition2018,E0594"

marcusklaas:
Cool, thanks. Are delimited codeblocks without instructions treated like indented codeblocks? Because you could just go off the language snippet then, right?

GuillaumeGomez:
The same way at least. If no additional data is provided, we just assume it's a rust codeblock.

oberien:
As marcus said, in your given example ```` ```compile_fail,edition2018,E0594 ````, you can parse the `lang` attribute of the code-block. That's also [what I'm doing in heradoc](https://github.com/oberien/heradoc/blob/master/src/frontend/mod.rs#L318), where I also add custom attributes you can use within the language definition of the code-block.

GuillaumeGomez:
We do. The problem is that we still need to know if we *have to*.

GuillaumeGomez:
Ok, I updated the PR to include the fence string as well (using an enum). I also added some tests. I think it's ready?

GuillaumeGomez:
Any news?

marcusklaas:
It's still not exactly clear to me what problem this solves. If you could provide a comprehensive summary that would be very useful.

GuillaumeGomez:
"\`\`\`" should be consided content in a codeblock:

``````
    ```
``````

This one should be not indented code block:

``````
1. this is a list!
2. Yeay
    ```rust
    content
    ```
``````

In both cases, I need to know if I am in an indented codeblock or not to know if I need to check the "\`\`\`" content. I found a trick by checking if whitespaces were present before the codeblock start but I realized it would fail in case it was in a list.

To provide even more context: if the "syntax" is empty, we assume by default this is rust. We then check for this "\`\`\`" to see if extra information has been passed down. This is the issue.

**Comment by marcusklaas** [comment]
I still don't understand. What do you mean by "```" content? And by this extra information, do you mean stuff like `compile_fail,edition2018,E0594`? That's in the syntax string right? It's not yet clear to me why you can't just check the syntax string and pass that along, or default to rust when it's not set.

**Comment by GuillaumeGomez** [comment]
I'm bad at explaining. Please read the link to the rustdoc PR in the first comment to have full information. That might be better than me trying to resume the situation...

**Comment by ehuss** [comment]
I can try to clarify.  When rustdoc fails to parse a code block, it has a suggestion to add the `text` annotation. But it only gives that suggestion for fenced code blocks (with \`\`\`), since it wouldn't make sense for indented code blocks.

The problem is that it can be tricky to determine if something is a fenced code block or indented code block.  You can't just check if the code block starts with \`\`\`, because someone could have an indented code block that starts with \`\`\`.  Due to some changes in the 0.6 API, this is now even harder than before.

I'm still not convinced my suggestion given in https://github.com/rust-lang/rust/pull/65894#discussion_r347158530 wouldn't work.  That is, check if `Event::Start(Tag::CodeBlock(_))` and `Event::Text(_)` start at the same offset, then it should probably be an indented code block.

It would be cleaner if pulldown-cmark indicated which kind of code block it was, but I also think that semver-breaking changes are disruptive.  I think a temporary workaround would be better, and then make this change (or something like it) when there are enough changes to justify a bump to 0.7.

**Comment by GuillaumeGomez** [comment]
Thanks @ehuss !

I don't see the problem to make a medium release with such small amount of changes. It's just a release and they're pretty easy to do. Anyway, I'd love to finally be able to update this rustdoc's dependency.

**Comment by ehuss** [comment]
> I don't see the problem to make a medium release with such small amount of changes.

0.x changes are the same as major version changes.  It's a problem for projects like mdbook, where plugins need to keep semver-compatible versions in sync.  It also requires all projects to manually opt-in to the new version.

**Comment by GuillaumeGomez** [comment]
If they want to use it, sure. But nothing forces them to.

**Comment by marcusklaas** [comment]
Thanks @ehuss, that did the trick!

> It would be cleaner if pulldown-cmark indicated which kind of code block it was, but I also think that semver-breaking changes are disruptive. I think a temporary workaround would be better, and then make this change (or something like it) when there are enough changes to justify a bump to 0.7.

I feel much the same way. Exposing the distinction between indented and delimited codeblocks seems to make sense for the next major version change, but I'm very hesitant to release straight away.

Concerning the implementation: do we need to expose the delimiters themselves? Those could be recovered using the sourcemap, right? We could just wrap the syntax string in an option, where `Some(syntax_string)` would indicate a delimited codeblock and `None` an indented one.

**Comment by GuillaumeGomez** [comment]
Well, we're forced to remain on an older buggy pulldown-cmark version on rustdoc so I personally see a lot of reasons to make an update with "just" this.

> Concerning the implementation: do we need to expose the delimiters themselves? Those could be recovered using the sourcemap, right? We could just wrap the syntax string in an option, where `Some(syntax_string)` would indicate a delimited codeblock and `None` an indented one.

I like your suggestion but I find it a bit less clear from just reading the code. So as you prefer. Just tell me if I need to update the current PR.

**Comment by GuillaumeGomez** [comment]
Any news?

**Comment by marcusklaas** [comment]
No new insights here. I believe the way forward is to wrap the syntax string in either an `Option` or a bespoke enum. We could then include this patch in the next major release. I understand there is an effort under way to add smart punctuation, which may be a good feature to do a release for.

**Comment by GuillaumeGomez** [comment]
I prefer the enum way then: more explicit from my point of view. So I guess this PR is ready from my point of view. Waiting this next very eagerly.

**Comment by marcusklaas** [comment]
There doesn't seem to be a need to add the fence. A codeblock could just contain a `CodeBlockKind`, where the `CowStr` refers to the syntax string instead of the fence.

**Comment by GuillaumeGomez** [comment]
Ok, updated.

File: src/parse.rs [hunk 2]
```diff
@@ -204,7 +234,7 @@ enum ItemBody {
 
     Rule,
     Heading(u32), // heading level
-    FencedCodeBlock(CowIndex),
+    FencedCodeBlock(CowIndex, CowIndex),
```
marcusklaas: The fence is still stored here, which is unnecessary and increases the size of a `Node<Item>` by a word.

src/parse.rs
```diff
@@ -38,6 +38,37 @@ use crate::tree::{Tree, TreeIndex, TreePointer};
...
+    }
+
+    pub fn is_fenced(&self) -> bool {
+        match *self {
+            CodeBlockKind::Fenced(_) => true,
+            _ => false,
+        }
+    }
+
+    pub fn get_fences(&self) -> Option<&CowStr<'a>> {
```
marcusklaas: This returns the syntax string and not the fence, right?

marcusklaas: Since the enum is so simple, I'd argue we could drop the `impl` block altogether.

GuillaumeGomez: I'll keep the two other methods, it's pretty convenient. :)

GuillaumeGomez:
Removed the unneeded parts.

GuillaumeGomez:
Ping?

src/parse.rs
```diff
@@ -38,6 +38,30 @@ use crate::tree::{Tree, TreeIndex, TreePointer};
 // https://spec.commonmark.org/0.29/#link-destination
 const LINK_MAX_NESTED_PARENS: usize = 5;
 
+/// Codeblock kind.
+#[derive(Clone, Debug, PartialEq)]
+pub enum CodeBlockKind<'a> {
```
marcusklaas: This definition should probably be re-exported [here](https://github.com/raphlinus/pulldown-cmark/blob/master/src/lib.rs#L73), as there is currently no way to extract the language string.

GuillaumeGomez:
Done!

GuillaumeGomez:
Any news about a new release?

Others:
Hi, I'm wondering about the status of:
rust-lang/rust#66035 (an error improvement)
which is blocked on:
rust-lang/rust#65894 (upgrading `pulldown-cmark`)
which is (in my understanding) blocked on a new release of `pulldown-cmark`

@marcusklaas is there any work that I can help with that will move us towards a `pulldown-cmark` release? It's now been three months since the last release of the crate, so I do not think another release would be premature :)

marcusklaas:
Sorry for the delay in replying here. We could do a release, but because of a very minor API change it would have to be a 'major' version bump. I see several options:

- simply bump a major version and do a release
- wait until a new feature lands, like smart punctuation or custom header ids. This could take a while.
- do a minor release of the master branch before the breaking change

I'm not sure what's optimal. Thoughts?

cc @GuillaumeGomez @raphlinus @ScottAbbey

GuillaumeGomez:
Why not making a major release anyway and another later? I don't have a problem with it. :)

marcusklaas:
The 0.7.0 release is finally out!

GuillaumeGomez:
Awesome, thanks!

PR #420, opened 2020-01-04
Allow paragraphs before tables

terhechte:
I'm running into all kinds of corner cases with one of my projects, this being another one. It is even more tricky because tables are not part of CommonMark so it is less clear what the right behavior is. The problem this PR solves is that a table that has a paragraph right before it will not be parsed:

```
Hello
| abc | def |
| --- | --- |
| bar | baz |
```

Becomes a long paragraph the table syntax as part of the paragraph.

This PR fixes that and inserts a `Hello` paragraph before the table in the parsed markdown. In my project, I kind of rely on the possibility of having a paragraph before an element which is how I ran into this issue. I did some research to see how other projects, notably GitHub render the example above.
Interestingly, GitHub exhibits the same issue in their markdown parsing, while [this online editor](https://jbt.github.io/markdown-editor/) supports it just fine.

However, other examples, such as

```
# Hello
| abc | def |
| --- | --- |
| bar | baz |
```

render correctly on GitHub. It is only paragraphs, that are not supported right before a table. Therefore, at least to me, this feels more like an accidental bug than correct behavior.

Personally, for me, it would be great if this would be merged, as I'd need to continue using a fork otherwise. Another, maybe, viable option might be to have this behavior be optional via a feature flag (although I can see that this is a road that, once taken, can lead to madness if more and more of the non-standardized idiosyncratic markdown _features_ end up as feature flags).

This PR also contains tests, and I'd be more than happy to improve the code if it is not up to the quality standards of pulldown-cmark.

### DIFF
--- a/src/firstpass.rs
+++ b/src/firstpass.rs
@@ -201,6 +201,11 @@ impl<'a, 'b> FirstPass<'a, 'b> {
         if let Some((n, fence_ch)) = scan_code_fence(&bytes[ix..]) {
             return self.parse_fenced_code_block(ix, indent, fence_ch, n);
         }
+
+        if let Some(n) = scan_options_start(&bytes[ix..]) {
+            return self.parse_options_block(ix, n);
+        }
+
         self.parse_paragraph(ix)
     }
 
@@ -416,6 +421,21 @@ impl<'a, 'b> FirstPass<'a, 'b> {
                 match byte {
                     b'\n' | b'\r' => {
                         if let TableParseMode::Active = mode {
+                            return LoopInstruction::BreakAtWith(ix, None);
+                        }
+
+                        // If tables are not enabled yet and the next char is a pipe,
+                        // and there was content before then a table begins here but
+                        // there was content before the table:
+                        // ```
+                        // Hello
+                        // | a | b | c |
+                        // ```
+                        if TableParseMode::Scan == mode &&
+                            bytes.len() > (ix + 1) &&
+                            bytes[ix + 1] == b'|' &&
+                            begin_text < ix &&
+                            pipes == 0 {
                             return LoopInstruction::BreakAtWith(ix, None);
                         }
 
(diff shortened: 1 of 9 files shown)


oberien:
The behaviour of pulldown-cmark is the same as github. It's just how Github-flavoured Markdown (GFM) defines tables.
In general, things in pulldown-cmark that are not part of CommonMark are usually part of GFM.

```md
Text
|foo|bar|
--- | ---
baz | qux
```
converts to 

Text
|foo|bar|
--- | ---
baz | qux

while
```md
Text

|foo|bar|
--- | ---
baz | qux
```
converts to

Text

|foo|bar|
--- | ---
baz | qux

marcusklaas:
Thank you for the pull request. It's not clear to me whether the [GFM](https://github.github.com/gfm/#tables-extension-) spec or CommonMark actually prescribe how to handle this. If not, we should probably ask for explicit clarification.

ScottAbbey:
> If not, we should probably ask for explicit clarification.

Notice that each other Leaf Block type has at least one specific example that tells whether or not that type can interrupt a paragraph.  The authors of the GFM tables extension failed to state whether that type can interrupt a paragraph or not.

mity:
See also this: https://github.com/github/cmark-gfm/issues/141

Apparently some time ago cmark-gfm tried to allow that tables can interrupt the preceding paragraph but my testing shows they still have some bugs related to it.

oberien:
My older comment (https://github.com/raphlinus/pulldown-cmark/pull/420#issuecomment-570826575) didn't seem to age too well. Apparently now it just works in GFM on github?!

mity:
@oberien It works **sometimes**.

This works:
```
foo
|  a  |  b  |
| --- | --- |
|  c  |  d  |
```

> foo
> |  a  |  b  |
> | --- | --- |
> |  c  |  d  |



This does not:
```
foo
 a  |  b
--- | ---
 c  |  d
```

> foo
>  a  |  b
> --- | ---
>  c  |  d

expectocode:
@mity it seem like Github has updated once again: your comment now renders a paragraph followed by a table both times. (screenshot below)
image

mity:
Interesting. Do they still use cmark-gfm? Because https://github.com/github/cmark-gfm has seen the last commit in September 2019.

Martin1887:
I see here different features in the same pull request, @terhechte. Tables interrupting paragraphs is a feature to be included in the 0.10 release, since it is the GitHub Flavor Markdown behavior, but the other three commits should be removed before merging.

The best approach is probably to create a new pull request from the current master branch and to close this pull request without merging.

Issue #430, opened 2020-02-15
into_offset_iter() does not give correct range for link references


dsherret:
For example, given the following markdown:

```markdown
# H1
[testing](https://github.com)
[testing][Some reference]

[Some reference]: https://github.com
```

The events and ranges are like so:

```
Event: Start(Heading(1))
Range: 0..5
Event: Text(Borrowed("H1"))
Range: 2..4
Event: End(Heading(1))
Range: 0..5
Event: Start(Paragraph)
Range: 5..61
Event: Start(Link(Inline, Borrowed("https://github.com"), Borrowed("")))
Range: 5..34
Event: Text(Borrowed("testing"))
Range: 6..13
Event: End(Link(Inline, Borrowed("https://github.com"), Borrowed("")))
Range: 5..34
Event: SoftBreak
Range: 34..35
Event: Start(Link(Reference, Borrowed("https://github.com"), Borrowed("")))
Range: 35..36
Event: Text(Borrowed("testing"))
Range: 36..43
Event: End(Link(Reference, Borrowed("https://github.com"), Borrowed("")))
Range: 35..36
```

So for inline links it will give the full range of `5..34`, which is correct. The reference link is incorrect though as it gives `35..36` instead of `35..60`.

~~I can look into this issue.~~ The only reason I need this though is because I am manually parsing out link reference text because it's not provided another way. Edit: Actually, I'm just going to work around this as I don't have time at the moment to fix it.

marcusklaas:
Good catch, and thank you for reporting it. I shall investigate.

dsherret:
Also, `FootnoteReference` does not have the full width:

```markdown
Testing this[^1] out.

[^1]: Footnote.
```

Outputs:

```
Event: Start(Paragraph)
Range: 0..22
Event: Text(Borrowed("Testing this"))
Range: 0..12
Event: FootnoteReference(Borrowed("1"))
Range: 12..13 <-- should be 12..16
Event: Text(Borrowed(" out."))
Range: 16..21
Event: End(Paragraph)
Range: 0..22
Event: Start(FootnoteDefinition(Borrowed("1")))
Range: 23..39
Event: Start(Paragraph)
Range: 29..39
Event: Text(Borrowed("Footnote."))
Range: 29..38
Event: End(Paragraph)
Range: 29..39
Event: End(FootnoteDefinition(Borrowed("1")))
Range: 23..39
[...]

PR #432, opened 2020-02-17
Issue 430

marcusklaas:
Addresses https://github.com/raphlinus/pulldown-cmark/issues/430.

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -1675,8 +1675,8 @@ impl InlineStack {
 
 #[derive(Debug, Clone)]
 enum RefScan<'a> {
-    // label, next node index
-    LinkLabel(CowStr<'a>, TreePointer),
+    // label, next node index, source ix of label end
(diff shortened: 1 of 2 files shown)


Gozala:
I'm noticing that this had being around since February. Does it needs more work before it can be merged it ?

marcusklaas:
I think it's ready to go, just needs a review.

@raphlinus, do you have time for this?

alerque:
I was about to try to hack around this downstream ... it would be really nice to see this merged.

alerque:
I've compiled this PR and checked it against what I expected the ranges to be and it is now returning end bytes for Reference links similar to the way Inline links currently work. I can't speak for every possible scenario but at least that bit seems right here.

marcusklaas:
Cool, thanks for doing a testrun @alerque! Let us know if you find any other issues :-)

Issue #401, opened 2019-09-13
Experiment with arenas


marcusklaas:
In pulldown_cmark, the main data structure is a vector backed tree. The final tree which is traversed to generate events in two passes. Roughly speaking, the tree structure is constructed in the first pass (blocks) and the leaf nodes are finalized in the second pass. Especially in this second pass, many tree mutations are done which correspond to indexed vector accesses. These operations incur the cost of bound checking in Rust, which could possibly be avoided by replacing the vector by an efficient arena. One such candidate could be [id_arena]. This would a great display of getting maximal performance while still enjoying Rust's safety guarantees.

There may be a tricky bit in the emphasis handling part of the second pass, where we currently construct some vector indices with arithmetic, which will not work with arena ids. One way to get around this may be to ask the arena to "upgrade" a numeric value to a valid arena id. This would do a one-time bounds check, after which we could use the id for access without checks after. It may also be possible to simply reuse arena ids that are already in the tree.

Another place where vector backed stores without deletion are used is the `Allocations` struct. It contains details of links, images, owned strings and table alignments. Tree nodes may store keep indexes to the `Allocations` struct, which are much smaller than the details themselves. These vector backed datastructures are also prime candidates for arena replacement. We don't do any index arithmetic for those.

[id_arena]: https://docs.rs/id-arena/2.2.1/id_arena/

PR #433, opened 2020-03-19
Faster CowStr

maciejhirsz:
+ Replaced the `CowStr` with `beef::lean::Cow`, which is a 2-word wide `std::borrow::Cow` replacement.
+ Turned `scan_entity` into a method on an `EntityScanner` struct. The struct holds a `[u8; 4]` buffer internally so it can write `char` into it and borrow `&str`s. Return type is now `(usize, Option<&str>)`.

I think I can do better still, but I'm open to feedback on this first, quick step.


src/parse.rs
```diff
@@ -855,13 +856,13 @@ impl<'a> FirstPass<'a> {
...
                 }
-                b'&' => match scan_entity(&bytes[ix..]) {
+                b'&' => match scanner.scan_entity(&bytes[ix..]) {
                     (n, Some(value)) => {
                         self.tree.append_text(begin_text, ix);
                         self.tree.append(Item {
                             start: ix,
                             end: ix + n,
-                            body: ItemBody::SynthesizeText(self.allocs.allocate_cow(value)),
+                            body: ItemBody::SynthesizeText(self.allocs.allocate_cow(value.to_string().into())),
```
maciejhirsz: Codepoint `&str`s are transient so this unfortunately requires an allocation. I don't know what the `self.allocs` is for at this point, but I think this might be avoidable.

maciejhirsz:
Unescapes are hurting due to the allocation pointed out above. Benches, relative to master on my machine. It's pretty RNG with my CPU, but overall consistently it's consistently faster.

```
Gnuplot not found, using plotters backend
crdt_total              time:   [218.59 us 219.44 us 220.32 us]                       
                        change: [-26.345% -26.021% -25.688%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 6 outliers among 100 measurements (6.00%)
  2 (2.00%) high mild
  4 (4.00%) high severe

crdt_html               time:   [95.406 us 95.537 us 95.696 us]                      
                        change: [-20.034% -19.840% -19.646%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high mild

crdt_parse              time:   [160.50 us 160.94 us 161.61 us]                       
                        change: [-13.110% -12.150% -11.166%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 5 outliers among 100 measurements (5.00%)
  3 (3.00%) high mild
  2 (2.00%) high severe

links_n_emphasis        time:   [2.6204 us 2.6250 us 2.6305 us]                              
                        change: [+3.1111% +3.4208% +3.7429%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 4 outliers among 100 measurements (4.00%)
  1 (1.00%) high mild
  3 (3.00%) high severe

unescapes               time:   [14.665 us 14.690 us 14.718 us]                       
                        change: [+63.946% +64.518% +65.048%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 3 outliers among 100 measurements (3.00%)
  1 (1.00%) high mild
  2 (2.00%) high severe

autolinks_n_html        time:   [2.7307 us 2.7391 us 2.7474 us]                              
                        change: [-24.499% -24.208% -23.897%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 100 measurements (2.00%)
  1 (1.00%) high mild
  1 (1.00%) high severe

     Running target/release/deps/lib-5114c3e6f5288098

running 2 tests
test to_html::advanced_pathological_codeblocks ... bench:     245,749 ns/iter (+/- 7,968)
test to_html::pathological_codeblocks1         ... bench:     595,396 ns/iter (+/- 36,942)

test result: ok. 0 passed; 0 failed; 0 ignored; 2 measured; 0 filtered out
```

marcusklaas:
This is interesting work. If we could achieve speedups close to what you are seeing, that would definitely outweigh the lost speed on relatively rare unescapes.

I'm not getting the same numbers as you are however. This is my output from `cargo bench`:
```text
crdt_total              time:   [391.32 us 392.19 us 393.14 us]                       
                        change: [-1.3258% -0.7319% -0.1992%] (p = 0.01 < 0.05)
                        Change within noise threshold.
Found 10 outliers among 100 measurements (10.00%)
  1 (1.00%) low mild
  6 (6.00%) high mild
  3 (3.00%) high severe

crdt_html               time:   [169.21 us 169.37 us 169.54 us]                      
                        change: [-12.929% -12.551% -12.153%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 7 outliers among 100 measurements (7.00%)
  1 (1.00%) low mild
  2 (2.00%) high mild
  4 (4.00%) high severe

crdt_parse              time:   [237.51 us 237.81 us 238.17 us]                       
                        change: [+2.2821% +2.7400% +3.1748%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 14 outliers among 100 measurements (14.00%)
  3 (3.00%) low mild
  4 (4.00%) high mild
  7 (7.00%) high severe

links_n_emphasis        time:   [4.0782 us 4.0878 us 4.0972 us]                              
                        change: [-0.4782% +0.2343% +0.9198%] (p = 0.52 > 0.05)
                        No change in performance detected.
Found 9 outliers among 100 measurements (9.00%)
  1 (1.00%) low severe
  2 (2.00%) low mild
  4 (4.00%) high mild
  2 (2.00%) high severe

unescapes               time:   [23.210 us 23.396 us 23.638 us]                       
                        change: [+77.249% +78.178% +79.350%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 12 outliers among 100 measurements (12.00%)
  2 (2.00%) low severe
  2 (2.00%) high mild
  8 (8.00%) high severe

autolinks_n_html        time:   [5.1786 us 5.1890 us 5.1994 us]                              
                        change: [-0.5561% -0.2451% +0.0666%] (p = 0.13 > 0.05)
                        No change in performance detected.
Found 7 outliers among 100 measurements (7.00%)
  2 (2.00%) low mild
  4 (4.00%) high mild
  1 (1.00%) high severe

     Running target/release/deps/lib-076f3121abafd46a

running 2 tests
test to_html::advanced_pathological_codeblocks ... bench:     417,954 ns/iter (+/- 22,749)
test to_html::pathological_codeblocks1         ... bench:   1,314,135 ns/iter (+/- 47,122)

test result: ok. 0 passed; 0 failed; 0 ignored; 2 measured; 0 filtered out
```

It may be due to a different CPU architecture? I'm on an ancient AMD 780K, which may not benefit from the change as much as more modern CPUs.

maciejhirsz:
Curious. I'm running i7-8850H, which is a pain to benchmark with since there is quite some variation in throughput between cores.

In any way, I reckon I can use [an arena](https://crates.io/crates/toolshed) to avoid allocations, maybe even make it conditional so that the `'static` escapes aren't affected. Might a good first step towards #401 too.

maciejhirsz:
Ok, I've done some [experiments with arena](https://github.com/maciejhirsz/pulldown-cmark/pull/1/files). It does solve the performance issue on unescapes (as expected), but there are some major soundness issues here:

Because the `Parser` is exposed in public API and not just a temporary internal thing, for arena allocations to work these invariants must be true:
+ The arena must be contained within `Parser`, which makes borrowing slices out of it extremely problematic: because the borrowed slices have to live for lifetime `'a` which is the lifetime of the source, but the arena must be contained inside the `Parser` which in theory can be dropped before the source is dropped, any `&'a str` which borrows from the arena (instead of the source) that still exists would introduce UB.
+ `Parser` **musn't** implement `Clone`.

Good news is, since `pulldown_cmark` is already using indexes instead of references across its tree, we can get around all of this, and not have to introduce any unsafe code, with something that looks as follows:

```rust
struct ArenaStr {
    offset: usize,
    len: usize,
}

struct ArenaPtr<T: Copy> {
    offset: u64,
    marker: PhantomData<T>,
}

impl Arena {
    fn alloc_str(&mut self, transient: &str) -> ArenaStr {...}
    fn alloc<T: Copy>(&mut self, value: T) -> ArenaPtr<T> {...}
    fn get_str(&self, aref: ArenaStr) -> &str {...}
    fn get<T: Copy>(&self, aref: ArenaPtr<T>) -> &T {...}   
}
```

To make this _really_ sound, we could imagine using an anonymous type `X` (dummy closure could be abused here), so that `Arena<X>`, `ArenaStr<X>`, `ArenaPtr<X, T>` can only interact with one another if `X` is the same type, basically forcing Rust to keep track of which reference belongs to which arena, so we don't have to make the `get` and `get_str` return Options, and avoid all potential UB there.

The `Arena` itself could use a simple `Vec<u8>` as a backend, the only catch is that `alloc` would have to keep track of proper alignment for `T: Copy`, but that's trivial. Since we don't actually return any `&` refs, there are no issues with lifetimes. `Clone` would still be out of the question, we would have to clone the `Arena<X>` with a different anonymous type `Arena<Y>`, which I don't think is possible in the Rust type system.

It should also be possible to provide a `ArenaStrBuf<'a, X>` that mutably borrows `Arena<X>` and allows you to construct a single `ArenaStr<X>` by pushing transient `&str`s onto it, thus replacing all the use of temporary `String`s in the parser.

With all of this done:
+ We can remove any in-place allocations, and instead relegate them to the `Arena`, which might already have enough capacity. We can almost guarantee that it does if we create an `Arena` with capacity == length of source string before we start parsing.
+ `CowStr` would have to be replaced with a new enum:
```rust
enum StrBorrow<'a, X> {
    Source(&'a str),
    Arena(ArenaStr<X>),
}
```
+ This re-introduces branching on obtaining a readable `&str` reference, but avoids branching when dropping (which `Cow` has to do).
+ Implementing this would be still small enough in scope to not warrant a complete refactor of the code, while opening the codebase to further optimizations - in particular putting the entire `Tree` into the `Arena`.

I'm happy to have my go at implementing this. I reckon for starters we would only need the `alloc_str` and `get_str` without the arbitrary type `T`, which should make things even easier. Whatcha think @marcusklaas?

maciejhirsz:
Here is an example of what I mean by abusing the closures to make sure the Arenas are not compatible with one another: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=7decfe8bb14e7261ad23f38f4fa5d795

The catch here is that this type would have to be propagated up to the `Parser<X>`, so obtainig a new parser might require a return signature `-> Parser<'a, impl Copy>`, not sure if that's something we want to burden users with?

maciejhirsz:
Ok, so I did some experiments and pushed to the branch with an arena implementation that works well enough for the strings. On my machine it is faster in all benchmarks now. That said, there are soundness issues...

To make it UB free the API would have to be changed:

```rust
let events: Vec<_> = Parser::new(input).collect();
```

Since the `events` outlive the `Parser`, any arena allocated strings would be dangling pointers. A solution here would require splitting the iterator out of the `Parser`, like so:

```rust
// Create a parser from a source with some lifetime 'a
let mut parser = Parser::new(input);

// Create a parser iterator from `parser`, which borrows it for lifetime 'b
let mut iterator = parser.iter()

// Iterator is dropped, however the Vec of events still borrows the parser for lifetime 'b
let events: Vec<_> = iterator.collect();
```

skanfd:
Sorry to interrupt.
But why we have a `CowStr` other than `std::borrow::Cow`?
And what is a `InlineStr`? Why it has 3 words long but stores  a `len` in it's last byte?
Does it exists to accelerate the situation that string is shorter than 3 words?

Martin1887:
Any update about this? I think it could be closed.

maciejhirsz:
@Martin1887 there might be something here, but by now the PR is ancient and would need a new one anyway. Plus it seems like `Cow<str>` will get the 3-word optimization in std anyway.

Issue #434, opened 2020-03-29
Include reference link identifier in the events


Gozala:
Currently parsing following markdown content

```
[link][link-id]

[link-id]:https://my.link/foo "my link"

```
outputs following events:

```
Start(Paragraph)
Start(Link(Reference, Borrowed("https://my.link/foo"), Borrowed("my link")))
Text(Borrowed("link"))
End(Link(Reference, Borrowed("https://my.link/foo"), Borrowed("my link")))
End(Paragraph)
```

In my use case I would like to know the reference link identifier, in this example `link-id` and would like to propose changing output to produce following events instead:

```
Start(Paragraph)
Start(Link(Reference(Borrow("link-id")), Borrowed("https://my.link/foo"), Borrowed("my link")))
Text(Borrowed("link"))
End(Link(Reference(Borrow("link-id")), Borrowed("https://my.link/foo"), Borrowed("my link")))
End(Paragraph)
```

Unfortunately due to #430 emitted ranges also can not be used to capture reference identifiers.

Gozala:
I have started digging into the code base to try and implement proposed change (I am however new to Rust so it's not going all that fast), it would be good to know if what's proposed does make sense and if it would be welcomed.

In the meantime I'll try to post comments as I struggle through in a hope to get some insights.

Gozala:
My current changes are:

```diff
diff --git a/src/parse.rs b/src/parse.rs
index 5ec712b..6fa4044 100644
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -99,41 +99,41 @@ pub enum Tag<'a> {
     Strikethrough,
 
     /// A link. The first field is the link type, the second the destination URL and the third is a title.
-    Link(LinkType, CowStr<'a>, CowStr<'a>),
+    Link(LinkType<'a>, CowStr<'a>, CowStr<'a>),
 
     /// An image. The first field is the link type, the second the destination URL and the third is a title.
-    Image(LinkType, CowStr<'a>, CowStr<'a>),
+    Image(LinkType<'a>, CowStr<'a>, CowStr<'a>),
 }
 
 /// Type specifier for inline links. See [the Tag::Link](enum.Tag.html#variant.Link) for more information.
-#[derive(Clone, Debug, PartialEq, Copy)]
-pub enum LinkType {
+#[derive(Clone, Debug, PartialEq)]
+pub enum LinkType<'a> {
     /// Inline link like `[foo](bar)`
     Inline,
[...]

Issue #423, opened 2020-02-02
Preserve link reference definitions in parser output


dsherret:
Hello,

I'm using pulldown-cmark to create a markdown code formatter. So far it's been working well and thanks for this library!

I've hit a bit of a hurdle though because the parser does not have events for link reference definitions.

For example, given the following text:

```markdown
[testing][Some reference]

[Some reference]: https://github.com

testing
```

The parser output is the following:

```
Event::Start(Paragraph)
Event::Start(Link(Reference, Borrowed("https://github.com"), Borrowed("")))
Event::Text(Borrowed("testing"))
Event::End(Link(Reference, Borrowed("https://github.com"), Borrowed("")))
Event::End(Paragraph)
Event::Start(Paragraph)
Event::Text(Borrowed("testing"))
Event::End(Paragraph)
```

Is it possible to know that the reference definition appeared between the two paragraphs? Worst case scenario, I will just parse this information out of the file myself.

Thanks!

ScottAbbey:
> create a markdown code formatter.

Aha. I was wondering what purpose this request could have. Then I realized you meant this is for markdown-to-markdown, not "formatting code blocks within markdown".

In addition to where the link reference definition was, wouldn't you also want to know *what* the reference name was? Currently "Some reference" does not appear in the event list you showed.  Also, it 

It seems like there could be some other requests alongside this, like:

- Know more info about a list that was encountered...such as which marker or delimiter, and if it is tight or loose.
- There is probably some situation where you would want to know whether or not there was a blank line in there somewhere
- Was this Heading an ATX heading or a setext heading?

So, it seems the request is more along the lines of "Can we have significantly more detail in Event?" Stuffing extra things into Event is probably going to have some sort of performance cost, maybe that is where the discussion would need to start?

--- 

> I will just parse this information out of the file myself.

This would be relatively complex, see `scan_refdef()`...it's not the simplest function out there.

dsherret:
@ScottAbbey I think these details could be omitted by default and only returned when an option is specified... something like `options.insert(Options::INCLUDE_LINK_DEFINITIONS);`.

Alternatively, perhaps they could be provided by request off the `Parser` object (and include their ranges in the document so I can know when to deal with them). That might be a bit better because then it wouldn't mean everyone else has to ignore some link reference event.

> In addition to where the link reference definition was, wouldn't you also want to know what the reference name was? Currently "Some reference" does not appear in the event list you showed.

[...]

PR #436, opened 2020-03-29
add link identifiers into emitted tags

Gozala:
This is implementation for #434. I do not like the fact that `id` is added for both inline and reference links, however different approaches were way more intrusive (and more difficult for newbie to implement). I have proposed multiple options in the issues maybe we can discuss to identify one that would be most appropriate.


marcusklaas:
Sorry about the long radio silence. I acknowledge that your use case is a valid one, but I have my doubts about widespread its use is.

Another way you could achieve this goal without having to modify pulldown itself is to use the links source map to read the tag.

Gozala:
> Sorry about the long radio silence. I acknowledge that your use case is a valid one, but I have my doubts about widespread its use is.
 
I do not know. However reference labels are internally parsed and saved to do resolution so it not really introducing complexity merely just exposing information that is already there.

> Another way you could achieve this goal without having to modify pulldown itself is to use the links source map to read the tag.

Perhaps once #430 is fixed that would be an option.

Gozala:
@marcusklaas I think it is also worth considering issue https://github.com/raphlinus/pulldown-cmark/issues/423 in this context. And specifically:

> I think it is worth considering how currently [`Event::FootnoteReference(CowStr<'a>)`](https://docs.rs/pulldown-cmark/0.7.0/pulldown_cmark/enum.Event.html#variant.FootnoteReference) and [`Tag::FootnoteDefinition(CowStr<'a>)`](https://docs.rs/pulldown-cmark/0.7.0/pulldown_cmark/enum.Tag.html#variant.FootnoteDefinition) are decoupled from each other and how one can be present without the other.
>
> That is pretty much what I believe we need for link / image reference and their reference definitions.

It appears to me that it would make a lot more sense to have link / image reference be similar to footnote references. Instead of resolving those and concealing details like reference identifiers. In such design exposing reference identifiers is natural. It also would imply that second pass that I believe is there to do reference resolutions would be unnecessary (at least for use cases that don't require serializing markdown to html).

alerque:
> but I have my doubts about widespread its use is.

I have a use-case for it.

dsherret:
+1 would be great to have this or provide a way in pulldown-cmark to lazily get this information out of the node when providing a range for a link.

Martin1887:
I think that enum variants get too much complicated to have unnamed fields instead of named fields, so a new pull request replicating the functionality will be implemented if this is not changed.

Thanks.

src/parse.rs
```diff
@@ -98,11 +98,13 @@ pub enum Tag<'a> {
     Strong,
     Strikethrough,
 
-    /// A link. The first field is the link type, the second the destination URL and the third is a title.
-    Link(LinkType, CowStr<'a>, CowStr<'a>),
+    /// A link. The first field is the link type, the second the destination URL and the third is a title,
+    /// the fourth is the link identifier.
+    Link(LinkType, CowStr<'a>, CowStr<'a>, CowStr<'a>),
```
Martin1887, line 103: Too much complicated variant to have 4 unnamed fields.

src/parse.rs
```diff
@@ -98,11 +98,13 @@ pub enum Tag<'a> {
...
 
-    /// An image. The first field is the link type, the second the destination URL and the third is a title.
-    Image(LinkType, CowStr<'a>, CowStr<'a>),
+    /// An image. The first field is the link type, the second the destination URL and the third is a title,
+    /// the fourth is the link identifier.
+    Image(LinkType, CowStr<'a>, CowStr<'a>, CowStr<'a>),
```
Martin1887, line 107: Too much complicated variant to have 4 unnamed fields.

Martin1887 (commented):
See comments in the code, too much complicated enum variants to have unnamed fields.

PR #438, opened 2020-04-06
Break html scan on eof

### LINKED ISSUE #437: Hangs on unterminated html tag
This was initially reported in `rust-clippy` in https://github.com/rust-lang/rust-clippy/issues/4917
that clippy hangs on doc comment `/// <foo`

I can see that same issue occurs directly in pulldown_cmark parser:

```rust
#[test]
fn test_unterminated_foo() {
    use pulldown_cmark::Parser;
    let parser = Parser::new("<foo");
    for _ in parser {}
}
```

raphlinus:
Fixes #437

### DIFF
--- a/src/scanners.rs
+++ b/src/scanners.rs
@@ -1007,6 +1007,9 @@ pub(crate) fn scan_html_block_inner(
             loop {
                 i += scan_whitespace_no_nl(&data[i..]);
                 if let Some(eol_bytes) = scan_eol(&data[i..]) {
+                    if eol_bytes == 0 {
+                        return None;
+                    }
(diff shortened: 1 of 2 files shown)


ehuss:
@raphlinus Would it be possible to get a point release with this fix?

sjackman:
I just ran into this bug. `rustdoc` 1.43.0 is affected by it. It hangs indefinitely on `/// <b`. A point release would be very much appreciated. Thank you @raphlinus for the fix.

raphlinus:
The 0.7.1 release is now out, and includes this fix. Thanks for the interest!

PR #447, opened 2020-06-04
Fix whitespace issue in lists

marcusklaas:
We would previously close lists when a line wasn't completely empty but didn't have the required indentation. Now lines that only contain whitespace are also correctly identified as empty.


tests/suite/regression.rs
```diff
@@ -850,8 +850,44 @@ fn regression_test_61() {
 
 #[test]
 fn regression_test_62() {
-    let original = "<foo";
```
marcusklaas, line 853: Where did this testcase go???

raphlinus: It was added as part of #438 and should be retained, I think.

marcusklaas: Agreed, but it was somehow only removed from the *generated* tests. It looks like it was never added to the test source?

marcusklaas: Ah yes, looking at your PR that seems to be the case. I shall add it to the test source in this PR.

raphlinus: Right, that's what happened, I added it to the wrong file. Thanks!

specs/regression.txt
```diff
@@ -747,3 +747,54 @@ ISSUE 421
...
+```````````````````````````````` example
+* A list.
+
+   * A sublist.
+
+   * Another sublist.
+
+
+* A list.
+
```
marcusklaas, line 762: It's unfortunate that GitHub doesn't show whitespace on blank lines here. It makes this test look a little silly ^^

PR #450, opened 2020-06-11
Remove TreePointer in favor of Option<TreeIndex>

jyn514:
This makes it easier for new contributors to understand the code by
reducing the number of abstractions that need to be learned.


marcusklaas:
This seems like the better design. Thanks for making the changes!

jyn514:
No problem!

Issue #452, opened 2020-06-28
into_offset_iter() - Incorrect range for Table event


dsherret:
Given the following:

```markdown
Testing|This|Outtt
--|:--:|--:
Some Data|Other data|asdf
```

The table range starts at `18`, but I think it should start at `0`:

```
Event: Start(Table([None, Center, Right]))
Range: 18..57

Event: Start(TableHead)
Range: 0..19
```

marcusklaas:
That probably should start at 0 indeed. Thanks for reporting this!

PR #453, opened 2020-07-02
Don't adjust table offset start

marcusklaas:
The update didn't include the table header, it was right at start!

This fixes https://github.com/raphlinus/pulldown-cmark/issues/452.


marcusklaas:
Yes they were, should have put them in separate commit 😬

raphlinus:
No worries, just wanted to make sure I wasn't missing something.

PR #458, opened 2020-07-09
Properly deal with inline html split over multiple leafs

### LINKED ISSUE #399: html tag spanning lines in a blockquote closes unexpectedly
The following markdown:

```
> Note: Though you should not rely on this, all pointers to <abbr
> title="Dynamically Sized Types">DSTs</abbr> are currently twice the size of
> the size of `usize` and have the same alignment.
```

causes the `<abbr>` tag to close at the newline.  That is, it becomes:

```
<blockquote>
<p>Note: Though you should not rely on this, all pointers to <abbr
>title=&quot;Dynamically Sized Types&quot;&gt;DSTs</abbr> are currently twice the size of
the size of <code>usize</code> and have the same alignment.</p>
</blockquote>
```

I would expect the blockquote to be mostly transparent and allow the tag to continue.  commonmark.js dingus seems to handle it correctly.  GitHub handles it too (though it strips out `<abbr>` tags).

marcusklaas:
This is one of the last remaining open bugs that we're aware of. The PR adds bookkeeping to HTML parsing to switch over to keeping a buffer whenever we need to skip some intermediary bytes, e.g. in blockquotes:

```md
> inline <html skip
> ="the-right-angle-at-line-start">
```

It's not very pretty, but it does the job reasonably efficiently I believe.

Fixes #399.


src/parse.rs
```diff
@@ -2543,22 +2550,29 @@ impl<'a> Parser<'a> {
     }
 
     /// Returns the next byte offset on success.
-    fn scan_inline_html(&mut self, bytes: &[u8], ix: usize) -> Option<usize> {
+    fn scan_inline_html(&mut self, bytes: &[u8], ix: usize) -> Option<(Option<Vec<u8>>, usize)> {
```
raphlinus, line 2553: The comment above the function is now stale, and this is a case where it would probably be useful to document the type because it's not obvious.

Also, and I'm typing as I go, the `Option<Vec>` feels like it might not be necessary. Is `Some(vec![])` a valid response, important to be distinguished from `None`? If not, maybe just use `is_empty` for its consumer?

src/scanners.rs
```diff
@@ -855,32 +855,71 @@ fn scan_attribute_name(data: &[u8]) -> Option<usize> {
...
+    Some(i)
+}
+
+/// Returns the index immediately following the attribute value on success.
 fn scan_attribute_value(
     data: &[u8],
+    mut i: usize,
     newline_handler: Option<&dyn Fn(&[u8]) -> usize>,
+    buffer: &mut Vec<u8>,
+    buffer_ix: &mut usize,
```
raphlinus, line 920: The name `buffer_ix` is slightly confusing to me, because at first read it sounds like an index *within* buffer. On reading more deeply, it also includes skipped bytes. Maybe coming up with a better name is not easy, but a one line comment explaining it in the docstring would be helpful.

src/parse.rs
marcusklaas: Good insights. It turns out that `Option<Vec<u8>>` is not necessary indeed.

marcusklaas:
I've addressed your comments. Thank you for the review!

Issue #459, opened 2020-07-22
Footnotes parsed without Options::ENABLE_FOOTNOTES


lunaryorn:
A parser created with `Parser::new_ext` *without* including `Options::ENABLE_FOOTNOTES` still emits a `FootnoteReference` event when seeing a footnote.  

I can provide a minimal example but I'd first like to clarify whether this is actually unexpected behaviour.

marcusklaas:
This may be a bug indeed. Investigating...

PR #460, opened 2020-07-22
Don't emit footnote references when option is not set

marcusklaas:
We would correctly skip parsing footnote definitions when the option was not set, but we'd still emit references.

Now, footnote references (of the form [^ref]) are treated as regular links unless footnotes are enabled in the options.

Fixes https://github.com/raphlinus/pulldown-cmark/issues/459.


lunaryorn:
@marcusklaas This branch fixes my issue, thanks :pray:

Issue #373, opened 2019-07-13
Pass link type to broken link callback


euclio:
It would be helpful to know what kind of link failed in the broken link callback to improve diagnostics in rustdoc for intra-doc links.

marcusklaas:
We could add this, as well as https://github.com/raphlinus/pulldown-cmark/issues/165 fairly easily, but it'd be a breaking change. We're about to do a major release, so there is some room for such changes. However, I don't know what impact it'd have on other projects using broken link callbacks. Rustdoc is the big one and possibly the only one, but it'd be good to do a quick survey.

Thoughts, @raphlinus?

raphlinus:
Seems like a good idea to me. If it were me, I'd ask for a PR, or at least a suggested signature, but maybe Marcus has other preferences.

marcusklaas:
Those work for me for sure :+1:

euclio:
I think the following would suffice as a signature:

```rust
fn callback(broken_reference: &'a str, title: &'a str, link_type: LinkType) -> Option<(String, String)>
```

This is assuming that it's safe to assume that `broken_reference` is borrowed from the original markdown (#165).

Michael-F-Bryan:
This should also include *where* in the text a broken link is found. That way tools like link validators can give the user better error messages (including line numbers, content around the broken link, etc.).

marcusklaas:
Done!

Issue #165, opened 2018-11-19
Find offset of link in broken link callback


euclio:
I'd like to be able to distinguish the location of two broken links with the same text.

In rustdoc, we attempt to do this with a callback not unlike the test case below, where we compare the offsets of the pointer of the non-normalized reference and the markdown pointer itself:

```rust
#[test]
fn test_duplicate_broken_links() {
    use ::{Parser, Options};

    let links = RefCell::new(vec![]);

    let md = "[link] [link]";

    let locate = |s: &str| unsafe {
        let s_start = s.as_ptr();
        let s_end = s_start.add(s.len());
        let md_start = md.as_ptr();
        let start = s_start as usize - md_start as usize;
        let end = s_end as usize - md_start as usize;
        start..end
    };

    {
        let callback = |_: &str, s: &str| {
            links.borrow_mut().push((s.to_owned(), locate(s)));
            None
        };

        let p = Parser::new_with_broken_link_callback(md, Options::empty(), Some(&callback));
        for _ in p {}
    }

    assert_eq!(links.into_inner(), vec![
        (String::from("link"), 1..5),
        (String::from("link"), 8..12),
    ]);
}
```

In addition to being unsafe, it doesn't appear to actually work!

```
---- parse::tests::test_duplicate_broken_links stdout ----
thread 'parse::tests::test_duplicate_broken_links' panicked at 'assertion failed: `(left == right)`
  left: `[("link", 8..12), ("link", 8..12)]`,
 right: `[("link", 1..5), ("link", 8..12)]`', src/parse.rs:1829:9
```

While it would be nice to fix this bug, I think it would less error prone if the callback was passed the location of the current broken link, perhaps as a byte range?

marcusklaas:
Done!

PR #469, opened 2020-08-23
A new broken link callback design

marcusklaas:
This is a ~~WIP~~ PR to address several outstanding issues with the current broken link callback design.

Firstly, it provides additional data (source mapping, link type) to the callback to improve diagnostics (https://github.com/raphlinus/pulldown-cmark/issues/373) and help disambiguate links with identical references (https://github.com/raphlinus/pulldown-cmark/issues/165). Further, this design also prevents the callback from being called twice on the same reference (https://github.com/raphlinus/pulldown-cmark/issues/444). And lastly, the callback now returns `CowStr`s, so that it is possible to generate titles and urls without memory allocations, for example when they are static strings or derived from text in the source.

Feedback is greatly appreciated. Would this cover your use-cases? Is this an improvement over the old design?

cc @euclio @jyn514 @GuillaumeGomez

### DIFF
--- a/examples/broken-link-callbacks.rs
+++ b/examples/broken-link-callbacks.rs
@@ -9,12 +9,12 @@ fn main() {
     // Setup callback that sets the URL and title when it encounters
     // a reference to our home page.
     let callback = &mut |broken_link: BrokenLink| {
-        if broken_link.reference.as_ref() == "my website" {
+        if broken_link.reference == "my website" {
             println!(
                 "Replacing the markdown `{}` of type {:?} with a working link",
                 &input[broken_link.span], broken_link.link_type,
             );
-            Some(("http;//example.com".into(), "my example website".into()))
+            Some(("http://example.com".into(), "my example website".into()))
         } else {
             None
         }
@@ -29,7 +29,7 @@ fn main() {
 
     // Check that the output is what we expected.
     let expected_html: &str =
-        "<p>Hello world, check out <a href=\"http;//example.com\" title=\"my example website\">my website</a>.</p>\n";
+        "<p>Hello world, check out <a href=\"http://example.com\" title=\"my example website\">my website</a>.</p>\n";
     assert_eq!(expected_html, &html_output);
 
     // Write result to stdout.
--- a/examples/broken-link-callbacks.rs
+++ b/examples/broken-link-callbacks.rs
@@ -1,0 +1,37 @@
+extern crate pulldown_cmark;
+
+use pulldown_cmark::{html, BrokenLink, Options, Parser};
+
+fn main() {
+    let input: &str = "Hello world, check out [my website][].";
+    println!("Parsing the following markdown string:\n{}", input);
+
+    // Setup callback that sets the URL and title when it encounters
+    // a reference to our home page.
+    let callback = &mut |broken_link: BrokenLink| {
+        if broken_link.reference.as_ref() == "my website" {
+            println!(
+                "Replacing the markdown `{}` of type {:?} with a working link",
+                &input[broken_link.span], broken_link.link_type,
+            );
+            Some(("http;//example.com".into(), "my example website".into()))
+        } else {
+            None
+        }
+    };
+
+    // Create a parser with our callback function for broken links.
+    let parser = Parser::new_with_broken_link_callback(input, Options::empty(), Some(callback));
+
+    // Write to String buffer.
+    let mut html_output: String = String::with_capacity(input.len() * 3 / 2);
+    html::push_html(&mut html_output, parser);
+
+    // Check that the output is what we expected.
+    let expected_html: &str =
+        "<p>Hello world, check out <a href=\"http;//example.com\" title=\"my example website\">my website</a>.</p>\n";
+    assert_eq!(expected_html, &html_output);
+
+    // Write result to stdout.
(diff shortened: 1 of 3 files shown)


src/parse.rs
```diff
@@ -277,6 +278,12 @@ enum TableParseMode {
     Disabled,
 }
 
+pub struct BrokenLink<'a> {
+    pub range: ::std::ops::Range<usize>,
```
jyn514, line 282: nit: I would call this the 'span' of the link, range makes me think of number

src/parse.rs
```diff
@@ -1922,13 +1928,20 @@ pub(crate) struct HtmlScanGuard {
     pub declaration: usize,
 }
 
+pub struct BrokenLinkFix<'a> {
+    title: CowStr<'a>,
+    url: CowStr<'a>,
+}
+
+pub type BrokenLinkCallback<'a> = Option<&'a dyn Fn(BrokenLink) -> Option<BrokenLinkFix<'a>>>;
```
jyn514, line 1936: While you're at it, can you change this to `FnMut` so it doesn't have to use `Cell` and `RefCell` for modifying state? I ran into trouble with this while writing the test case.
```suggestion
pub type BrokenLinkCallback<'a> = Option<&'a mut dyn FnMut(BrokenLink) -> Option<BrokenLinkFix<'a>>>;
```

marcusklaas: Yes, can do!

src/parse.rs
marcusklaas: Good one, thanks.

euclio:
This looks great to me!

src/parse.rs
```diff
@@ -2206,12 +2240,20 @@ impl<'a> Parser<'a> {
                                     .or_else(|| {
                                         self.broken_link_callback
                                             .and_then(|callback| {
+                                                // Construct brokenlink struct, which is what the callback
+                                                // will take
```
jyn514, line 2244:
```suggestion
                                                // Construct a BrokenLink struct, which will be passed to the callback
```

examples/broken-link-callbacks.rs
```diff
@@ -0,0 +1,37 @@
...
+
+    // Setup callback that sets the URL and title when it encounters
+    // a reference to our home page.
+    let callback = &mut |broken_link: BrokenLink| {
+        if broken_link.reference.as_ref() == "my website" {
+            println!(
+                "Replacing the markdown `{}` of type {:?} with a working link",
+                &input[broken_link.span], broken_link.link_type,
+            );
+            Some(("http;//example.com".into(), "my example website".into()))
```
jyn514, line 17:
```suggestion
            Some(("http://example.com".into(), "my example website".into()))
```

examples/broken-link-callbacks.rs
```diff
@@ -0,0 +1,37 @@
...
+    // Create a parser with our callback function for broken links.
+    let parser = Parser::new_with_broken_link_callback(input, Options::empty(), Some(callback));
+
+    // Write to String buffer.
+    let mut html_output: String = String::with_capacity(input.len() * 3 / 2);
+    html::push_html(&mut html_output, parser);
+
+    // Check that the output is what we expected.
+    let expected_html: &str =
+        "<p>Hello world, check out <a href=\"http;//example.com\" title=\"my example website\">my website</a>.</p>\n";
```
jyn514, line 32:
```suggestion
        "<p>Hello world, check out <a href=\"http://example.com\" title=\"my example website\">my website</a>.</p>\n";
```

examples/broken-link-callbacks.rs
raphlinus: This seems like it should go in, I think.

src/parse.rs
```diff
@@ -277,6 +278,12 @@ enum TableParseMode {
     Disabled,
 }
 
+pub struct BrokenLink<'a> {
+    pub span: ::std::ops::Range<usize>,
```
raphlinus, line 282: I don't think you need the `::` now (as of 2018 edition).

src/parse.rs
```diff
@@ -277,6 +278,12 @@ enum TableParseMode {
...
+    pub link_type: LinkType,
+    pub reference: &'a CowStr<'a>,
```
raphlinus, line 284: This seems really strange - it's a reference to a reference. I think `&'a str` would do as well, and that can point to the contents of a CowStr. But maybe I'm missing something here.

marcusklaas: You're not missing anything. In fact, you've caught something ;-) Thanks.

src/parse.rs
marcusklaas: TIL!

examples/broken-link-callbacks.rs
marcusklaas: Done.

marcusklaas:
Sending a big thank to @jyn514, @euclio and @raphlinus for the feedback!

Michael-F-Bryan:
This is awesome, thank you @marcusklaas! 

Now I just need to update [`mdbook-linkcheck`](https://github.com/Michael-F-Bryan/mdbook-linkcheck) so we can take advantage of it and remove the previous ~~hack~~ workaround.

PR #474, opened 2020-09-01
Bump version to 0.8.0; update dependencies

marcusklaas:
With this version bump, we can release 0.8 and bring the latest round of fixes and features to the users.


marcusklaas:
:partying_face:

Issue #402, opened 2019-09-13
Split parse module into first and second pass


marcusklaas:
The parse module has gotten fairly large at 3k lines of code. It contains both the first and second pass of parsing, which don't have much code in common. Creating separate modules for the passes and moving shared data structures to the root module may improve maintainability and more clearly signal to new contributors the structure of the parser.

PR #476, opened 2020-09-07
Move code relating to the first pass to its module

marcusklaas:
This splits the massive parse module in two, making a clear distinction what code relates to the first pass exclusively, the second or both. The new `firstpass` module only exports a single function called `run_first_pass`. It also makes the simd module a submodule of this new firstpass module, to make it explicit what the SIMD code is used for.

It also moves a number of key (public) types into the crate root.

This PR is the start of my effort to improve the structure and maintainability of the repository.


marcusklaas:
Closes https://github.com/raphlinus/pulldown-cmark/issues/402.

raphlinus (approved):
Ok, this seems reasonable. I definitely agree that moving the first pass into its own module is a win.

The number of lines is huge, but on a scan it seemed like basically no logic got changed, just the coarse structure of where things live. I'll take that on faith. (One downside to these kinds of PR's is that it makes it harder to sort through git blame because there's no automated way to distinguish between a simple restructure and a rewrite from scratch, but that's a price to pay for better maintainability).

marcusklaas:
No logic was changed in the making of this PR :-)

Issue #6, opened 2015-06-16
math support


blaenk:
This simply means enabling the use of `$` and `$$` delimiters (inline and block-level respectively, as in Pandoc) to very conveniently leverage something like [MathJax](https://www.mathjax.org/). Hoedown also supports this.

raphlinus:
It should be relatively straightforward, but nothing is ever simple. Take `*a $b*c$` for example. At the very least the scanner for emphasis has to be aware of the math spans to properly skip over them (the rule is likely quite similar to that for `` ` ``.

Not on the critical path for rustdoc, but worth doing.

vyp:
Is this in commonmark?

raphlinus:
@vyp No, however it is one of the more commonly requested extensions. See <http://talk.commonmark.org/t/mathematics-extension/457> for more discussion.

vyp:
Oh sorry, I just saw #4. Intention was that commonmark may have guidelines on cases like `*a $b*c$`.

vyp:
@raphlinus Ah yes, I'm familiar with that thread. Second commenter there.

raphlinus:
I'm very open to doing this. By far the most helpful thing from the community would be a precise spec. One got suggested in a [reddit thread](https://www.reddit.com/r/rust/comments/5ljxyo/whats_everyone_working_on_this_week_12017/dbzr59k/). But are these the right rules? Would the rules for `_` just work? I'm not sure that even `$1234` presents a serious problem, because it would need a closing `$` to become a false positive. That thread also links this [Survey of syntaxes for math in markdown](https://github.com/cben/mathdown/wiki/math-in-markdown), which is an incredibly useful resource for figuring this out.

jeanm:
I've also made [a table](https://gist.github.com/jeanm/38af25711e06d1546d93afae6f5ca144#) to summarise some of the syntaxes. I've only focused on the platforms that are likely to be used by mathsy people, since they're the ones likely to use this feature anyway. Feel free to add a comment there if there's anything relevant I've missed! 

> I'm not sure that even `$1234` presents a serious problem, because it would need a closing `$` to become a false positive. 

Not by itself, but there might be more than one dollar sign in a sentence (say if you're making a list of prices)

azerupi:
>Not by itself, but there might be more than one dollar sign in a sentence (say if you're making a list of prices)

That wouldn't be a problem I think. A space before an ending `$` would not be allowed. So 
```
[...]

PR #477, opened 2020-09-09
Add MathJax Support

litao91:
For block math:

```
$$
.... math content
$$
```

For inline math:

```
$items$
```

It handles inline math in a way the same as pandoc:

> Anything between two $ characters will be treated as TeX math. The opening $ must have a character immediately to its right,
> while the closing $ must have a character immediately to its left. Thus, $20,000 and $30,000 won’t parse as math.
> If for some reason you need to enclose text in literal $ characters,
> backslash-escape them and they won’t be treated as math delimiters.


marcusklaas:
Hello and thank you for your pull request. Support for math sequences is one of the longest standing and most highly requested features. There have been [long discussions](https://github.com/raphlinus/pulldown-cmark/issues/6) on what the delimiters for inline and block math sequences should be, but we have unfortunately not yet come to a decision. The latex-style you've implemented is certainly a contender, but there are some drawbacks relating to unintended sequences and graceful fallback.

Before we can move forward with any implementation, we'd have to agree on a format (read: delimiters) and a spec. That said, I did have a quick look at the code and the implementation generally looks good. If we do choose to go with dollars, this PR should make an excellent basis.

You have prompted me to try and form a proposal so we can hopefully finally get this sorted. I will include you in future discussion. Thanks again!

litao91:
Hi @marcusklaas, actually I've noticed the discussions. I desperately need the math support, so I rushed to implement a basic working version for my personal needs. It's true that I haven't consider too much about the edges cases and this one shouldn't be a complete and sound solution. 

I am a heavy markdown user and this is certainly one of the best implementation. It would be perfect if it has any kind of math support. 

Thanks

ollpu:
Any content on a line that starts with `$$` is currently ignored, and something like this, which I think is relatively common:

```
$$ \LaTeX $$
```

breaks the rest of the document.

Just mentioning to keep track. Hopefully we can get this merged soon!

tgross35:
Hey @marcusklaas did you ever put that proposal together? Or what is the status of planning for this?

#6 is a mess of bikeshedding over the delimiter. `$...$` and `$$...$$` is basically ubiquitous at this point (used by StackExchange, Pandoc, Jupyter, and now GitHub $E=mc^2$) so should definitely be included. Choosing other delimiters shouldn't block implementation work since they could be added later.

Assuming that logic is satisfactory, what is blocking this from being merged? There hasn't been too much of a discussion on KaTeX vs MathJax and pre- vs. post- rendering, but the work here is already mostly done, and could be adapted later if desired.

@litao91 are you interested in doing work to make this mergable, if needed? I see you are at least keeping this branch up to date with upstream

Martin1887:
Closed in favor of #734.

PR #482, opened 2020-09-23
Deny unsafe code unless the SIMD feature is enabled

marcusklaas:


raphlinus (approved):
Yes, I like that this communicates the intent simply.

PR #483, opened 2020-09-23
Add command line option for smart punctuation

marcusklaas:


raphlinus (approved):
Yes, this makes me think, "what an oversight, I'm glad Markus is on it." -- So thanks!

alerque (approved):

Issue #487, opened 2020-09-27
bug: \t-\tp broken


generateui:
Hi! I came across this bug on BabelMark, as I'm writing a markdown CST for C#. 

this lib writes an `<ul>` when having the input text `\t-\tp` (`\t` being a tab and `p` being paragraph content.

Version: 0.8

generateui:
https://babelmark.github.io/?text=%09-%09It+offers

marcusklaas:
Yup, that looks like a bug. Thanks for reporting it!

PR #488, opened 2020-09-28
Fix bug with tab indented codeblocks being interpreted as lists

marcusklaas:
This was fairly subtle. We would scan up to 3 spaces, consuming a tab and leaving one space in the buffer. But it would also already bump the internal text pointer forward past the tab, so if we'd find a list marker there, it would incorrectly be interpreted as the list because we didn't look at the space buffer.

Fixes https://github.com/raphlinus/pulldown-cmark/issues/487.

### DIFF
--- a/specs/regression.txt
+++ b/specs/regression.txt
@@ -814,3 +814,11 @@ lo"></p>
 </blockquote>
 </blockquote>
 ````````````````````````````````
+
+ISSUE 487
+
+```````````````````````````````` example
+	-	the whitespace here are tabs
+.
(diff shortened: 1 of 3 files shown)


raphlinus (approved):
This stuff is subtle. I haven't worked through all the implications of the logic but trust you, especially given the test suite. (If you're feeling less certain and want to ask for a more in-depth review, I can also do that)

marcusklaas:
I took conservative approach to this fix, and checked for any other places where this could have been happening, but this seems to be the only one. Thank for your confidence :-)

PR #493, opened 2020-10-09
Use `std::fmt::Write` instead of custom `StrWrite` trait

### LINKED ISSUE #492: Use `core::fmt::Write` instead of custom `StrWrite`?
Is there a particular reason that pulldown-cmark defines its own type, [`StrWrite`](https://docs.rs/pulldown-cmark/0.8.0/pulldown_cmark/escape/trait.StrWrite.html), rather than using the built-in Rust trait [`core::fmt::Write`](https://doc.rust-lang.org/nightly/core/fmt/trait.Write.html)? Previously `core::fmt::Write`'s documentation said

> This trait should generally not be implemented by consumers of the standard library.

But in a [recent PR](https://github.com/rust-lang/rust/pull/76750), I changed it to remove that section since no one knew of a reason why it said that. Perhaps it would be good to move pulldown over to that built-in trait?

camelid:
Closes #492.

### DIFF
--- a/src/escape.rs
+++ b/src/escape.rs
@@ -21,8 +21,8 @@
 //! Utility functions for HTML escaping. Only useful when building your own
 //! HTML renderer.
 
-use std::fmt::{Arguments, Write as FmtWrite};
-use std::io::{self, ErrorKind, Write};
+use std::fmt::{self, Arguments};
+use std::io;
 use std::str::from_utf8;
 
 #[rustfmt::skip]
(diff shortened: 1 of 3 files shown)


src/escape.rs
```diff
@@ -42,67 +42,36 @@ static AMP_ESCAPE: &str = "&amp;";
...
 where
-    W: StrWrite,
+    W: io::Write,
 {
     #[inline]
-    fn write_str(&mut self, s: &str) -> io::Result<()> {
-        (**self).write_str(s)
+    fn write_str(&mut self, s: &str) -> fmt::Result {
+        match self.0.write_all(s.as_bytes()) {
+            Ok(()) => Ok(()),
```
marcusklaas, line 57: We could `.map_err` here.

src/escape.rs
```diff
@@ -42,67 +42,36 @@ static AMP_ESCAPE: &str = "&amp;";
...
+            Err(_) => Err(fmt::Error),
+        }
     }
 
     #[inline]
-    fn write_fmt(&mut self, args: Arguments) -> io::Result<()> {
-        (**self).write_fmt(args)
+    fn write_fmt(&mut self, args: Arguments) -> fmt::Result {
+        match self.0.write_fmt(args) {
+            Ok(()) => Ok(()),
```
marcusklaas, line 65: Ditto.

marcusklaas (commented):
The implementation looks good, but I just now noticed that `fmt::Write` has its own error. This means that it's no longer possible to pass along IO errors. If a write fails to a `Write` object, the user will have no way to discover why. This seems like a serious drawback of the change.

camelid:
> The implementation looks good, but I just now noticed that `fmt::Write` has its own error. This means that it's no longer possible to pass along IO errors. If a write fails to a `Write` object, the user will have no way to discover why. This seems like a serious drawback of the change.

Yeah, I agree :/

Maybe not worth doing for now.

camelid:
Not sure if you're on the Rust Zulip, but I posted this: https://rust-lang.zulipchat.com/#narrow/stream/219381-t-libs/topic/fmt.3A.3AResult.20doesn't.20allow.20having.20an.20error.20message

edward-shen:
From that topic: 

> simulacrum: It is perhaps worth noting that io::Write::write_fmt does preserve the io Error in the raw form (https://doc.rust-lang.org/nightly/src/std/io/mod.rs.html#1498-1513)
> simulacrum: so if you're using write! or similar with io sources, and not the fmt trait directly, you'll get the error

From my understanding, our use case falls under the former category, so is this still considered a blocker?

leeola:
I wonder if this PR could be extended to allow a more generic representation of `push_html(&mut String,..)`? I'm running into a related problem where a template library (Sailfish) has a `Buffer` type that impls `fmt::Write`, but not `io::Write` for similar reasons *(i assume)* that this library originally wrote `StrWrite`.

As far as i can tell this requires users of Sailfish to allocate a secondary `String` to receive html from this lib. Making the `push_html` generic seems a good way to deal with this, and `fmt::Write` seems a possible candidate to support that with this PR. Thoughts?

Martin1887:
This is an old pull request and in draft state, it should be redesigned and thought again. Sorry, but this will not happen in the short term unless someone else begin to work in this.

PR #495, opened 2020-10-13
Add convenience empty InlineStr ctor

lnicola:
This helps avoid writing `use std::convert::TryFrom; InlineStr::try_from("").unwrap()`, as seen in https://github.com/rust-analyzer/rust-analyzer/pull/6217/files#diff-4e6b4f7c9db0646b27efab3b7c999593ba2480b3afe3312889ff471251bc870bR72.

### DIFF
--- a/src/strings.rs
+++ b/src/strings.rs
@@ -19,6 +19,15 @@ pub struct InlineStr {
     inner: [u8; MAX_INLINE_STR_LEN],
 }
 
+impl InlineStr {
+    // Returns an empty `InlineStr`.
+    pub fn empty() -> Self {
+        Self {
+            inner: [0; MAX_INLINE_STR_LEN],
+        }
+    }
+}
+
 impl<'a> AsRef<str> for InlineStr {
     fn as_ref(&self) -> &str {
         self.deref()


src/strings.rs
```diff
@@ -19,6 +19,15 @@ pub struct InlineStr {
     inner: [u8; MAX_INLINE_STR_LEN],
 }
 
+impl InlineStr {
+    // Returns an empty `InlineStr`.
+    pub fn new() -> Self {
```
oberien, line 24: What do you think about renaming it to `pub fn empty()`? I'd expect `SomeStr::new` to take an argument which is converted to that `SomeStr`. /bikeshedding

lnicola: I think I prefer that, but picked `new` to match `String`. I can rename it, of course.

marcusklaas:
I'm not against this, but I think that `"".into()` would work as well there. That'd create a borrowed `CowStr` instead of an inlined one, but it won't allocate either.

lnicola:
Yes, it does seem to work (with a `'static` lifetime). I didn't try that because in the first place in that PR, trying to use the borrowed variant yielded some nasty lifetime errors.

marcusklaas:
Sorry for leaving this open for so long. Since it seems this functionality is already covered by the current code, I'll close the PR.

lnicola:
Woah, 2021 flew by so quickly :rofl:.

PR #497, opened 2020-10-17
Relax broken link callback requirements

vojtechkral:
Hi,
    it's nice to see the broken link callback's got some refactoring.
However, I find the new design hard to use.
In particular, it seems it's not possible with the current design for the parser
iterator to escape the function scope in which the callback lives
(unless I use `unsafe`).

In the PR I propose a change of the broken link requirements to make this possible.

You can see a simplified version of my use-case in the added `broken_link_callback_move` unit test.

The important bit is that the `BrokenLink` type carries the lifetime of the input
rather than some random internal stack lifetime, which makes it possible to
use a HRTB to store the closure in a box.

I also relaxed `FnMut` to just `Fn`, it seems to me to make the usage easier,
but if you don't like it I don't insist on this part.

Let me know if this makes sense (or maybe whether I missed something).
Thanks! :slightly_smiling_face:

### DIFF
--- a/examples/broken-link-callbacks.rs
+++ b/examples/broken-link-callbacks.rs
@@ -9,10 +9,10 @@ fn main() {
     // Setup callback that sets the URL and title when it encounters
     // a reference to our home page.
     let callback = &mut |broken_link: BrokenLink| {
-        if broken_link.reference == "my website" {
+        if broken_link.reference.as_ref() == "my website" {
             println!(
                 "Replacing the markdown `{}` of type {:?} with a working link",
-                &input[broken_link.span], broken_link.link_type,
+                broken_link.source(), broken_link.link_type,
             );
             Some(("http://example.com".into(), "my example website".into()))
         } else {
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -121,10 +121,19 @@ impl<'a> Default for ItemBody {
     }
 }
 
+#[derive(Debug)]
 pub struct BrokenLink<'a> {
-    pub span: std::ops::Range<usize>,
+    pub span: Range<usize>,
     pub link_type: LinkType,
-    pub reference: &'a str,
+    pub reference: CowStr<'a>,
+    pub input: &'a str,
+}
+
+impl<'a> BrokenLink<'a> {
+    /// The source text for this broken link
+    pub fn source(&self) -> &'a str {
+        &self.input[self.span.clone()]
+    }
 }
 
 /// Markdown event iterator.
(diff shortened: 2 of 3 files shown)


marcusklaas:
Hi, and thank you for your PR. Can you expand a bit on this bit, and perhaps provide a bit of code which is currently blocking you?

> In particular, it seems it's not possible with the current design for the parser iterator to escape the function scope in which the callback lives

Concerning the `Fn` vs `FnMut`: changing the function trait to the latter was a common request before we updated the callback design. Unless a glaring oversight or major issue pops up, I am very reluctant to reverse that decision.

vojtechkral:
> Hi, and thank you for your PR. Can you expand a bit on this bit, and perhaps provide a bit of code which is currently blocking you?

I am working on a [this project](https://github.com/vojtechkral/bard/tree/doc) which is currently nearing a first release. Basically, I'm defining a custom markdown subset on top of the pulldown parser. The relevant code is [here](https://github.com/vojtechkral/bard/blob/b5ca5e4caa3b012c2cea8ab72aa84ad99a68a204/src/parser.rs#L551) the `parse()` function returns an iterator of events (`Event`) representing my md subset.

With the current design I can't have the callback boxed in the `MDFile` type. It could be done with unsafe code.
Conceivably it could be done if the `MDFile` didn't own the data, only reference them (I didn't yet try this), although I'd have to re-design the thing somwhat...

> Concerning the `Fn` vs `FnMut`: changing the function trait to the latter was a common request before we updated the callback design. Unless a glaring oversight or major issue pops up, I am very reluctant to reverse that decision.

Oh, ok, I'll remove that part...

vojtechkral:
>  Conceivably it could be done if the `MDFile` didn't own the data, only reference them (I didn't yet try this), although I'd have to re-design the thing somwhat...

Tested, and no, I don't think it can be done. I still get lifetime errors even if the `MDFile` struct only references the input.
I tried with `pub fn parse<'s>(&'s mut self) -> Events<'s> where 's: 'a { ... }` and I get errors as soon as I try to obtain the iterator.

```
   --> src/book.rs:488:26
    |
488 |         let mut events = parser.parse();
    |                          ^^^^^^ borrowed value does not live long enough
...
540 |     }
    |     -
    |     |
    |     `parser` dropped here while still borrowed
    |     borrow might be used here, when `parser` is dropped and runs the destructor for type `parser::Parser<'_>`
```

Btw. this is in no way specific to my custom `Events` iterator, exactly the same thing happens with `OffsetIter` as well.
I think this can only be done with either the HRTB as in the PR or by extracting the link callback out of the parser, but honestly I don't like that very much as that means extracting the parser implementation details outside of it.

I'll revert the `FnMut` → `Fn` change shortly...

vojtechkral:
Hm. I think #510 probably fixes this in a better way (and I have since stopped using this library actually). Closing...

PR #499, opened 2020-10-25
Export BrokenLinkCallback from the 'parse' module

### LINKED ISSUE #498: Export the `BrokenLinkCallback` type
Hi,

This is just a little issue to ask if you could export the `BrokenLinkCallback` type, as it's currently public in the `parse` module, which is not public itself. This would be simplier to pre-declare broken link callbacks in code that uses this crate.

Thanks in advance and thanks for this amazing library :) !

ClementNerma:
Fixes #498

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -68,7 +68,7 @@ mod scanners;
 mod strings;
 mod tree;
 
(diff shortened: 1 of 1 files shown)


marcusklaas:
Thank you!

PR #502, opened 2020-11-11
Fixup fuzzer

edward-shen:
(Sorry for not getting in contact first, I saw the contributing guidelines while making this PR).

I was looking around your issues and I saw that there was some low hanging fruit for the fuzzer. When I tried to work on it, it looked like the fuzzer framework was a little broken.

This PR implements multiple changes that get the fuzzer into a working state, of which some fixes some problems that didn't have an issue attached:
 - Fixed where the fuzzer fails to build as of [1.45 due to an incompatibility with `rustc`'s `lto=fat` and `embed-bitcode=no`](https://github.com/rust-lang/rust/blob/master/RELEASES.md#compatibility-notes-2).
 - Fixed regex matching in `group-panics` since the double quotes are no longer printed.
 - Fixed setting the niceness value. The previous implementation set the nice level to 19 (most nice) rather than -20 (least nice), so I would suspect a lot of inconsistencies with non-linear performance detection. Hopefully this should reduce number of false positives a bit.
 - Adds struct-opt parsing via clap 3.0 -- the codebases have merged in 3.0.
 - Minor bash script improvements. These were changes that whose implementation were incorrect or ambiguous, and were detected through shellcheck.


marcusklaas:
Thanks so much Edward! The fuzzer code has been neglected, and some love was much needed. This looks very good. I will have a quick spin of the code on my machine to see if any issues pop up. If not, we'll merge this as soon as possible.

fuzzer/group-panics
```diff
@@ -29,7 +29,7 @@ def test(pattern, release=True):
     return stderr
 
 for panic in panics:
-    pattern = re.match(br'Panic caught\. Pattern: "(.*)"', panic).group(1)
+    pattern = re.match(br'Panic caught\. Pattern: (.*)', panic).group(1)
```
marcusklaas, line 32: This isn't because your change, but I'm getting this error:
```
Traceback (most recent call last):
  File "./group-panics", line 32, in <module>
    pattern = re.match(br'Panic caught\. Pattern: (.*)', panic).group(1)
AttributeError: 'NoneType' object has no attribute 'group'
```
Does `.match(...)` return `None` when no match is found? Should that be possible and if yes, how should we handle it?

marcusklaas:
I've done a few test runs on my machine, and it seems to work well again. I'm still getting false positives, even after retesting, but I'm convinced this isn't because of your changes or timing issues, but rather some unexplored subtleties.

If you happen to know a quick fix for the python error flagged above, that'd be great. If not, that's no problem and we can fix it at a later time. Thanks again so much for this awesome PR! :+1:

fuzzer/group-panics
edward-shen: Yeah, it does when the pattern match fails. This occurs when the output file is non-empty but does not contain any crashes.

I've added a null check to skip the line if it doesn't match on anything, which should cover this problem. Honestly, this would be indicative that a more in-depth refactor might be worth considering, but considering how this PR is suppose to just get it back into a working state, I've decided against a larger change like that.

edward-shen:
Yeah, from the readme it sounds like there were originally a bunch of false positives to begin with; I'm only expecting my changes to _maybe_ reduce the ratio, not eliminate it.

edward-shen:
Also not sure what's going on with the CI failure. Looks like an infrastructure error (`error: failed to parse lock file at: /home/vsts/work/1/s/fuzzer/Cargo.lock`) as both you and I were able to build it on our machines.

marcusklaas:
Yup, that is weird.. Only happens for the fuzzer for some reason; all the other tests pass. If it persists for other PRs, I'll have another look.

Thanks again!

edward-shen:
Glad to be of help!

marcusklaas:
Thanks for following up on this. This seems like a valid reason for a minor MSRV bump since the next time we update the main cargo lock file this will bite us again. I will make an issue for this.

Issue #503, opened 2020-11-11
Bump MSRV to 1.38


marcusklaas:
This means we can move to V2 cargo lock files, which are now the default.

PR #505, opened 2020-11-22
Bump MSRV to 1.42

marcusklaas:
This allows us to use the cargo lock file V2 and the `matches!` macro,
avoiding issues with forward compatibility and clippy warnings.

Version 1.42 is a jump up from 1.36, but it is still 6 versions behind
the most recent stable version of Rust. This allows for plenty time
for pulldown users to upgrade.

Closes https://github.com/raphlinus/pulldown-cmark/issues/503.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2,923 +2,927 @@
 # It is not intended for manual editing.
 [[package]]
 name = "aho-corasick"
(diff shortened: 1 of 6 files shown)


raphlinus (approved):
Seems good to me!

Issue #269, opened 2019-04-15
update benchmark suite


marcusklaas:
We've mostly been focused on xi-editor's crdt.md, which is a good general specimen, but it's probably wise to start including other snippets as to not hyper-optimize for one specific document :sweat_smile: 

Also - should we consider [Criterion.rs](https://bheisler.github.io/criterion.rs/book/)? It looks *very* powerful.

Some starting points for markdown benchmark document searches:
- https://talk.commonmark.org/t/corpus-of-markdown-documents-for-benchmarking/2310
- https://github.com/commonmark/commonmark.js/tree/master/bench/samples
- https://github.com/commonmark/commonmark.js#performance

PR #506, opened 2020-12-01
Miscellaneous code quality improvements

edward-shen:
This set of patches incorporates multiple smaller improvements whose overall theme is "improved code quality". Each commit should be atomic, so I encourage looking at each commit individually, rather than the patch as a whole.

#### Changelog:

 - Implement [`FusedIterator`](https://doc.rust-lang.org/std/iter/trait.FusedIterator.html) marker trait for `Parser`. This is a non-breaking optimization that allows `.fuse()` methods on the iterator on the parser to be a no-op.
 - `extern crate` syntax is [entirely optional for Rust 2018](https://doc.rust-lang.org/nightly/edition-guide/rust-2018/module-system/path-clarity.html) and [implicitly discouraged](https://github.com/rust-lang/rust/pull/53684), so I've removed all references of the syntax, including examples.
 - Looks like #269 suggested migrating to criterion and we already use it, so I've just converted the `lib.rs` benchmark to use criterion now.
 - On that note, I've also added some corpus to bench against. Specifically, I've added the JS reference benchmarks. I didn't find the other linked items to be as meaningful as the JS benchmarks, but I'm open to adding more. I've written it to be very easy to add more.
 - Also on that note, I've refactored the `lib.rs` to use more of criterion's tools. We should now be able to see how pathological codeblocks behave over a longer inputs. That being said, I didn't do a perfect migration as benching 1000 times seems excessive, and I don't have full context on the benchmark, so I'd request extra scrutiny here.


marcusklaas:
Thanks for this PR! The changelog looks very good. I will do a thorough review of the commits tomorrow.

benches/lib.rs
```diff
@@ -1,35 +1,38 @@
...
-            // grow faster than that.
-            let mut buf = String::new();
-            for i in 1..1000 {
-                buf.push_str(&"`".repeat(i));
-                buf.push(' ');
-            }
-
-            b.iter(|| render_html(&buf, Options::empty()));
-        });
+    pub fn pathological_codeblocks1(c: &mut Criterion) {
```
marcusklaas, line 7: This is my favorite commit. This code now actually directly measures what we're interested in: throughput. Bravo!

benches/spec.rs
```diff
@@ -0,0 +1,25 @@
...
+    let folder = read_dir("./benches/spec_samples").unwrap();
+    for entry in folder {
+        let entry = entry.unwrap();
+
+        if entry.metadata().unwrap().is_file() {
+            let filename = &entry.file_name().into_string().unwrap();
+            let corpus = read_to_string(entry.path()).unwrap();
+            let mut result = String::with_capacity(corpus.len() * 3 / 2);
+
+            c.bench_function(filename, |b| {
```
marcusklaas, line 15: Could we add a throughput measurement in here? That seems very useful in comparing with other parsers.

benches/spec_samples/README.md
```diff
@@ -0,0 +1,285 @@
+CommonMark
```
marcusklaas, line 1: Yes, more test files finally! Could you move these to `/third_party/CommonMark` so that it's clear where they came from?

edit: just realized that the javascript tests may have a different origin. If so, could you please add a separate third party folder for them, including a license file?

marcusklaas (approved):
Ok, so it ended up being even better than I expected. I especially love that you understood what the weird lib benchmarks were for despite it not being clear at all from the code and made them so that they are actually useful.

Left some minor comments in the review. Once they are addressed, I will be happy to merge this. Thanks again!

benches/spec_samples/README.md
edward-shen: The newly added samples are now correctly attributed to markdown-it, and a license has been added to reflect this.

benches/spec.rs
edward-shen: From my understanding, the throughput benchmarks are usually done with multiple inputs, e.g. how it scales with more input, so I don't think it works here since we'd only have one data point (the size of the input).

I've uploaded an [example report] of the current code though (extract and view `report.html`) -- perhaps it's sufficient?

[example report]: https://drive.google.com/file/d/1h8WpfF8QAs7ih7Ud40kaOa6JzGowyOXU/view?usp=sharing

marcusklaas: Ok, that makes sense. Let's leave it as is.

marcusklaas:
Thanks again!

PR #510, opened 2020-12-08
Fix the lifetimes on BrokenLinkCallback

jyn514:
Here is a simple callback which marks every link as working:

```rust
fn callback<'a>(link: BrokenLink<'a>) -> Option<(CowStr<'a>, CowStr<'a>)> {
    Some(("#".into(), link.reference.into()))
}
```

On its own, it typechecks fine. Unfortunately, this doesn't work with `Parser::with_broken_link_callback`:

```rust
fn f(txt: &str) {
    for _ in Parser::new_with_broken_link_callback(txt, Options::empty(), Some(&mut callback)) {
    }
}
```
```
error: implementation of `FnOnce` is not general enough
   --> src/lib.rs:8:80
    |
8   |       for _ in Parser::new_with_broken_link_callback(txt, Options::empty(), Some(&mut callback)) {
    |                                                                                  ^^^^^^^^^^^^^ implementation of `FnOnce` is not general enough
    |
   ::: /home/joshua/.local/lib/rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ops/function.rs:219:1
    |
219 | / pub trait FnOnce<Args> {
220 | |     /// The returned type after the call operator is used.
221 | |     #[lang = "fn_once_output"]
222 | |     #[stable(feature = "fn_once_output", since = "1.12.0")]
...   |
227 | |     extern "rust-call" fn call_once(self, args: Args) -> Self::Output;
228 | | }
    | |_- trait `FnOnce` defined here
    |
    = note: `for<'a> fn(pulldown_cmark::BrokenLink<'a>) -> Option<(pulldown_cmark::CowStr<'a>, pulldown_cmark::CowStr<'a>)> {callback}` must implement `FnOnce<(pulldown_cmark::BrokenLink<'_>,)>`
    = note: ...but `FnOnce<(pulldown_cmark::BrokenLink<'_>,)>` is actually implemented for the type `for<'a> fn(pulldown_cmark::BrokenLink<'a>) -> Option<(pulldown_cmark::CowStr<'a>, pulldown_cmark::CowStr<'a>)> {callback}`

error: aborting due to previous error
```
The issue is that `BrokenLinkCallback` is typed as having the same lifetime as its outputs: https://github.com/raphlinus/pulldown-cmark/blob/e97974b8d76195c953f0d427e8725ef9ad1a0c17/src/parse.rs#L1270
That means that the callback can't e.g. be passed to two different parsers, because the first call will fix a set lifetime:

```rust
 fn f(txt: &str) {
    let mut callback = |link: BrokenLink<'_>| -> Option<(CowStr<'_>, CowStr<'_>)> {
        Some(("#".into(), link.reference.to_owned().into()))
    };

    for _ in Parser::new_with_broken_link_callback(txt, Options::empty(), Some(&mut callback)) {
    }

    for _ in Parser::new_with_broken_link_callback(txt, Options::empty(), Some(&mut callback)) {
    }
}
```
```
error[E0499]: cannot borrow `callback` as mutable more than once at a time
  --> src/lib.rs:11:80
   |
8  |     for _ in Parser::new_with_broken_link_callback(txt, Options::empty(), Some(&mut callback)) {
   |                                                                                ------------- first mutable borrow occurs here
...
11 |     for _ in Parser::new_with_broken_link_callback(txt, Options::empty(), Some(&mut callback)) {
   |                                                                                ^^^^^^^^^^^^^
   |                                                                                |
   |                                                                                second mutable borrow occurs here
   |                                                                                first borrow later used here
```

The fix I was thinking of was something like this:

```diff
diff --git a/src/parse.rs b/src/parse.rs
index d6388b1..bccd68c 100644
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -129,12 +129,12 @@ pub struct BrokenLink<'a> {
 }

 /// Markdown event iterator.
-pub struct Parser<'a> {
-    text: &'a str,
+pub struct Parser<'input, 'callback: 'input> {
+    text: &'input str,
     options: Options,
     tree: Tree<Item>,
-    allocs: Allocations<'a>,
-    broken_link_callback: BrokenLinkCallback<'a>,
+    allocs: Allocations<'input>,
+    broken_link_callback: BrokenLinkCallback<'callback>,
     html_scan_guard: HtmlScanGuard,

     // used by inline passes. store them here for reuse
@@ -1266,8 +1267,8 @@ pub(crate) struct HtmlScanGuard {
     pub declaration: usize,
 }

-pub type BrokenLinkCallback<'a> =
-    Option<&'a mut dyn FnMut(BrokenLink) -> Option<(CowStr<'a>, CowStr<'a>)>>;
+pub type BrokenLinkCallback<'b> =
+    Option<&'b mut dyn for<'a> FnMut(BrokenLink<'a>) -> Option<(CowStr<'a>, CowStr<'a>)>>;

 /// Markdown event and source range iterator.
 ///
```
This does two things:
1. Separates the lifetime of the link from the lifetime of the callback (by changing `&'a FnMut() -> &'a str` to `&'b for<'a> FnMut() -> &'a str`).
2. Separates the lifetime of the link from the lifetime of the link (by adding a new `'callback` lifetime).

Unfortunately, this uncovers that the change *can't* work:

```
error[E0597]: `link_label` does not live long enough
   --> src/parse.rs:457:64
    |
145 |   impl<'a, 'b> Parser<'a, 'b> {
    |        -- lifetime `'a` defined here
...
450 |                                       .or_else(|| {
    |                                                -- value captured here
...
457 |                                                       reference: link_label.as_ref(),
    |                                                                  ^^^^^^^^^^ borrowed value does not live long enough
...
460 | /                                                 callback(broken_link).map(|(url, title)| {
461 | |                                                     (link_type.to_unknown(), url, title)
462 | |                                                 })
    | |__________________________________________________- returning this value requires that `link_label` is borrowed for `'a`
...
503 |                               }
    |                               - `link_label` dropped here while still borrowed
```

The issue is that `link_label` is only alive for the duration for the length of a single loop iteration, *not* the lifetime of the input. Even though it's parameterized by a lifetime `'input`, it has a [`Box` variant](https://docs.rs/pulldown-cmark/0.8.0/pulldown_cmark/enum.CowStr.html#variant.Boxed), so if it gets dropped the compiler has to conservatively assume the entire link is invalid.

To fix this, pass the `CowStr` itself into the callback, which allows
returning the same struct with the same lifetime.

This is a breaking change.

Closes https://github.com/raphlinus/pulldown-cmark/issues/509.
r? @marcusklaas

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -22,8 +22,8 @@
 
 use std::cmp::{max, min};
 use std::collections::{HashMap, VecDeque};
-use std::ops::{Index, Range};
 use std::iter::FusedIterator;
+use std::ops::{Index, Range};
 
 use unicase::UniCase;
 
@@ -125,16 +125,16 @@ impl<'a> Default for ItemBody {
 pub struct BrokenLink<'a> {
     pub span: std::ops::Range<usize>,
     pub link_type: LinkType,
-    pub reference: &'a str,
+    pub reference: CowStr<'a>,
 }
 
 /// Markdown event iterator.
-pub struct Parser<'a> {
-    text: &'a str,
+pub struct Parser<'input, 'callback> {
+    text: &'input str,
     options: Options,
     tree: Tree<Item>,
-    allocs: Allocations<'a>,
-    broken_link_callback: BrokenLinkCallback<'a>,
+    allocs: Allocations<'input>,
+    broken_link_callback: BrokenLinkCallback<'input, 'callback>,
     html_scan_guard: HtmlScanGuard,
 
     // used by inline passes. store them here for reuse
@@ -142,14 +142,14 @@ pub struct Parser<'a> {
     link_stack: LinkStack,
 }
 
-impl<'a> Parser<'a> {
+impl<'a, 'b> Parser<'a, 'b> {
     /// Creates a new event iterator for a markdown string without any options enabled.
-    pub fn new(text: &'a str) -> Parser<'a> {
+    pub fn new(text: &'a str) -> Self {
         Parser::new_ext(text, Options::empty())
     }
 
     /// Creates a new event iterator for a markdown string with given options.
-    pub fn new_ext(text: &'a str, options: Options) -> Parser<'a> {
+    pub fn new_ext(text: &'a str, options: Options) -> Self {
         Parser::new_with_broken_link_callback(text, options, None)
     }
 
@@ -161,8 +161,8 @@ impl<'a> Parser<'a> {
     pub fn new_with_broken_link_callback(
         text: &'a str,
         options: Options,
-        broken_link_callback: BrokenLinkCallback<'a>,
-    ) -> Parser<'a> {
+        broken_link_callback: BrokenLinkCallback<'a, 'b>,
+    ) -> Self {
         let (mut tree, allocs) = run_first_pass(text, options);
         tree.reset();
         let inline_stack = Default::default();
@@ -453,7 +453,7 @@ impl<'a> Parser<'a> {
                                                 let broken_link = BrokenLink {
                                                     span: (self.tree[tos.node].item.start)..end,
                                                     link_type,
-                                                    reference: link_label.as_ref(),
+                                                    reference: link_label,
                                                 };
 
                                                 callback(broken_link).map(|(url, title)| {
@@ -848,7 +848,7 @@ impl<'a> Parser<'a> {
     /// Consumes the event iterator and produces an iterator that produces
     /// `(Event, Range)` pairs, where the `Range` value maps to the corresponding
     /// range in the markdown source.
-    pub fn into_offset_iter(self) -> OffsetIter<'a> {
+    pub fn into_offset_iter(self) -> OffsetIter<'a, 'b> {
         OffsetIter { inner: self }
     }
 }
@@ -1266,8 +1266,8 @@ pub(crate) struct HtmlScanGuard {
     pub declaration: usize,
 }
 
-pub type BrokenLinkCallback<'a> =
-    Option<&'a mut dyn FnMut(BrokenLink) -> Option<(CowStr<'a>, CowStr<'a>)>>;
+pub type BrokenLinkCallback<'input, 'borrow> =
+    Option<&'borrow mut dyn FnMut(BrokenLink<'input>) -> Option<(CowStr<'input>, CowStr<'input>)>>;
 
 /// Markdown event and source range iterator.
 ///
@@ -1276,11 +1276,11 @@ pub type BrokenLinkCallback<'a> =
 ///
 /// Constructed from a `Parser` using its
 /// [`into_offset_iter`](struct.Parser.html#method.into_offset_iter) method.
-pub struct OffsetIter<'a> {
-    inner: Parser<'a>,
+pub struct OffsetIter<'a, 'b> {
+    inner: Parser<'a, 'b>,
 }
 
-impl<'a> Iterator for OffsetIter<'a> {
+impl<'a, 'b> Iterator for OffsetIter<'a, 'b> {
     type Item = (Event<'a>, Range<usize>);
 
     fn next(&mut self) -> Option<Self::Item> {
@@ -1403,7 +1403,7 @@ fn item_to_event<'a>(item: Item, text: &'a str, allocs: &Allocations<'a>) -> Eve
     Event::Start(tag)
 }
 
-impl<'a> Iterator for Parser<'a> {
+impl<'a, 'b> Iterator for Parser<'a, 'b> {
     type Item = Event<'a>;
 
     fn next(&mut self) -> Option<Event<'a>> {
@@ -1433,7 +1433,7 @@ impl<'a> Iterator for Parser<'a> {
     }
 }
 
-impl<'a> FusedIterator for Parser<'a> {}
+impl FusedIterator for Parser<'_, '_> {}
 
 #[cfg(test)]
 mod test {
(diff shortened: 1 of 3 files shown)


src/parse.rs
```diff
@@ -1764,4 +1764,35 @@ mod test {
...
+
+        for _ in Parser::new_with_broken_link_callback("static lifetime", Options::empty(), Some(&mut closure)) {
+        }
+        /* This fails to compile. Because the closure can't say `for <'a> fn(BrokenLink<'a>) ->
+         * CowStr<'a>` and has to use the enclosing `'b` lifetime parameter, `temporary_str` lives
+         * shorter than `'b`. I think this is unlikely to occur in real life, and if it does, the
+         * fix is simple: move it out to a function that allows annotating the lifetimes.
+         */
+        //for _ in Parser::new_with_broken_link_callback(&temporary_str, Options::empty(), Some(&mut callback)) {
+        //}
```
jyn514, line 1792: Let me know if I should make this a `compile_fail` doctest instead; I'm not sure whether this is a little too much to hit people with at once, but it would make sure that this is tested.

jyn514:
Hmm, I'm not sure how this interacts with https://github.com/raphlinus/pulldown-cmark/pull/497.

src/parse.rs
```diff
@@ -125,31 +125,31 @@ impl<'a> Default for ItemBody {
 pub struct BrokenLink<'a> {
     pub span: std::ops::Range<usize>,
     pub link_type: LinkType,
-    pub reference: &'a str,
+    pub reference: CowStr<'a>,
```
marcusklaas, line 128: Is this change required to solve the issue? This seems like the greatest breaking change.

src/parse.rs
```diff
@@ -1266,8 +1266,8 @@ pub(crate) struct HtmlScanGuard {
     pub declaration: usize,
 }
 
-pub type BrokenLinkCallback<'a> =
-    Option<&'a mut dyn FnMut(BrokenLink) -> Option<(CowStr<'a>, CowStr<'a>)>>;
+pub type BrokenLinkCallback<'input, 'borrow> =
```
marcusklaas, line 1269: This type signature looks good. How does this compare to the previous type? The old type can be expressed as `BrokenLinkCallback<'a, 'a>`, right? If so, this would be a strict supertype. (and therefore not breaking unless explicit lifetimes were used by users)

src/parse.rs
marcusklaas: I think we can do without. We're not really testing a feature of the library by ensuring that that doesn't compile.

marcusklaas (commented):
Thanks for the PR. I think I understand the problem now, and this seems like proper solution. Could you clarify the points brought up in the review? I am mostly concerned with minimizing compatibility breaks. If we could do without the reference change and not break anyone using a closure as a callback, that would be excellent.

src/parse.rs
jyn514: Yes, this is required. Otherwise you run into lifetime issues in pulldown itself - see 'Unfortunately, this uncovers that the change can't work:' in the PR description.

There's another (hidden) issue here: if you don't specify the lifetimes to the closure, this will given an inference error:

```rust
        let mut closure = |link: BrokenLink| Some(("#".into(), link.reference.into()));
```
```
error: lifetime may not live long enough
    --> src/parse.rs:1775:46
     |
1775 |         let mut closure = |link: BrokenLink| Some(("#".into(), link.reference.into()));
     |                            ----            - ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ returning this value requires that `'1` must outlive `'2`
     |                            |               |
     |                            |               return type of closure is Option<(strings::CowStr<'_>, strings::CowStr<'2>)>
     |                            has type `parse::BrokenLink<'1>`
```
but I don't think this is actually an issue because you couldn't return a reference to the input before anyway, that was the whole issue. So I don't think this is actually regression in practice, although maybe it should be documented.

If you don't return a reference it works fine/.

src/parse.rs
jyn514: Yes, the previous type was equivalent to `BrokenLinkCallback<'input, 'input>` (where `'input` is the lifetime of the markdown).

camelid: > (and therefore not breaking unless explicit lifetimes were used by users)

Probably will be breaking for people who write top-level functions that take `BrokenLink<'_>` (*ahem*, me) unfortunately. But, it will be very easy to adapt to (just do `BrokenLink<'_, '_>`).

src/parse.rs
```diff
@@ -125,31 +125,31 @@ impl<'a> Default for ItemBody {
...
+    broken_link_callback: BrokenLinkCallback<'input, 'callback>,
     html_scan_guard: HtmlScanGuard,
 
     // used by inline passes. store them here for reuse
     inline_stack: InlineStack,
     link_stack: LinkStack,
 }
 
-impl<'a> Parser<'a> {
+impl<'a, 'b> Parser<'a, 'b> {
```
camelid, line 145: It's probably better to use the lifetime names declared on the `Parser` struct so it's not confusing for pulldown developers and people reading the docs.

src/parse.rs
jyn514: @camelid `BrokenLink` isn't changing, just `BrokenLinkCallback`. This is a private type alias.

camelid: Oops, somehow missed that :)

marcusklaas: It was made public to resolve https://github.com/raphlinus/pulldown-cmark/issues/498.

marcusklaas:
Ok, I think we can merge this. I was considering folding the parameterization of the callback in this PR, but that can be done in a different pull request. That's a further reaching change.

Thanks for this PR!

jyn514:
Thanks for your patience, I know this is a bit of a complicated problem :laughing:

jyn514:
Hi @marcusklaas, are you planning to do a release any time soon? It would be useful to have this fix published.

PR #512, opened 2021-01-02
Add method to provide access to the refdef tracker

marcusklaas:
Currently there is no way for users to get information about reference definitions. So far, the following use cases have been identified for it:

- recreating markdown documents (e.g. markdown-to-markdown formatting)
- linting on unused definitions
- reporting errors at link definition (for [rustdoc](https://github.com/raphlinus/pulldown-cmark/issues/423#issuecomment-753297494))

Instead of emitting these definitions as events, we provide a separate call for these. These were the considerations:

- not all renderers will be interested in refdefs, so we should provide them as opt-in to prevent unneeded complexity
- the current exposed structure provides quick lookups, as the underlying datastructure is a hashmap

This new functionality will only really start to shine when link references labels are exposed (https://github.com/raphlinus/pulldown-cmark/issues/434). That is not included in this pull request for two reasons: it will be a breaking change, and the ideal design for exposing reference labels is not yet clear. Simply adding it to the link event (type) will blow up the size of all items by 3 words, which will have non-negligible impact on performance.

Until then the reference label can be recovered in most cases by slicing the source document using the link's span (parsing up to the first `]`).

Closes https://github.com/raphlinus/pulldown-cmark/issues/423.

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -1210,7 +1210,7 @@ pub(crate) struct Allocations<'a> {
 }
 
 /// Keeps track of the reference definitions defined in the document.
-#[derive(Clone)]
+#[derive(Clone, Default)]
 pub struct RefDefs<'input>(pub(crate) HashMap<LinkLabel<'input>, LinkDef<'input>>);
 
 impl<'input, 'b, 's> RefDefs<'input>
@@ -1231,7 +1231,7 @@ where
 impl<'a> Allocations<'a> {
     pub fn new() -> Self {
         Self {
-            refdefs: RefDefs(HashMap::new()),
+            refdefs: RefDefs::default(),
             links: Vec::with_capacity(128),
             cows: Vec::new(),
             alignments: Vec::new(),
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -180,6 +180,12 @@ impl<'input, 'callback> Parser<'input, 'callback> {
         }
     }
 
+    /// Returns a reference to the internal `RefDefs` object, which provides access
+    /// to the internal map of reference definitions.
+    pub fn reference_definitions(&self) -> &RefDefs {
+        &self.allocs.refdefs
+    }
+
     /// Handle inline markup.
     ///
     /// When the parser encounters any item indicating potential inline markup, all
@@ -435,7 +441,7 @@ impl<'input, 'callback> Parser<'input, 'callback> {
                                 let type_url_title = self
                                     .allocs
                                     .refdefs
-                                    .get(&UniCase::new(link_label.as_ref().into()))
+                                    .get(link_label.as_ref())
                                     .map(|matching_def| {
                                         // found a matching definition!
                                         let title = matching_def
@@ -1131,10 +1137,12 @@ enum LinkStackTy {
     Disabled,
 }
 
+/// Contains the destination URL, title and source span of a reference definition.
 #[derive(Clone)]
-pub(crate) struct LinkDef<'a> {
+pub struct LinkDef<'a> {
     pub dest: CowStr<'a>,
     pub title: Option<CowStr<'a>>,
+    pub span: Range<usize>,
 }
 
 /// Tracks tree indices of code span delimiters of each length. It should prevent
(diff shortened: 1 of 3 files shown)


src/parse.rs
```diff
@@ -1195,16 +1203,35 @@ pub(crate) struct AlignmentIndex(usize);
...
     alignments: Vec<Vec<Alignment>>,
 }
 
+/// Keeps track of the reference definitions defined in the document.
+#[derive(Clone)]
+pub struct RefDefs<'input>(pub(crate) HashMap<LinkLabel<'input>, LinkDef<'input>>);
+
+impl<'input, 'b, 's> RefDefs<'input>
+where
+    's: 'b,
```
jyn514, line 1218: Is this bound necessary? I'm pretty sure it will make the output of `get()` unusable since it can't be used after the `self` borrow is released.

jyn514:
It would be great to see an example of how to use this :) both as documentation and as a test for things like https://github.com/raphlinus/pulldown-cmark/pull/512#discussion_r550905356.

src/parse.rs
marcusklaas: It appears that it is necessary. And I think that it means something different: that the output of `get` cannot outlive `self` (not the borrow of self), which is indeed required for memory safety.

To confirm this, I added `let linkdef = parser.reference_definitions().get("a b c").unwrap();` to the test case and it compiles. That seems to confirm that the linkdef reference can indeed outlive the `RefDefs` reference.

jyn514: Try this:
```diff
diff --git a/src/parse.rs b/src/parse.rs
index 35e9647..3f8dfdf 100644
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -1203,7 +1203,7 @@ pub(crate) struct AlignmentIndex(usize);
 
 #[derive(Clone)]
 pub(crate) struct Allocations<'a> {
-    pub refdefs: RefDefs<'a>,
+    pub refdefs: RefDefs<'a, 'a>,
     links: Vec<(LinkType, CowStr<'a>, CowStr<'a>)>,
     cows: Vec<CowStr<'a>>,
     alignments: Vec<Vec<Alignment>>,
@@ -1211,19 +1211,16 @@ pub(crate) struct Allocations<'a> {
 
 /// Keeps track of the reference definitions defined in the document.
 #[derive(Clone)]
-pub struct RefDefs<'input>(pub(crate) HashMap<LinkLabel<'input>, LinkDef<'input>>);
+pub struct RefDefs<'input, 'output>(pub(crate) HashMap<LinkLabel<'input>, LinkDef<'output>>);
 
-impl<'input, 'b, 's> RefDefs<'input>
-where
-    's: 'b,
-{
+impl<'input, 'output> RefDefs<'input, 'output> {
     /// Performs a lookup on reference label using unicode case folding.
-    pub fn get(&'s self, key: &'b str) -> Option<&'b LinkDef<'input>> {
+    pub fn get<'s>(&'s self, key: &'input str) -> Option<&'s LinkDef<'output>> {
         self.0.get(&UniCase::new(key.into()))
     }
 
     /// Provides an iterator over all the document's reference definitions.
-    pub fn iter(&'s self) -> impl Iterator<Item = (&'s str, &'s LinkDef<'input>)> {
+    pub fn iter<'s>(&'s self) -> impl Iterator<Item = (&'s str, &'s LinkDef<'output>)> {
         self.0.iter().map(|(k, v)| (k.as_ref(), v))
     }
 }
```
That
- Decouples the lifetime of `LinkLabel` from `LinkDef`
- Returns a `LinkDef` that can live longer than the `key`

I think previously this case would break:
```rust
let parser = unimplemented!();
let def = {
    let key = String::from("hello");
    parser.allocs.refdefs.get(&key)
};
```

jyn514: That said my example is sort of convoluted and it's probably ok to just not fix this.

marcusklaas: Ah, indeed that case breaks. But even with your patch, I cannot get it to compile. From my understanding, it still demands that they key lives as long as the hashmap keys.

It should obviously be possible to do this. I believe that `LinkLabel<'a>` should implement `Borrow<LinkLabel<'b>> where 'a: 'b` to enable it.

I have a lot more confidence in your lifetime analysis expertise, so please let me know when I'm missing something. I'll keep tinkering in the meantime to try to get your case to compile.

marcusklaas: Ok, I did more digging and asked for help on the xi zulip chat [here](https://xi.zulipchat.com/#narrow/stream/255911-rust/topic/hashmap.20get.20with.20temp.20keys/near/221519321). It seems that this is a limitation of the Hashmap API (and ultimately Rust itself).

src/parse.rs
```diff
@@ -1195,16 +1203,35 @@ pub(crate) struct AlignmentIndex(usize);
...
+    pub fn iter(&'s self) -> impl Iterator<Item = (&'s str, &'s LinkDef<'input>)> {
+        self.0.iter().map(|(k, v)| (k.as_ref(), v))
+    }
+}
+
 impl<'a> Allocations<'a> {
     pub fn new() -> Self {
         Self {
-            refdefs: HashMap::new(),
+            refdefs: RefDefs(HashMap::new()),
```
raphlinus, line 1234: Generally I prefer to `#[derive(Default)]` on the type and use `default` here, but that's a style point.

raphlinus (commented):
Sure, this looks good. The lifetimes are slightly scary, but hopefully won't be a problem in real use.

src/parse.rs
marcusklaas: Yeah, that's cleaner. I'll make the change.

PR #513, opened 2021-01-05
Use enum for heading level.

### LINKED ISSUE #501: Heading levels
The `Tag` enum includes:
```
    /// A heading. The field indicates the level of the heading.
    Heading(u32),
```
HTML and Markdown specs are both explicit that heading levels 1-6 are valid (in HTML called `h1`..`h6`). From this pulldown-cmark spec. it is not clear whether other values may occur (and thus how consumers should handle unexpected values).

Suggestion:
```rust
#[repr(u32)]
enum Heading {
    H1 = 1,
    H2,
    // ...
}
```
This supports `Heading::H6 as u32` and thus is just a small extension over the current status. Alternatively documentation should clarify exactly which values may appear.

edward-shen:
Fixes #501. As discussed, this is a minor breaking change. 

This diff converts the `u32` that we used as the header level into a enum, `HeadingLevel`.

High level changelog:
 - Introduced a new public enum, `HeadingLevel`
 - As this is a public enum, all easily derived and implementable common traits are implemented, including `TryFrom` and `Display`.
 - Use the `Display` trait internally, so `start_tag` and `end_tag` no longer prefixes `h` in the respective areas.
 - Since `HeadingLevel` implements `TryFrom`, `scan_atx_heading` now uses `try_from` instead of hard-coding the heading level bounds in the function.

### DIFF
--- a/src/firstpass.rs
+++ b/src/firstpass.rs
@@ -3,12 +3,15 @@
 
 use std::cmp::max;
 
-use crate::linklabel::{scan_link_label_rest, LinkLabel};
 use crate::parse::{scan_containers, Allocations, Item, ItemBody, LinkDef};
 use crate::scanners::*;
 use crate::strings::CowStr;
 use crate::tree::{Tree, TreeIndex};
 use crate::Options;
+use crate::{
+    linklabel::{scan_link_label_rest, LinkLabel},
+    HeadingLevel,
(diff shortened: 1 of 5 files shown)


marcusklaas:
This looks good. Thank you Edward!

PR #515, opened 2021-01-20
Add serde suport

aDotInTheVoid:
For the tests, I just tested the `CowStr` implementation, as everything else is derived.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -27,6 +27,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cdb031dd78e28731d87d56cc8ffef4a8f36ca26c38fe2de700543e627f8a464a"
 
 [[package]]
+name = "bincode"
+version = "1.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f30d3a39baa26f9651f17b375061f3233dde33424a8b72b0dbe93a68a0bc896d"
+dependencies = [
+ "byteorder",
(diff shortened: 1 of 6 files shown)


marcusklaas:
Hi Nixon, thanks for your PR. Optional serde features seem very popular, so we will naturally consider it seriously. 

This is by no means a dismissal of your work, but I have some trouble immediately thinking of applications for serialization of these types. For review it would be most useful to have some motivation for these changes.

As a first note: your last commit seems to remove the `visit_string` method instead of the `visit_borrowed_str`.

aDotInTheVoid:
> I have some trouble immediately thinking of applications for serialization of these types.

The use case I had in mind was preprocessors for a static site generator. The idea would be to allow external programs to opperate on the markdown in a modular and external way. 

While this could be done by sending across markdown as a string, parsing into `Vec<Event>` and then transforming, what would the return type be. You don't want to render to HTML, because the next preprocessor is expecting markdown. So the options are to return the `Event`s or turn them back into markdown. The first seemed easier, as it's mostly just slaping on some serde derives. Also it would be more efficient as you only need to parse the markdown once, not for every preprocessor.

> As a first note: your last commit seems to remove the visit_string method instead of visit_borrowed_str.

The name was wrong, I ment to remove `visit_string`. I wasn't sure if it was needed before testing, but it looks like it isn't. I've renamed the commit.

marcusklaas:
As far as I can tell, this looks good. Thanks for your contribution!

PR #519, opened 2021-02-24
Make `html` and `escape` modules optional

jfrimmel:
This PR introduces a new enabled-by-default-feature `html`. If this feature is enabled, there are no functional differences. If it is disabled, the `html` and `escape` modules are _not_ compiled. This is helpful, if a dependent crate does not require any HTML rendering.

Note, that this will __break__ user crates, that use `--no-default-features` and also use one of the two modules in question. Therefore I understand, if this PR cannot be merged.

Note, that some tests and examples need new annotations, as they require the `html` module.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,18 @@ harness = false
 name = "markdown-it"
 harness = false
 
+[[example]]
+name = "event-filter"
+required-features = ["html"]
+
+[[example]]
+name = "string-to-string"
+required-features = ["html"]
+
+[[example]]
+name = "broken-link-callbacks"
+required-features = ["html"]
+
 [dependencies]
 bitflags = "1.2"
 unicase = "2.6"
@@ -48,6 +60,7 @@ serde_json = "1.0.61"
 bincode = "1.3.1"
 
 [features]
-default = ["getopts"]
+default = ["getopts", "html"]
 gen-tests = []
 simd = []
+html = []
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -14,6 +14,8 @@ steps:
     displayName: Cargo test with simd feature enabled
   - script: cargo test --all --features=serde
     displayName: Cargo test with serde feature enabled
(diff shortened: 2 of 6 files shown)


marcusklaas:
Thank you for your thoughtful PR. I am sympathetic to this change, but would like one more pair of eyes on this before I merge.

@raphlinus: does this look a good idea to you as well?

raphlinus:
Hmm, I have mixed feelings about it. I certainly understand the desire to keep compile times down and avoid bloat, but these modules don't strike me as especially heavy. The experience with default features is that they're almost always left on, and are somewhat inconvenient to control.

Do we have any kind of quantitative measurement about how much compile time is saved here?

jfrimmel:
I understand the mixed feelings :wink: 

I've run some measurements with [hyperfine](https://github.com/sharkdp/hyperfine) on my laptop (crates were all downloaded):
image

raphlinus:
That's a nontrivial improvement. I'm ok with this change. If the time delta had been much smaller I would have said it's not worth it. Thanks for the PR and measurements :)

jfrimmel:
Oh, I've noticed something:
I think the previous comparison was not 100% fair, as there is another additional crate built (`getopts`) when default features are active.

Here is a better benchmark, which shows the real difference:
image

ehuss:
I would caution against making a breaking change like this.  There are a number of projects that use default-features=false (like rustdoc, clippy, etc.).

Also, to me, 0.7 seconds does not seem worth a change.

jfrimmel:
Is there anything to do here? Should this PR be closed or left open until a decision is reached?

Martin1887:
This makes a lot of sense to me, and having an HTML renderer in a parser crate was a bit shocking the first time I checked the project. I think that both features should be in different crates, but keeping the current minimal functionality as optional features can make sense for tests.

All implications must be carefully reviewed though, so I will be back on this relatively soon.

Related to #526.

Martin1887:
This pull request is interesting for the 0.10 release, merged

PR #525, opened 2021-05-13
add rendering options for target="_blank"

ovidiu-ionescu:
Added options for the renderer, similar to the options for the parser.

Currently there is only one option, to render links that open in a new tab by adding `target="_blank` to the rendered anchor elements.

### DIFF
--- a/src/html.rs
+++ b/src/html.rs
@@ -27,7 +27,7 @@ use crate::escape::{escape_href, escape_html, StrWrite, WriteWrapper};
 use crate::strings::CowStr;
 use crate::Event::*;
 use crate::{Alignment, CodeBlockKind, Event, LinkType, Tag};
-use crate::{RenderingOptions};
+use crate::RenderingOptions;
 
 enum TableState {
     Head,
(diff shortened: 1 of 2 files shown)


src/html.rs
```diff
@@ -27,6 +27,7 @@ use crate::escape::{escape_href, escape_html, StrWrite, WriteWrapper};
 use crate::strings::CowStr;
 use crate::Event::*;
 use crate::{Alignment, CodeBlockKind, Event, LinkType, Tag};
+use crate::{RenderingOptions};
```
oberien, line 30:
```suggestion
use crate::RenderingOptions;
```

Keats:
I've added a few of these in Zola (https://github.com/getzola/zola/blob/master/components/config/src/config/markup.rs#L40-L62 for the various options). Would be nice if those options https://github.com/getzola/zola/blob/master/components/config/src/config/markup.rs#L16-L22 were available

ovidiu-ionescu:
> I've added a few of these in Zola (https://github.com/getzola/zola/blob/master/components/config/src/config/markup.rs#L40-L62 for the various options). Would be nice if those options https://github.com/getzola/zola/blob/master/components/config/src/config/markup.rs#L16-L22 were available

Google's Lighthouse insists that links with target="blank" should also specify rel noopener or noreferrer. I included noreferrer as it is the more complete of the two.

Martin1887:
Sorry, but this is out of scope of this project and it should be implemented in an external HTML renderer crate. See #645.

Issue #528, opened 2021-05-20
Questions about the code


vedantroy:
I've been browsing the code to get an understanding of it & I have a few questions if you wouldn't mind answering!

- Why is `TreeIndex` a wrapper for `NonZeroUsize` instead of a wrapper for `usize`? (I saw https://github.com/raphlinus/pulldown-cmark/pull/184 but didn't really understand it).
- In `scan_containers` we `clone` `line_start` before calling `scan_blockquote_marker`, but `scan_blockquote_marker` seems to save/restore internally, so why do we do it twice?

marcusklaas:
Hi Vedant, I'm happy to answer such questions!

> Why is `TreeIndex` a wrapper for `NonZeroUsize` instead of a wrapper for `usize`? (I saw #184 but didn't really understand it).

It is to allow for an optimization by the rust compiler. It understands that `NonZeroUsize` is never represented by the null word (all zeroes). So when it is wrapped in an `Option`, it can use that value to represent the `None` value. This is why `Option<TreeIndex>`, which is used a lot in pulldown's datastructures, is only one word long. Translated to machine code this should be exactly the same as using `usize` and remembering that `0` is never used, but this enforces things through the type system.

> In `scan_containers` we `clone` `line_start` before calling `scan_blockquote_marker`, but `scan_blockquote_marker` seems to save/restore internally, so why do we do it twice?

And that is because I wasn't being as sharp as you when I wrote that bit - excellent catch! Indeed we need not clone `line_start` there. Pull requests are welcomed! ;-)

vedantroy:
Thanks for the response! I'm curious about what is going on here: https://github.com/raphlinus/pulldown-cmark/blob/8db4423ce78a1dad8a8377e0f25ebd48e0e534fc/src/firstpass.rs#L133.

Specifically, do we do nothing if there's a `BlockQuote`, but otherwise call `pop`?

marcusklaas:
Yeah, that bit is somewhat tricky. All containers that are not blockquotes are ended by blank lines, or two in the case of list items. Blockquotes can have an arbitrary number of blank lines *within* them. Note that at the point in the code that you linked to, all the containers that are open have been scanned.

[...]

PR #529, opened 2021-05-20
Remove redundant save

vedantroy:
Remove the redundant save/restore as discussed in https://github.com/raphlinus/pulldown-cmark/issues/528.

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -865,9 +865,8 @@ pub(crate) fn scan_containers(tree: &Tree<Item>, line_start: &mut LineStart) -> 
     for &node_ix in tree.walk_spine() {
         match tree[node_ix].item.body {
             ItemBody::BlockQuote => {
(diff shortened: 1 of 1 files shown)


vedantroy:
While I'm here, I might as well ask, what is the purpose of `scan_containers`?

It seems to maybe implement this part of the algorithm from the [GFM spec](https://github.github.com/gfm/#phase-1-block-structure):

> First we iterate through the open blocks, starting with the root document, and descending through last children down to the last open block. Each block imposes a condition that the line must satisfy if the block is to remain open. For example, a block quote requires a > character. A paragraph requires a non-blank line. In this phase we may match all or just some of the open blocks. But we cannot close unmatched blocks yet, because we may have a lazy continuation line.

marcusklaas:
Thanks for the PR! And yes, that is exactly what `scan_containers` is meant to do!

PR #535, opened 2021-08-10
Fix unsafe_code lint control logic to work with simd

### LINKED ISSUE #534: Test suite fails
Probably a symptom of newer Rust toolchains, but the test suite currently fails 8 tests and eventually aborts for me using current Rust stable.

alerque:
Fixes #534


marcusklaas:
Thank you for the PR!

PR #542, opened 2021-11-17
Update ubuntu version in azure build description

marcusklaas:
We are currently trying to use an image that no longer is being offered. Moving to Ubuntu 20.04 should fix the build.


marcusklaas:
This goes against our policy of having every change peer reviewed, but since it only touches the build code which we kinda need to merge new PRs like the one for custom headers, I'll go ahead and merge this myself..

PR #400, opened 2019-09-09
Properly deal with refdefs inside blockquotes

marcusklaas:
There was an edge case with code spans within refdef labels. We relied on jumping to the next node whenever we encountered a newline inside a block (like a list or a blockquote), but code spans are always one node. This PR should cover this as well.

Closes #398.

Issue #521, opened 2021-03-22
Panic when parsing fuzzed file with malformed HTML in it


5225225:
There's already the support for fuzzing here, but I found this issue through https://github.com/rust-fuzz/targets and wrote a simple `cargo-fuzz` fuzzing target that *only* checks that parsing arbitrary HTML doesn't panic. (Might be worth having a cargo-fuzz target in addition to the existing one, simply because cargo fuzz is easier to get set up and seems to be able to run test cases faster? Though I know testing for panics isn't the main purpose of the existing fuzzer)

```rust
#![no_main]
use libfuzzer_sys::fuzz_target;

fuzz_target!(|data: &str| {
    let mut output = String::new();
    let parser = pulldown_cmark::Parser::new(data);
    pulldown_cmark::html::push_html(&mut output, parser);
});
```

Below's a standalone demo program. I tested that it crashes both with 0.8.0 from crates.io, as well as the latest version from git (d99667b3, if just going to the directory in `~/.cargo/git/checkouts` is correct).

```rust
fn main() {
    let mut output = String::new();
    let parser = pulldown_cmark::Parser::new(">\n >>><N\n");
    pulldown_cmark::html::push_html(&mut output, parser);
}
```

Backtrace is below

```
thread 'main' panicked at 'range start index 4 out of range for slice of length 3', /home/jess/.cargo/git/checkouts/pulldown-cmark-0e08cefe6e291038/d99667b/src/scanners.rs:1061:45
stack backtrace:
   0: rust_begin_unwind
             at /rustc/61edfd591cedff66fca639c02f66984f6271e5a6/library/std/src/panicking.rs:493:5
   1: core::panicking::panic_fmt
             at /rustc/61edfd591cedff66fca639c02f66984f6271e5a6/library/core/src/panicking.rs:92:14
   2: core::slice::index::slice_start_index_len_fail
             at /rustc/61edfd591cedff66fca639c02f66984f6271e5a6/library/core/src/slice/index.rs:34:5
   3: <core::ops::range::RangeFrom<usize> as core::slice::index::SliceIndex<[T]>>::index
             at /home/jess/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/slice/index.rs:322:13
   4: core::slice::index::<impl core::ops::index::Index<I> for [T]>::index
[...]

PR #545, opened 2021-11-21
Fix slice error when scanning HTML

5225225:
The core issue here is the one line change in `src/parse.rs`. I'm not entirely sure *why* the closure was using `bytes` from the environment and not the one being passed in, but all the tests pass with this change, so :shrug:.

Looks like that change was added in #400 . 

I've added a debug assert to hopefully make it more clear if this crops up again.

### DIFF
--- a/tests/suite/mod.rs
+++ b/tests/suite/mod.rs
@@ -4,10 +4,10 @@
 pub use super::test_markdown_html;
 
 mod footnotes;
-mod table;
-mod regression;
-mod spec;
-mod smart_punct;
-mod gfm_table;
 mod gfm_strikethrough;
+mod gfm_table;
 mod gfm_tasklist;
+mod regression;
+mod smart_punct;
+mod spec;
+mod table;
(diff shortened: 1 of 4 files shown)


tests/suite/mod.rs
```diff
@@ -4,10 +4,10 @@
 pub use super::test_markdown_html;
 
 mod footnotes;
```
5225225, line 6: Not sure why rustfmt wanted to do this, did it start reordering imports recently?

Should I revert this or just leave it?

5225225:
This does *not* fix the case mentioned in https://github.com/raphlinus/pulldown-cmark/issues/521#issuecomment-806705216 (`" \u{b}\\\r- "`), I figure it's best to at least fix this case, and then have a look at that one.

marcusklaas:
Thank you so much for this pull request! It is indeed very peculiar that we didn't use the bytes passed to it directly... Wish I could tell you why it was like that. Regardless, thanks for the fix and the new test cases. This is a great contribution!

tests/suite/mod.rs
marcusklaas: It's fine :-)

PR #546, opened 2021-11-22
Sort auto-generated tests modules by spec names

lo48576:
By this change, `tests/suite/mod.rs` will be generated as an already formatted code.
This prevents rustfmt from creating new diffs.


marcusklaas:
Thanks!

Issue #533, opened 2021-06-03
Generating align attribute for th and td tags is invalid HTML


GuillaumeGomez:
I read through https://github.github.com/gfm/#example-199 and #145 to understand a bit better and I'm not sure what's the best course of action here: the [HTML spec](https://html.spec.whatwg.org/multipage/tables.html#the-td-element) for `td` (and `th` just below) don't like `align` as a valid attribute.

For more context, I discovered this issue when working on https://github.com/rust-lang/rust/pull/84480. The goal is to check that rustdoc generated pages are completely valid by using `tidy`. The problem is that it's an invalid attribute because not listed in the spec. The solution would be to rename `align` into `data-align`. However, the browsers are using this attribute to change the content display. So maybe the best thing to do is to simply ignore this warning and keep things as is... If so, please just close the issue.

Mazwak:
I have this problem in a zola generated site.

The correct property seems to be "text-align".

https://www.w3schools.com/Css/css_table_align.asp

marcusklaas:
Yeah, it's tricky, because indeed this should be done using the css property `text-align`, but if we'd set that inline, I think we would override other css rules. Inline css isn't great in any case. An alternative would be to set `data-align` as Guillaume suggested, but then it wouldn't display correctly _without_ any css. I'm not sure what the best solution is here..

Mazwak:
I don’t understand what using 'text-align' would break.

If I use ':----' syntax, I want to align my columns left. I want it to override the default 'text-align'.
Since colgroup are very limited in style, it has to be inline, direct or class. [1]
'align', which does not work, is already inlined. It just needs to be replaced with 'text-align'

[1] Knowing align in advance, with a table id (generated by pulldown-cmark) and nth-child, it might be possible to not inline it.
Something like :
```css
#cmarkid001 tr :nth-child(2) {
  text-align: center;
}
```

marcusklaas:
You've convinced me that indeed inline CSS is the way to go. I've created a pull request to make the change.

marcusklaas:
This has been fixed!

PR #552, opened 2021-12-17
Use inline CSS for table cell text alignment

marcusklaas:
This seems to be the best solution to https://github.com/raphlinus/pulldown-cmark/issues/533.


raphlinus (approved):
Seems reasonable to me, thanks. This is one of those cases where it would be nice to have an official spec extension to say what the translated HTML should be, but in any case I think this should work.

Issue #553, opened 2021-12-19
Incorrect paragraph interrupting list following nested lists


marcusklaas:
Okay, this is a weird one. When following a nested list, it appears paragraphs can be interrupted by lists starting with an index greater than zero, even though the spec prohibits this. First discovered [here](https://github.com/rust-lang/rust/issues/92005#issuecomment-997397508).

Minimal test case to reproduce this:
```md
1. a
   1. a

a
2. a
```

marcusklaas:
It has to do with allowing lists to interrupt paragraphs inside of list items in pulldown, which isn't quite right. But it also doesn't keep track of when we are inside a list item correctly. So it's really two bugs it looks like. An even more minimal test case:
```md
1. a
   2. b
```
This shouldn't start a nested list, but pulldown (and many others!) do.

PR #554, opened 2021-12-19
Fix some list interrupting paragraph incorrectness

marcusklaas:
This fixes https://github.com/raphlinus/pulldown-cmark/issues/553.


raphlinus (approved):
I can't say I fully understand the logic, but it seems like a reasonable change and I'm willing to let the tests have final say. It's a bit fascinating to me we're still seeing issues like this.

marcusklaas:
This is somewhat of an edge case, but I too had assumed that we'd finally gotten the list logic right 😅

PR #555, opened 2021-12-19
Update dependencies; update specs; bump version to 0.9.0

marcusklaas:
This includes some changes to the test suite that I had forgotten to include in the most recent two PRs. The tests in those PRs were up-to-date with these specs though.


marcusklaas:
Tests fail? 🤨

marcusklaas:
I investigate.

kaj:
It complains about `if` not allowed in const fn.  I think that implies it tries to compile with a rustc older than 1.46.0.  What is the minimum supported rust version for pulldown-cmark?   If we talk about a 0.9.0, that may be a good time to raise it?   The MSRV for the bitflags dependency is apparently 1.46.0.

marcusklaas:
Yeah, I think we can do a slight bump in MSRV so we can keep up with our dependencies. I'll update this PR later today.

kaj:
Great!  Looking forward to the release of 0.9.0!

raphlinus (approved):
Looks good, thanks! Also looking forward to the release.

PR #562, opened 2022-01-10
Add fuzzer that just parses &str input

5225225:


fuzz/fuzz_targets/parse.rs
```diff
@@ -0,0 +1,6 @@
+#![no_main]
+use libfuzzer_sys::fuzz_target;
+
+fuzz_target!(|data: &str| {
+    for _ in pulldown_cmark::Parser::new(data) {};
```
5225225, line 5: This should be fuzzing options too, since default options seem fine, but extended options crash quickly.

5225225:
Now, the output from the fuzzer looks like

```rust
FuzzingInput {
    markdown: "s!!!!!!!!!!E^\u{19}\u{0}\u{1}\u{0}\u{0}\u{0}\n*\t[][ >\n\t<p\n\t\t0o]Z\n\t[]\u{19}\t\u{19}&#x",
    tables: false,
    footnotes: false,
    strikethrough: false,
    tasklists: false,
    smart_punctuation: false,
    heading_attributes: false,
}
```

which is rather pretty :)

Could go farther and make the debug output of FuzzingInput be a valid test case.... but that seems a bit too weird, and it only saves a bit of time.

fuzz/fuzz_targets/parse.rs
```diff
@@ -0,0 +1,46 @@
...
+
+    if data.smart_punctuation {
+        opts.insert(Options::ENABLE_SMART_PUNCTUATION);
+    }
+
+    if data.heading_attributes {
+        opts.insert(Options::ENABLE_HEADING_ATTRIBUTES);
+    }
+
+    for _ in pulldown_cmark::Parser::new(data.markdown) {};
```
marcusklaas, line 45: Oops! It seems like you forgot to pass the options to the parser here 😄

marcusklaas:
Yesss, I love the idea of choosing the options randomly also! Who knows what weird interactions we might find.

So I think this is ready to go as soon as the minor oversight that was flagged has been fixed. Maybe it would be nice to automatically try to minimize any troublesome input we find also, but that's maybe fun for a follow-up PR some time.

Thanks!

marcusklaas:
Thank you for this! And no sweat about being the person that only reports. The pulldown internals are not super accessible (an import problem of itself!) and each of these reports is making pulldown a little better.

I think we should submit to the Google fuzzers once we've done the low-hanging fruit. They actually reached out to us at some point in last year, but I didn't have the energy to get into it at that time. Let's see if the offer still stands once we are ready.

Thanks again!

PR #564, opened 2022-01-11
Fix some panics found by the fuzzer

marcusklaas:
This addresses two, and adds a (disabled) test for a third one, which I hope to fix later.

### DIFF
--- a/src/firstpass.rs
+++ b/src/firstpass.rs
@@ -412,19 +412,22 @@ impl<'a, 'b> FirstPass<'a, 'b> {
                 self.extract_and_parse_heading_attribute_block(header_start, header_end);
             attrs = attrs_;
 
-            // remove trailing block attributes
-            let cur_ix = self
-                .tree
-                .truncate_siblings(self.text.as_bytes(), content_end);
+            // strip trailing whitespace
+            let new_end = if has_trailing_content {
+                content_end
+            } else {
+                let trailing_ws =
+                    scan_rev_while(&bytes[header_start..content_end], is_ascii_whitespace_no_nl);
+                content_end - trailing_ws
+            };
 
-            if let Some(cur_ix) = cur_ix {
-                // strip trailing whitespace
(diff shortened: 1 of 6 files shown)


5225225:
Is

```rust
#[test]
fn unwrap_none() {
    parse("*\t[][\n\t<p]>\n\t[]");
}
```

one of the panics addressed here (`thread 'unwrap_none' panicked at 'called `Option::unwrap()` on a `None` value', src/parse.rs:1083:46`)?

marcusklaas:
It appears it is not! That example also fails without any options set. Nice find!

src/firstpass.rs
```diff
@@ -457,276 +456,273 @@ impl<'a, 'b> FirstPass<'a, 'b> {
         let mut last_pipe_ix = start;
         let mut begin_text = start;
 
-        let (final_ix, brk) =
```
marcusklaas, line 460: This is unfortunately just `rustfmt` deciding that this entire expression needs to rewritten. Sorry for the noise! Although it does reduce the entire indentation level by one, which is nice. I'm just hoping it doesn't jump back on further modification...

marcusklaas:
Okay, this now fixes the last of the known panics. I did another quick 15 min fuzz which did not find anything.

The low hanging fruit is officially gone!

5225225:
Cool! ~~When~~ *If* I find any more issues, I'll add them as new issues.

raphlinus (approved):
I did a quick skim and don't see anything that concerns me. I think increasingly we want to rely on tests and fuzzing to make sure it's doing the right thing, it's hard to reason about the logic. If there's anything that you feel needs special scrutiny, lemme know and I can take a closer look.

marcusklaas:
Thanks for the review Raph!

I am so happy with the outcome of the fuzzer spree so far. It's not only some missing checks that were triggered, but real bugs that were just hard to find. The code has become better as a result :)

PR #565, opened 2022-01-12
Fix smart punctuation

GuillaumeGomez:
As reported on rustdoc: https://github.com/rust-lang/rust/issues/92578, the smart punctuation was incorrect in case the opening quote isn't followed right away by a letter. Based on the [commonmark spec](https://spec.commonmark.org/dingus/?text=Single%20space%20at%20start%3A%20%22%20like%20this%22.%0ASingle%20space%20at%20end%3A%20%22like%20this%20%22.%0ASpaces%20at%20start%20and%20end%3A%20%22%20like%20this%20%22.%0ANo%20spaces%3A%20%22like%20this%22.&smart=1), it's fine. Only the closing quote matters as it seems.

I didn't write a specialized version but I can if you prefer to handle double quotes on their own?

### DIFF
--- a/src/firstpass.rs
+++ b/src/firstpass.rs
@@ -1444,18 +1444,24 @@ fn surgerize_tight_list(tree: &mut Tree<Item>, list_ix: TreeIndex) {
 /// suffix is &s[ix..], which is passed in as an optimization, since taking
 /// a string subslice is O(n).
 fn delim_run_can_open(s: &str, suffix: &str, run_len: usize, ix: usize) -> bool {
-    let next_char = if let Some(c) = suffix.chars().nth(run_len) {
+    let delim = suffix.chars().next().unwrap();
+    let mut iter = suffix.chars().skip(run_len);
+    let next_char = if delim == '"' {
+        iter.skip_while(|c| c.is_whitespace()).next()
+    } else {
+        iter.next()
+    };
+    let next_char = if let Some(c) = next_char {
         c
     } else {
         return false;
     };
-    if next_char.is_whitespace() {
+    if delim != '"' && next_char.is_whitespace() {
         return false;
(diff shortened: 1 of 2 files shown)


GuillaumeGomez:
Interestingly enough, the behaviour is different with a simple quote. I'll add a test for them as well.

marcusklaas:
It's tricky to conflate the CommonMark spec and the CommonMark reference implementation. They are not the same thing.

Does the specification make any statements on leading spaces after quotes?

GuillaumeGomez:
I couldn't find something specific about this. If we refer to the rules described in [emphasis and strong emphasis](https://spec.commonmark.org/0.30/#emphasis-and-strong-emphasis), then the current implementation is correct. However, why is CommoMark reference implementation handling the double quotes differently than the rest?

raphlinus:
I would prefer some discussion within the CommonMark community to nail down what the expected semantics should be before making a change to the code. It seems like our existing behavior is one reasonable outcome.

GuillaumeGomez:
I'm fine with it either way. I'll let it up to you then. ;)

Do you want me to open a PR with the tests I added to enforce the current behaviour?

raphlinus:
I guess I was asking for a thread to be started on the CommonMark discussion board. I think it would be useful for the broader community to understand this is currently underspecified. Once we have some semblance of consensus then I'd happily welcome a fix (if needed) + a test to solidify the behavior.

vassudanagunta:
"Smart Quotes" behavior is not part of the [CommonMark spec](https://spec.commonmark.org/0.30/). It's just an optional added feature of the reference implementation. You can turn it off. If it's buggy, you should log a bug against the reference implementation, not the CommonMark spec.

Issue #567, opened 2022-01-16
Mixed nested unordered list types no longer work


ehuss:
The following list type no longer seems to render correctly:

```md
* foo
    + bar
    + baz
```

In 0.9.0 renders as:

```
<ul>
<li>foo
+ bar
+ baz</li>
</ul>
```

In 0.8.0 it was:

```
<ul>
<li>foo
<ul>
<li>bar</li>
<li>baz</li>
</ul>
</li>
</ul>
```

This seems to be caused by #554.

marcusklaas:
Oof, this is not good. I will look into this shortly.

marcusklaas:
Okay, found the problem. The PR you linked uncovered this bug, but didn't cause it. We checked for unordered list by inspecting the list marker. We compared it co `*` and `-`, but forgot about `+` lists! So luckily this had an easy fix :-)

PR #569, opened 2022-01-17
Fix bug where we didn't let + lists break paragraphs

marcusklaas:
We were also able to simplify the paragraph interruption checks into one function.

Fixes https://github.com/raphlinus/pulldown-cmark/issues/567.


camelid:
Thanks for the quick fix! Once this PR is merged, could you issue a patch release so we could fix it in rustdoc?

raphlinus (approved):
Yes, this looks right. It was definitely a mistake to leave out "+". I'm not 100% sure why just adding "+" to interrupt_paragraph_by_list wouldn't have fixed it, but agree that the new code is cleaner, less duplication of logic.

marcusklaas:
Thanks for the quick review! You're absolutely right that adding `+` to `interrupt_paragraph_by_list` would've done it, but while investigating I noticed that that needn't be a separate function anymore, since it doesn't use any context from `self`.

@ehuss Will do a quick release after this is merged!

PR #570, opened 2022-01-17
Bump version to 0.9.1

marcusklaas:
Getting ready to do a patch release.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -57,9 +57,9 @@ dependencies = [
 
 [[package]]
 name = "bumpalo"
-version = "3.8.0"
+version = "3.9.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


raphlinus (commented):
Update README.md as well?

marcusklaas:
I _always_ forget to update the README, but I think it may not be necessary in this case? It says `0.9`, which should pick up the patch release automatically, right?

raphlinus:
It does if you add it de novo, but won't update an existing lockfile. I personally consider it good practice to include the point, but agree it's not absolutely required.

marcusklaas:
> but won't update an existing lockfile
I did not know this! I will update the README as well.

raphlinus (approved):

marcusklaas:
@ehuss: we just released a [new version](https://crates.io/crates/pulldown-cmark/0.9.1) which includes a fix for the bug you reported!

ehuss:
Thanks! ❤️

PR #573, opened 2022-02-02
Address cargo's more pedantic suggestions

marcusklaas:


Martin1887:
This will be rechecked using clippy.

PR #582, opened 2022-04-24
Correct documentation for `TableHead`

AlexApps99:
`TableHead` does not contain any `TableRow`s.

This tripped me up while writing a renderer.


marcusklaas:
Good catch, thanks!

PR #583, opened 2022-04-24
Make InlineStr's len its own field

### LINKED ISSUE #581: InlineStr stores length inside string buffer
While not really a problem, it's weird to me to store length of a string inside the string buffer itself, especially without any comments about it in the code.

It shouldn't break anything to make it a separate field in struct (and decrease `MAX_INLINE_STR_LEN` by one), and it would make it more readable/obvious what it is doing.

```rust
const MAX_INLINE_STR_LEN: usize = 3 * std::mem::size_of::<isize>() - 2;

pub struct InlineStr {
    inner: [u8; MAX_INLINE_STR_LEN],
    len: u8,
}
```

If, for some reason, it's not possible, I suggest at least putting a comment explaining the behaviour and reasoning.

Code change would be what's in playground here: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=267627926e4a7d246e4b4e3ecb10483d and changing https://github.com/raphlinus/pulldown-cmark/blob/ea1b4cd4dbcf912129f14dd81ad4809f0d5d1999/src/strings.rs#L278-L280 to

```rust
// we need 4 bytes to store a char
assert!(MAX_INLINE_STR_LEN >= 4);
```

Shadlock0133:
Fixes #581

### DIFF
--- a/src/strings.rs
+++ b/src/strings.rs
@@ -5,7 +5,7 @@ use std::hash::{Hash, Hasher};
 use std::ops::Deref;
 use std::str::from_utf8;
 
-const MAX_INLINE_STR_LEN: usize = 3 * std::mem::size_of::<isize>() - 1;
+const MAX_INLINE_STR_LEN: usize = 3 * std::mem::size_of::<isize>() - 2;
 
 /// Returned when trying to convert a `&str` into a `InlineStr`
 /// but it fails because it doesn't fit.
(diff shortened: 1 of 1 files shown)


src/strings.rs
```diff
@@ -275,9 +276,8 @@ mod test_special_string {
 
     #[test]
     fn max_inline_str_len_atleast_five() {
```
marcusklaas, line 278: Nit: we should probably rename this test as well!

marcusklaas:
It seems the test `strings::test_special_string::inlinestr_fits_twentytwo` is failing. Once that has been addressed, this looks good to go!

marcusklaas:
Thank you!

Issue #585, opened 2022-05-08
Support for a C API?


gjtorikian:
Hi! I maintain Commonmarker, the Ruby bindings to the Commonmark C lib. For [a few years now](https://github.com/gjtorikian/commonmarker/issues/107 ) I’ve wanted to move off of the C lib and onto a Rust-generated dylib. 

Would you support a PR that exposed a C entry point (eg. `crate-type = ["staticlib", "cdylib", "rlib"]`) and a header, so that other FFIs can bind to this library?

marcusklaas:
That seems like a great idea to me!

PR #586, opened 2022-05-09
Implementation of C API

gjtorikian:
This PR is designed to provide C bindings with the end of goal of allowing higher level languages (eg., Ruby) to make use of this Rust library. The PR will support taking in a `char * data` and `int options` to convert Markdown into HTML.

I'm opening a very early draft of this PR to get 👍 approval for the overall direction. This isn't my project, so I'm just creating folders and adding dependencies willy-nilly; I hope you'll find them acceptable and not too outrageous. I need to support options parsing next, and add many more tests before I can call this complete.

You can `cd` into `c-api/tests` and run `cargo run` to build/run the C test suite. Ideally, all a consumer of this C library would need is the lib (generated by `cargo build` in  the `c-api` folder) and the header file, which exposes the one and only available method.

Closes https://github.com/raphlinus/pulldown-cmark/issues/585.


gjtorikian:
@marcusklaas Sorry for the annoying ping! I don’t know if draft PRs emit notifications. Just looking for general approval here before I push forward. Thank you! 🙂

marcusklaas:
Hey, no worries about the ping. I had totally lost track of this PR, so it's good that you reminded me.

I have had a quick peek at the code, and thought a bit about the design. Here's my two considerations:
- it seems that the C adaptation can be built cleanly on-top of pulldown, without requiring any changes to the internal code. That's excellent, as it nicely separates concerns. Further, it should also allow us to take this C interface into a separate crate, right? I am thinking that that may be a good place to start, at least for now. One reason being my second point:
- the C interface only supports a markdown string -> HTML string operation right now, and while that's plenty for many use cases, I think the true charm of pulldown_cmark over other markdown parsers is exactly its pull-based design. Meaning that we emit events that can be used for more than straight HTML formatting (or manipulate them before producing HTML). I fully acknowledge that this probably _way_ more difficult in C; I certainly wouldn't know how to do it. To come back to my first point, this may be why it may be better to gestate this feature in a separate crate, to experiment with designs that work for C without committing to one right away, if us maintainers would even be qualified to make such a call.

I realize this is a change in position from my initial "Yes, let's do it!" response. I hadn't considered it carefully enough. I do think it would be awesome if pulldown could be used in languages like Ruby through C, so I do hope you continue work. And I'm certainly open to discussing including a C interface it in this repository, but it does feel a little premature right now. My apologies for the mixed messaging!

gjtorikian:
> * That's excellent, as it nicely separates concerns. Further, it should also allow us to take this C interface into a separate crate, right?

Yep, this is a pattern I've seen other projects use and it just makes sense. I'd need you to obviously set up CI and the cargo deploy process since I don't have access but otherwise this should be completely self-contained.

> Meaning that we emit events that can be used for more than straight HTML formatting (or manipulate them before producing HTML). I fully acknowledge that this probably way more difficult in C; I certainly wouldn't know how to do it. 

I can defintely explore this before saying "I can't do this!" 😆 I'm pretty sure libxml2 is pull-based so it might serve as a useful pattern to explore. I think I've also seen other Rust-backed C libraries around HTML manipulation that use pull parsing. 

More broadly, though, I would not like perfect to be the enemy of good, and so if I can capture what exactly the C API needs to do to support the pull parsing—or start an incomplete version, with comments and TODOs—it should be a nice comprimise between satisfying the "just convert a Markdown file" use case versus the more "advanced" pull handling, which can be chipped away at over the next few weeks. Mostly, I'm worried about my lack of knowlede of the Rust lib. I see [the `pub` events exposed here by the lib](https://github.com/raphlinus/pulldown-cmark/blob/f29bd1e228913690e5092c9594e4e607423ff0aa/src/lib.rs#L228), and I'm  confident I can easily cover those broad events, but I don't know if there is a long tail of minor details that also need to be covered. What do you think?

gjtorikian:
@marcusklaas I’m not entirely sure I have the time to fully support the pull event stream in C within this PR. Please let me know if you’d like me to go forward with supporting just the commonmark -> HTML processing here or whether the PR should just be closed.

PR #587, opened 2022-05-20
Do not %-encode `^` and `~` characters

GuillaumeGomez:
@lilyball gave a very nice explanation on rust repository about this. You can see it [here](https://github.com/rust-lang/rust/issues/97125).


marcusklaas:
This looks sensible to me. Thanks!

GuillaumeGomez:
Thanks for the quick merge! If you could make a release with this fix whenever you can, it'd be super appreciated too. :)

GuillaumeGomez:
Sorry to ask but do you know when will be the next release? It would allow us to fix this bug: https://github.com/rust-lang/rust/issues/97125. Thanks in advance!

marcusklaas:
We just released [v0.9.2](https://crates.io/crates/pulldown-cmark/0.9.2) that includes your fix.

GuillaumeGomez:
Thanks a lot!

PR #588, opened 2022-05-22
Use `Write::write_all` instead of `Write::write` in build.rs

lo48576:
Fix [`clippy::unused_io_amount`](https://rust-lang.github.io/rust-clippy/master/#unused_io_amount) lint.
This lint is deny-by-default in Rust 1.61.0 and `cargo clippy --all-features` treats this as hard error.


marcusklaas:
Thanks for this!

PR #599, opened 2022-07-03
Rename the erroneously called `SLASH_ESCAPE` to `SINGLE_QUOTE_ESCAPE`

BenjaminRi:
The change is trivial - a simple renaming of a variable. The code already worked correctly despite the wrong name.

The erroneous name came in with the following commit which refactored string literals into variables: https://github.com/raphlinus/pulldown-cmark/commit/9238b70faa77331bd861ce79033fba7c96ff59fd


raphlinus (approved):
Good catch, thanks!

PR #603, opened 2022-07-29
Add "no-backline" feature

GuillaumeGomez:
This is part of an effort to reduce the number of unneeded characters in the generated HTML. So using this feature when generating the `std` gave this result:

```
Using `du -s doc/std/`:

before: 105860
after: 105812

Using `wc -c std/index.html`:

before: 62859
after: 62631
```

So surprisingly, the number of characters seems to be quite high on a given file, but the impact on the documentation seems quite small.

So what do you think of this feature? Would you be interested into it? It diverges from the commonmark spec, hence why I suggest to add it as a feature and not a parameter.

### DIFF
--- a/src/html.rs
+++ b/src/html.rs
@@ -28,6 +28,11 @@ use crate::strings::CowStr;
 use crate::Event::*;
 use crate::{Alignment, CodeBlockKind, Event, LinkType, Tag};
 
+#[cfg(feature = "no-backline")]
+const BACKLINE: &str = "";
+#[cfg(not(feature = "no-backline"))]
+const BACKLINE: &str = "\n";
+
 enum TableState {
     Head,
     Body,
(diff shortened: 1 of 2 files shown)


src/html.rs
```diff
@@ -108,13 +113,13 @@ where
                     self.write_newline()?;
                 }
                 HardBreak => {
-                    self.write("<br />\n")?;
+                    self.write(&(String::from("<br />") + BACKLINE))?;
```
raphlinus, line 116: Are these allocating? If so, that seems like a performance hit.

GuillaumeGomez: It's const string cancatenation so it should have no impact at runtime (to be confirmed though!).

GuillaumeGomez: Ah maybe not. I think `String::from` isn't const.

GuillaumeGomez: Originally I wanted to use `if cfg!(feature = "no-backline")` but found it too noisy. Well in any case, it's another possibility. But the question here is really if you're interested by this feature or not. If you are, I need to add tests too. :)

marcusklaas:
I am not sure whether an output size savings of 0.05% to 0.3% is worth the extra complexity this would introduce, especially when we would have to use a bunch of `cfg!(feature = "no-backline")` checks to prevent performance regressions.

src/html.rs
GuillaumeGomez: Possibly. But considering the small impact, I think it's not worth it currently. Maybe for later on.

GuillaumeGomez:
Considering the code it adds compared to the result, I don't think it's worth it currently so closing it.

PR #611, opened 2022-09-07
Implement `Debug` for all public types

### LINKED ISSUE #609: `BrokenLink` does not implemnet `Debug`
All fields of `BrokenLink` implements `Debug`, so `BrokenLink` can implement `Debug`.

Implementing `Debug` is useful for debugging.

gifnksm:
This pull request implements `Debug` for all public types detected by `missing-debug-implementations` lint.

This also enables the lint checking by default so that similar problems can be detected in the future.

fixes #609

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -123,6 +123,7 @@ impl<'a> Default for ItemBody {
     }
 }
 
+#[derive(Debug)]
 pub struct BrokenLink<'a> {
     pub span: std::ops::Range<usize>,
     pub link_type: LinkType,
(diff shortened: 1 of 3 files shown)


src/parse.rs
```diff
@@ -143,6 +144,24 @@ pub struct Parser<'input, 'callback> {
     link_stack: LinkStack,
 }
 
+impl<'input, 'callback> std::fmt::Debug for Parser<'input, 'callback> {
```
marcusklaas, line 147: Note that users of the library cannot access many of the internals of the parser that you're exposing here, so I'm not sure how useful adding them is. I'd limit the output to fields that have a public type. That also cuts down on the number of debug implementations we have to add, reducing library size (a little).

marcusklaas:
Thanks for the PR. I left one comment on the code. Once that's addressed, this is good to go.

src/parse.rs
gifnksm: I've added the commit to stop exposing the fields that have private types in Debug implementation

gifnksm:
Thank you for your review. I have addressed your comments.

marcusklaas:
Merged. Thanks for your contribution!

PR #614, opened 2022-10-08
Allow dynamically generated header ids and classes

SergioBenitez:
Changes the second (fragment identifier) and third (classes) fields of `Tag::Heading` to be of type `CowStr` from the previous `&str`, allowing dynamically generated values to be used in their place.

In my use of `pulldown-cmark`, I want to automatically generate heading IDs and have the markdown renderer emit them. This change makes that possible. Workarounds are possible, but maintaining the tree as close to the original source as possible is desired.


Martin1887:
This pull request must be updated because the new custom attributes field of headings.

Could you create a new commit fixing this, @SergioBenitez? I you cannot, I will create another pull request handling this or, if you grant me write permissions to your branch, I could update it.

Anyway, benchmarks are needed before merging the pull request to check that the performance impact is not excessive.

Thanks.

Martin1887:
Performance is a bit lower in some benches, but consistency is important and the performance hurts are also reduced in use cases where headings ids, classes and attrs must be modified, so this can be merged.

PR #617, opened 2022-11-11
Remove unnecessary default feature `getopts`

rhysd:
`getopts` crate is only used for executable, but added to default features. This can reduce number of dependencies when this crate is installed as lib.


rhysd:
Sorry, this requires additional arguments on `cargo install` so it might not be good for bin users.

PR #622, opened 2022-11-19
Support math extension

rhysd:
This PR adds support for [mathematical expressions](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/writing-mathematical-expressions) behind `Options::ENABLE_MATH` flag, which was introduced to GitHub recently.

This extension supports both inline-level and block-level expressions:

```markdown
Inline-level:  $\sqrt{3x-1}+(1+x)^2$

Block-level:

$$\left( \sum_{k=1}^n a_k b_k \right)^2 \leq \left( \sum_{k=1}^n a_k^2 \right) \left( \sum_{k=1}^n b_k^2 \right)$$
```

They are rendered as follows on GitHub:

Inline-level:  $\sqrt{3x-1}+(1+x)^2$

Block-level:

$$\left( \sum_{k=1}^n a_k b_k \right)^2 \leq \left( \sum_{k=1}^n a_k^2 \right) \left( \sum_{k=1}^n b_k^2 \right)$$


src/parse.rs
```diff
@@ -530,6 +533,43 @@ impl<'input, 'callback> Parser<'input, 'callback> {
...
+                            prev = cur;
+                            cur = self.tree[cur_ix].next;
+                            continue;
+                        };
+
+                    let node = scan_nodes_to_ix(&self.tree, next, end_ix + 1);
+                    if end_ix > start_ix {
+                        let cow_ix = self
+                            .allocs
+                            .allocate_cow(block_text[start_ix + 1..end_ix].into());
```
mattwidmann, line 561:
```suggestion
                            .allocate_cow(block_text[start_ix..end_ix].into());
```

Off-by-one? The first character of inline math seems to get dropped without this suggestion.

src/parse.rs
```diff
@@ -530,6 +533,43 @@ impl<'input, 'callback> Parser<'input, 'callback> {
...
+                        self.tree[cur_ix].item.body = ItemBody::Math(MathDisplay::Inline, cow_ix);
+                        self.tree[cur_ix].item.start = self.tree[cur_ix].item.start + 1;
```
mattwidmann, line 563:
```suggestion
                        self.tree[cur_ix].item.start = self.tree[cur_ix].item.start;
```

specs/math.txt
```diff
@@ -0,0 +1,58 @@
...
+See the following link to know the syntax.
+https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/writing-mathematical-expressions
+
+Inline-level mathematical expressions:
+
+```````````````````````````````` example
+This sentence uses `$` delimiters to show math inline:  $\sqrt{3x-1}+(1+x)^2$
+$\sum_{k=1}^n a_k b_k$: Mathematical expression at head of line
+.
+<p>This sentence uses <code>$</code> delimiters to show math inline:  <span class="math">sqrt{3x-1}+(1+x)^2</span>
```
mattwidmann, line 13: These examples show the issue where the first character of inline math is dropped.

rhysd:
Thanks! I'll take a look at review comments tomorrow.

rhysd:
@mattwidmann Yes, you're totally correct. I fixed the off-by-one error and test cases at 7123b76.

Schabolon:
Hi,
I hope this is the right place to add this information.
The math-option is very useful to me.
After using it for some time, I think I have discovered a bug.
The following Markdown: `$\{1, 2, 3\}$`
Gets converted to this HTML: `<div class="math inline">{1, 2, 3\}</div>`
The backslash before the first curly-bracket went missing.
This problem doesn't seem to occur on math blocks (they work fine).
I tried to find the problem, but couldn't figure it out (my only guess is, that it gets removed because all `\{` gets converted to `{` in HTML)
Would be amazing if this bug could be fixed.

rhysd:
@Schabolon Thanks. I fixed it at 7704ff0.

rhysd:
@Schabolon Thanks. I fixed it at cbcdf26.

```
> echo '$\{1, \{2, \{3\}\}\}$' | cargo run -- -M -e
    Finished dev [unoptimized + debuginfo] target(s) in 0.07s
     Running `target/debug/pulldown-cmark -M -e`
0..22: Start(Paragraph)
1..20: Math(Inline, Borrowed("\\{1, \\{2, \\{3\\}\\}\\}"))
0..22: End(Paragraph)
EOF
```

Schabolon:
Thanks a lot for the quick fix 👍

xelatihy:
Any chance this will be accepted soon? Math support is a needed feature for a project I am working on and I will soon have to choose a parser. Thanks.

Carlosted:
I would love for this PR to go through, as my project needs math expressions, Thanks!

ehuss:
Can you say more about how the rendering is expected to work? Does this require MathJax? Is there some guidance on how to integrate it properly? Does this work with katex?

rhysd:
@ehuss This PR does not render the math expressions. It just parses the mathematical expressions notation at Markdown level. So it parses `$...$` and `$$...$$` math blocks, but does not touch to their contents.

https://github.com/rhysd/pulldown-cmark/blob/61e4922f6bc9023e49fa220ff2a673e969ff3f9e/src/html.rs#L135-L144

The `text` arguments in the match arms are raw math block contents. How they are rendered is left to the user of this library.

> Does this require MathJax?

No. But some math expression renderer is at least necessary. However I recommend to use MathJax because GitHub uses MathJax for rendering math expressions in Markdown.

> Does this work with katex?

Yes.

> Is there some guidance on how to integrate it properly?

No. But I can explain it for you. It is almost the same as raw HTML. Regarding to the Markdown to HTML conversion provided by pulldown-cmark, currently this PR converts

```markdown
$\sqrt{3x-1}+(1+x)^2$
```

to

```html
<span class="math inline">\sqrt{3x-1}+(1+x)^2</span>
```

To render this HTML element, you would load the element and MathJax library in the same page on your browser and call some MathJax's API to render the element's content as math expression looking at [MathJax document](https://docs.mathjax.org/en/latest/index.html).

I think trying it on your side would be a good way to understand how it works.

```sh
git clone https://github.com/rhysd/pulldown-cmark.git -b math
cd pulldown-cmark
echo '$\sqrt{3x-1}+(1+x)^2$' > test.md
cargo run -- -M < test.md # See the converted HTML from the Markdown file
cargo run -- -M -e < test.md # See the parsed events from the Markdown file
```

ehuss:
> I don't know your use case. If you explain it, I may explain the usage better.

I'm the maintainer of [mdBook](https://github.com/rust-lang/mdBook/). It currently has MathJax support, which if I understand correctly, scans the entire page for delimiters and renders them. 

mdBook also supports plugins, such as [mdbook-katex](https://github.com/lzanini/mdbook-katex).

I assume the largest benefit here is to avoid problems with escaping delimiters?

What would the mathjax init code look like for basic usage?

rhysd:
Does mdBook currently scans math blocks separately from parsing Markdown? Then it would be great if parsing everything by Markdown parser.

> What would the mathjax init code look like for basic usage?

Since I'm not a heavy MathJax user, I don't know what is basic. In my app, I'm using MathJax with React and it would not be basic.

Instead, could you refer the official demo examples?

- On browser: https://github.com/mathjax/MathJax-demos-web
  - Working example: https://mathjax.github.io/MathJax-demos-web/
- On Node.js: https://github.com/mathjax/MathJax-demos-node

Since I don't know how mdBook processes converted HTML, I can't say which example is best for you. However, [input-tex2svg.html](https://github.com/mathjax/MathJax-demos-web/blob/master/input-tex2svg.html) is converting TeX input into SVG element in JavaScript and it might fit to your use case.

ehuss:
> Does mdBook currently scan math blocks separately from parsing Markdown? Then it would be great if parsing everything by Markdown parser.

The math scanning is all done by MathJax. There is just a little Javascript that handles everything:

```html
        <script>
        window.addEventListener('load', function() {
            MathJax.Hub.Register.StartupHook('End', function() {
                window.setTimeout(window.print, 100);
            });
        });
        </script>
```

Over at https://docs.mathjax.org/en/latest/web/configuration.html, I don't see an option to process only spans with a given class. That's what I'm looking for: the javascript code that would be compatible with this change.

rhysd:
I think what you're looking for is input options: https://docs.mathjax.org/en/latest/options/document.html#document-options

Please be careful that you're still using MathJax v2 and the document I shared is for v3.

ehuss:
Ah, thanks! Yea, we will be adding v3 support soon, but have to support both for a while. Looks like it is pretty easy to point it at a class, and to add `$` processing.

I'm testing this out a little bit, and it seems to be more eager than GitHub to convert spans to math. For example:

```
That costs between $100 and $200.
```

On GitHub renders as expected (without math):

That costs between $100 and $200.

But this PR converts it to a span:

```html
<p>That costs between <span class="math inline">100 and </span>200.</p>
```

ehuss:
Or taking the example from the GitHub documentation of using a span seems to get mangled:

```
To split <span>$</span>100 in half, we calculate $100/2$
```

becomes:

```html
<p>To split <span><span class="math inline">&lt;/span&gt;100 in half, we calculate </span>100/2$</p>
```

I think I would expect it to be:

```html
<p>To split <span>$</span>100 in half, we calculate <span class="math inline">100/2</span></p>
```

ehuss:
Hm, I'm trying to test MathJax 3, and I can't see a way to get it to work with this PR since this PR strips the delimiters. Do I have to override `FindTeX`?  Are you using MathJax for your use case? If not, what are you using?

I think it would be helpful to have a rudimentary example of how to integrate this change with MathJax.

systemsoverload:
Is there anything I could do to help get this PR across the line? I'd love to be able to use this feature.

SichangHe:
> Ah, thanks! Yea, we will be adding v3 support soon, but have to support both for a while. Looks like it is pretty easy to point it at a class, and to add `$` processing.
> 
> I'm testing this out a little bit, and it seems to be more eager than GitHub to convert spans to math. For example:
> 
> ```
> That costs between $100 and $200.
> ```
> 
> On GitHub renders as expected (without math):
> 
> That costs between $100 and $200.
> 
> But this PR converts it to a span:
> 
> ```
> <p>That costs between <span class="math inline">100 and </span>200.</p>
> ```

This is largely a choice issue. There are two choices that I have seen:

1. Treat whatever that is between `$…$` as an inline math block.
  LaTeX does this. `mdbook-katex` does this. Currently, this PR seems to do this as well.
  Slightly easier to implement and maintain.
  Users are forced to always escape dollar signs `\$` if they want dollar signs. They might find it unnatural to have to do `\$200`.
  I think this is totally fine except that you need to tell the users front and center when they enable the flag.

2. Only accept what is between `$…$` as an inline math block if no whitespaces are inside and right next to any of the dollar signs.
  This means expressions like `$ x^2 $` are not treated as math expressions. So, normal use of `$` to indicate an amount of money are usually ignored, which is handy.
  GitHub does this, the popular Markdown-it math plugin does it.
  Downsides: LaTeX is incompatible with this, that is, if you copy math expressions typed in LaTeX to a Markdown file and use these parsers, they may not render them. Migration from LaTeX would be painful.

getreu:
I am interested in this feature. What is the state of acceptance and implementation?

Martin1887:
This is also an interesting feature for us, but it requires more time to define a specification and to implement it, so it is not scheduled in the short term but in the middle term.

There are several pull requests doing it and probably the final implementation will be a mix of them and other things.

rambip:
I found another bug:
When you want to parse this markdown:
```md
math
$$x$$
```

This is not interpreted as math.
I tested it with the latest version:
```sh
echo 'math \n $$x$$' | cargo run -- -m -e
    Finished dev [unoptimized + debuginfo] target(s) in 0.03s
     Running `target/debug/pulldown-cmark -m -e`
0..14: Start(Paragraph)
0..8: Text(Borrowed("math \\n "))
8..10: Text(Borrowed("$$"))
10..11: Text(Borrowed("x"))
11..13: Text(Borrowed("$$"))
0..14: End(Paragraph)
EOF
```

MiSawa:
Maybe it's just out-of-scope for the PR, but I'm wondering if you're willing to support <code>$\`...\`$</code> / <code>\`\`\`math</code> as well.

rhysd:
- 1ae5351 added support for ``$`...`$`` notation
- 8eb58af fixed `$...$` math expression cannot start/end with spaces (but ``$`...`$`` can)

MiSawa:
Ah awesome, thank you!

rhysd:
Note that this PR branch is for HEAD of the master branch. If someone needs this feature with pulldown-cmark v0.9.3, [math-0.9.3 branch](https://github.com/rhysd/pulldown-cmark/tree/math-0.9.3) is available.

specs/math.txt
```diff
@@ -0,0 +1,127 @@
...
+Sole `$` character is handled as normal text. And Inline-level mathematical expressions cannot continue across newlines:
+
+```````````````````````````````` example
+Hello $ world.
+
+Dollar at end of line: $
+.
+<p>Hello $ world.</p>
+<p>Dollar at end of line: $</p>
+````````````````````````````````
```
ollpu, line 54: Great to see more activity around this long-standing issue! I'm willing to work on this if any help is needed.

On this part, I'm not sure I agree. In LaTeX and markdown files alike, a common practice is to justify paragraphs to a width of e.g. 80 characters. Single newlines do not affect the output, even inside math expressions.

There seems to be a divergence of behavior here between markdown files and issues/comments on GitHub. Comments preserve newlines from the raw text, but in markdown files this is not the case. This markup:

```
text and $math broken
into multiple lines$
```

renders in this comment as

text and $math broken
into multiple lines$

but is shown as math in a Gist: https://gist.github.com/ollpu/2b0e18b8fe20d76b97d9e89912f4df3e

It also works with [Pandoc](https://pandoc.org/try/).

Since with pulldown-cmark single newlines are generally ignored, it would make sense to ignore them inside inline math as well.

There is subtlety with how newlines can start/end a math expression though. GitHub and Pandoc have different behavior: On GitHub, an inline math expression cannot start with a newline right after the `$`, but it can end with one before the closing `$`. (See the [same Gist](https://gist.github.com/ollpu/2b0e18b8fe20d76b97d9e89912f4df3e) for examples of this) Pandoc is more consistent, and inline math can neither start or end with a newline – they are essentially treated the same as spaces for this purpose.

Escaped newlines (newline preceded by `\`) are another can of worms.

For what it's worth, I prefer following Pandoc over GitHub in this and other similar cases since the implementation is more mature, consistent and doesn't suffer from the robustness issues that GitHub did at least originally.

SichangHe: Similarly, I support following $\LaTeX$'s standards. They are around for a good while and solid.

ollpu: Another common thing supported by both GitHub are display-mode math blocks inside a paragraph.

```
display $$\LaTeX$$ math
```

renders as

display $$\LaTeX$$ math

rhysd: - `$` doesn't continue across lines. In this test case, there is no math expression.
- This PR refers GitHub's markdown renderer behavior. It may differs from LaTeX behavior but it is intended
- I know several edge cases are existing but I intentionally didn't implement it since they make implementation more complex and harder to maintain

ollpu: I'm eyeing a refactor to this PR branch to make math parsing kind of combine the logic of `ItemBody::MaybeEmphasis` and `ItemBody::MaybeCode`. #477 does something similar for inline math, but not display blocks. It seems like display math should behave a little closer to display math and inline code, but just not do the whitespace checks.

What do you think @rhysd?

ollpu: I went ahead and implemented that. More details in the PR: https://github.com/rhysd/pulldown-cmark/pull/1

I'd rather not have a third competing PR to upstream, so hopefully we can converge on a good design here.

rambip: > I found another bug: When you want to parse this markdown:
> 
> ```
> math
> $$x$$
> ```
> 
> This is not interpreted as math. I tested it with the latest version:
> 
> ```shell
> $ echo -e 'math \n$$x$$' | cargo run -- -m -e
>     Finished dev [unoptimized + debuginfo] target(s) in 0.04s
>      Running `target/debug/pulldown-cmark -m -e`
> 0..13: Start(Paragraph)
> 0..5: Text(Borrowed("math "))
> 5..6: SoftBreak
> 7..9: Text(Borrowed("$$"))
> 9..10: Text(Borrowed("x"))
> 10..12: Text(Borrowed("$$"))
> 0..13: End(Paragraph)
> EOF
> ```

I don't know if anybody saw this, but it still doesn't work.
I'm pretty sure it is not the default behaviour for most math-in-markdown implementations.

ollpu: @rambip 

Yeah, to be clear, I made a separate PR onto @rhysd's branch, but that is not incorporated here yet.

This case works with [my version](https://github.com/ollpu/pulldown-cmark/tree/alt-math) there though:

```bash
$ echo -e 'math \n$$x$$' | cargo run -- -m -e
   Compiling pulldown-cmark v0.9.2 (/home/ollpu/git/pulldown-cmark)
    Finished dev [unoptimized + debuginfo] target(s) in 3.77s
     Running `target/debug/pulldown-cmark -m -e`
0..12: Start(Paragraph)
0..5: Text(Borrowed("math "))
5..6: SoftBreak
6..11: Math(Display, Borrowed("x"))
0..12: End(Paragraph)
EOF
```

rambip: Ok nice !

rhysd: I'm sorry for the delay. I didn't have enough time to see each PRs I've created this week.

> There seems to be a divergence of behavior here between markdown files and issues/comments on GitHub.

I didn't know this. Thank you for your PR. I'll take a look.

rambip:
Is there additional work that need to be done before merging ?
I tested the @ollpu branch in some of my projects and didn't notice any bug; the only undocumented situation may be multi-line math with empty breaks, like
```md
$$
x=

1
$$
```

I would be really happy to help, 0.10 needs math IMO

ollpu:
@rambip That's good to note. I added such a case to the math spec in my branch.

Martin1887:
> Is there additional work that need to be done before merging ? I tested the @ollpu branch in some of my projects and didn't notice any bug; the only undocumented situation may be multi-line math with empty breaks, like
> 
> ```
> $$
> x=
> 
> 1
> $$
> ```
> 
> I would be really happy to help, 0.10 needs math IMO

Math is an important feature but including it requires a big analysis of the options in syntax, features and implementations. Coordination with other projects using pulldown-cmark and supporting math like mdBook would also be desirable.

Therefore, math support will not be included in the 0.10 version but it is a priority for the 0.11 version.

rambip:
> > Is there additional work that need to be done before merging ? I tested the @ollpu branch in some of my projects and didn't notice any bug; the only undocumented situation may be multi-line math with empty breaks, like
> > ```
> > $$
> > x=
> > 
> > 1
> > $$
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I would be really happy to help, 0.10 needs math IMO
> 
> Math is an important feature but including it requires a big analysis of the options in syntax, features and implementations. Coordination with other projects using pulldown-cmark and supporting math like mdBook would also be desirable.
> 
> Therefore, math support will not be included in the 0.10 version but it is a priority for the 0.11 version.

I can understand that but "a big analysis" seems pretty vague to me.
If there is any spec that have to be written or even a ROADMAP, I can try.

Martin1887:
There is no spec or roadmap yet, they are the first things to do and a survey of the possible options would be the first step for the spec.

The survey is a 'help-wanted' thing and a different pull request or discussion could be opened to review the possible options and discuss them.

Thanks!

rambip:
I was starting to write a very early spec to have a starting point in a github pull request, but I just realized @ollpu had a decent one in his own pull request.
Where should the conversation continue then ?

I have listed some ambiguous cases where a survey could be usefull, but I don't know what to do with it.
Maybe @ollpu could other another pull request called "math spec" ?

ollpu:
Yeah, I've gotten myself familiar with the various approaches as well and the spec on my branch might be a good starting point. I've been busy the last week so I didn't have time to respond yet.

I opened a PR to discuss the spec: #733

rhysd:
Should we continue the discussion at #733? Or should I merge @ollpu's PR in this branch and continue the discussion here? Which one is fine to me.

rhysd:
Closing this in favor of #734. Thank you @notriddle for pushing forward the feature.

PR #628, opened 2023-02-03
Add classes field on table and on tr tag

qiuzhiqian:


Martin1887:
Thanks for your contribution.

Am I missing something or this pull request does not parse the classes and always write an empty set of classes?

In which spec it is based to specify classes in the Markdown files?

If classes are not being parsed I don't see the usefulness of this pull request, can you elaborate? Thanks.

Issue #584, opened 2022-05-02
Support for audio and video media types


JShorthouse:
It would be nice if this library had an option to enable html embedding of audio and video.

As an example, pandoc already supports this using the image embed syntax.

Example file:

```markdown
![](image.png)

![](video.mp4)

![](audio.wav)
```
Pandoc:

```html
$ pandoc test.md --to html5
<p><img src="image.png" /></p>
<p><video src="video.mp4" controls=""><a
href="video.mp4">Video</a></video></p>
<p><audio src="audio.wav" controls=""><a
href="audio.wav">Audio</a></audio></p>
```

Currently, for the same input, pulldown-cmark outputs:

```html
<p><img src="image.png" alt="" /></p>
<p><img src="video.mp4" alt="" /></p>
<p><img src="audio.wav" alt="" /></p>
```

---

For anyone else with this issue: as workaround for now, you can just put `<video>` tags directly in the markdown, and pulldown-cmark will leave them there in the generated html.

ultrasaurus:
I have a similar use case, but the workaround doesn't work for me.  I'm working on a static site generator and want the user to be able to write pure markdown, but then generate a/v players in the HTML when there are a/v links -- where linking to a file vs. inline playback could be disambiguated with the text of the link (or lack thereof, like pandoc).

I would like a way to override the writing of specific tags, maybe extending `HtmlWriter` in some way.  This could allow people to address issues like https://github.com/raphlinus/pulldown-cmark/issues/524, https://github.com/raphlinus/pulldown-cmark/pull/628, and maybe implement some CommonMark extensions (that are consistent with standard CommonMark syntax) without changes to the core code.

ppearson:
JFYI: I've hacked around this use-case by manually filtering the events after parsing, and replacing Event::Start(Tag::Image)  with raw html depending on the filename of the image URL with a Event::Html (and dropping the corresponding Event::End as well), .

It's slightly annoying and hacky, but works for my purposes.

ultrasaurus:
Now there are two PRs with two different approaches to addressing issue.  Interested in what people think!

Martin1887:
Pull requests will be reviewed soon, but this is probably out of scope of this project and should be implemented in another crate.

bobhy:
test task list
+ [ ] ddd
+ [x] ooo

PR #629, opened 2023-02-05
make easier for developers to produce alternate HTML for specific events/tags

ultrasaurus:
This is a possible approach to addressing https://github.com/raphlinus/pulldown-cmark/issues/584
It could also allow for developers to experiment with CommonMark extension that conform to standard syntax.  

It involves making `HTMLWriter` public and refactoring to create a public `render_event` function.  I'm curious if the project maintainers see this as something that the library could support, and if so, whether this approach is reasonable.  

Happy to work on it further, with feedback. Or please let me know if this is *not* something you ever want to pursue, in which case, I'll consider a different library or possibly maintain my own fork for my use case.


Martin1887:
This can be used as inspiration for a full-featured external crate for the HTML rendering, but is out of scope in pulldown-cmark.

PR #631, opened 2023-02-12
example generating alternate HTML for a particular tag

ultrasaurus:
Example of how to address https://github.com/raphlinus/pulldown-cmark/issues/584
using approach described by @ppearson

alternate to https://github.com/raphlinus/pulldown-cmark/pull/629


Martin1887:
Again, this pull request can be useful for inspiration to create another full-featured HTML renderer crate, but it is out of scope of this project.

Thanks for your contribution anyway.

PR #632, opened 2023-02-23
add option to parse single * as strong like in slack

marianoguerra:
In slack single `*` mean `<strong>`, see https://slack.com/help/articles/202288908-Format-your-messages#markup

this PD adds an option that when set will emit single `*` emphasis as strong.


Martin1887:
Sorry, but this pull request is very out of the scope of the project, since it is conflicting with CommonMark and Slack is not even a Markdown flavor.

PR #638, opened 2023-04-16
feat!: allow custom headings attributes with optional value. Fixes #634

### LINKED ISSUE #634: Support custom heading attributes (`attr` and `attr=value`)
Heading attributes are not limited to classes and ids but any type of attributes.

I think that supporting the Pandoc definition would be a good improvement: https://pandoc.org/MANUAL.html#extension-header_attributes.

If this feature seems desirable but maintainers cannot work on it, please reach me to work in a pull request.

Thanks.

Martin1887:
This feature is a breaking change because it modifies the `Heading` variant of the `Tag` enum with named fields and the new `attrs` field.

Fixes: #634

### DIFF
--- a/examples/parser-map-tag-print.rs
+++ b/examples/parser-map-tag-print.rs
@@ -19,7 +19,7 @@ fn main() {
         "```\n",
         "my code block\n",
         "```\n",
-        "\n",       
+        "\n",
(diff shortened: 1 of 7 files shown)


raphlinus (approved):
Thanks, looks good!

PR #643, opened 2023-05-03
Fix parser bug where link label gets broken by ] in code span

### LINKED ISSUE #642: ] inside inline code inside link text gets parsed incorrectly if corresponding reference exists
For this input:
```markdown
[`]: xx:

[`]`]
```

It should be parsed like this, since there is no link reference `` `]` ``
```html
<p>[<code>]</code>]</p>
```

But it's currently parsed as a link:
```html
<p><a href="xx:"><code>]</code></a></p>
```

From the [commonmark spec](https://spec.commonmark.org/0.30/#example-341):
> Code span backticks have higher precedence than any other inline constructs except HTML tags and autolinks.

---

Additionally, the `OffsetIter`'s range of this link is only the front part up to the `]`, i.e. the this part:
```text
[`]`]
^^^
```

This is currently causing rustdoc nightly to panic in some cases, see https://github.com/rust-lang/rust/issues/111117.

<details>
<summary>demo program that shows this behavior</summary>

```rust
use pulldown_cmark::{Parser, html};

/// [`]: xx:
///
/// [`]`]
fn main() {
    let text = "\
        [`]: xx:\n\
        \n\
        [`]`]\n\
    ";

    for (event, span) in Parser::new(text).into_offset_iter() {
        println!("{:?}: {:?}", event, &text[span]);
    }

    let mut parsed = String::new();
    html::push_html(&mut parsed, Parser::new(text));
    println!("{parsed}");
}
```

output:
```text
Start(Paragraph): "[`]`]\n"
Start(Link(Shortcut, Borrowed("xx:"), Borrowed(""))): "[`]"
Code(Borrowed("]")): "`]`"
End(Link(Shortcut, Borrowed("xx:"), Borrowed(""))): "[`]"
End(Paragraph): "[`]`]\n"
<p><a href="xx:"><code>]</code></a></p>
```

This shouldn't be parsed as a link in the first place, but even if it is, the range should be `` "[`]`]" ``, not `` "[`]" ``.

If you run `cargo +nightly doc` it crashes.

</details>

lukas-code:
```text
 [before `]` after]
 ^        ^       ^
 |        |       |
(A)      (B)     (C)
```

When the parser encounters the `]` at (C), it tries to parse a [link label](https://spec.commonmark.org/0.30/#link-label) starting from the previous `[` at (A) until it encounters the first `]` at (B). This is fine, because
> A link label begins with a left bracket (`[`) and ends with the first right bracket (`]`) that is not backslash-escaped.

However, due to
> Code span backticks have higher precedence than any other inline constructs except HTML tags and autolinks.

the code span `` `]` `` takes precedence over the link closing.

Since a link label must not contain a `]` that is not immediately preceded by a `\`, the above must not be parsed as a link label.

This patch fixes #642, a bug in the parser, where the range (A)...(C) was previously parsed as a shortcut link with the label (A)...(B) if the text between (A) and (B) [here: `` before ` ``] is a valid link reference. We do this by simply comparing the found end (B) and supposed end (C) to ensure that the link label closed by (C) actually ends at (C).

---

As a bonus I've included a `#[derive(Debug)]` that helped me debug this and removed an unneeded `&`. But I can drop the unrelated changes if you don't want them.


Martin1887:
Looks fine, merged.

lukas-code:
@Martin1887 would it be possible to release a 0.9.3 patch with this fix?

This bug currently causes rustdoc to crash on some inputs, see for example https://github.com/rust-lang/rust/issues/111117 and https://github.com/rust-lang/docs.rs/issues/2131. We have a new lint, [`rustdoc::unescaped_backticks`](https://doc.rust-lang.org/nightly/rustdoc/lints.html#unescaped_backticks), which is supposed to detect "broken" inline code. This happens a lot on keyboard layouts where <kbd>`</kbd> is a [dead key](https://en.wikipedia.org/wiki/Dead_key). However, the lint can only work correctly if "broken" markdown is actually parsed according to the spec.

Looking at the [roadmap for 0.10](https://github.com/users/raphlinus/projects/1), it seems like the next major release still takes some time, so I'd appreciate it if you could make a point release. I've prepared an example diff for a potential patch release here, if that helps: https://github.com/raphlinus/pulldown-cmark/compare/v0.9.2...lukas-code:pulldown-cmark:0.9.3

Alternatively, we can work around this bug on the rustdoc side, which would probably just mean disabling the lint for the time being.

Martin1887:
I see it right. I could create a new branch from the `v0.9.2` tag and then you could create a pull request to that branch from your 0.9.3 branch.

Martin1887:
The branch `branch_0.9.3` has just been created. The pull request from your branch to this new branch can already be created and hence when Ralph sends his OK the 0.9.3 release will be instantly published.

PR #647, opened 2023-05-21
Experiment with tweaking Rust install

raphlinus:
Currently it's downloading Rust from rustup.rs, but I think that's not necessary as there's already a Rust installed. This patch changes it to assume there's already a rustup to set the default toolchain.


raphlinus:
This doesn't seem productive to me, closing.

Issue #551, opened 2021-12-01
Strikethrough sequence missed for specific inputs


wyes-us:
I'm using pulldown-cmark version 0.8.0 in a project with strikethrough enabled and decided to try a bunch of different nested spans.

The following combinations have an issue where the parser misses the strikethrough sequence (`~~`).

It appears to affect just 3 or 4 depth spans starting with "strike emphasis strong" (`~~*__`) preceded by a 3 depth span starting with "emphasis strike" (`*~~`) or "strong strike" (`**~~`).

Here, I used `*__` and `**_` for "emphasis strong" and "strong emphasis" in order to make it clearer, but the result is the same if I use `***`, `___`, or `_**` and `__*` instead. Using different characters for emphasis and strong don't make any difference.

```md
*~~__emphasis strike strong__~~* ~~*__strike emphasis strong__*~~

*~~__emphasis strike strong__~~* ~~*__`strike emphasis strong code`__*~~

*~~`emphasis strike code`~~* ~~*__strike emphasis strong__*~~

*~~`emphasis strike code`~~* ~~*__`strike emphasis strong code`__*~~

**~~_strong strike emphasis_~~** ~~*__strike emphasis strong__*~~

**~~_strong strike emphasis_~~** ~~*__`strike emphasis strong code`__*~~

**~~`strong strike code`~~** ~~*__strike emphasis strong__*~~

**~~`strong strike code`~~** ~~*__`strike emphasis strong code`__*~~
```

The following procedure should create a demonstration and show the "good" and "bad" tested inputs:

```bash
cargo new missed-strike
cd missed-strike
cargo add pulldown-cmark
cat <<EOF >src/main.rs
use pulldown_cmark as pdcm;

fn gen1(z: &str, a: &str) -> String {
    format!("{}{}{}", a, z, a)
}

fn gen2(z: &str, a: &str, b: &str) -> String {
    format!("{}{}{}{}{}", a, b, z, b, a)
}

fn gen3(z: &str, a: &str, b: &str, c: &str) -> String {
    format!("{}{}{}{}{}{}{}", a, b, c, z, c, b, a)
}

fn gen4(z: &str, a: &str, b: &str, c: &str, d: &str) -> String {
    format!("{}{}{}{}{}{}{}{}{}", a, b, c, d, z, d, c, b, a)
}

fn main() {
    let emphases = ["*", "_"];
    let strikes = ["~~"];
    let strongs = ["**", "__"];
[...]

PR #650, opened 2023-05-24
fix(parse): rare error searching in the emphasis stack

Martin1887:
Fixes: #551


Martin1887:
The algorithm is not easy to understand and it is almost undocumented, but it has no sense to me to start searching the stack in the stack's length (out of the stack), so the function only was doing something when the lower bound was less than the stack's length.

This change fix the issue #551 and does not break any test, but as I have not fully understood the algorithm I would like the @raphlinus or @marcusklaas validation.

Thanks.

src/parse.rs
```diff
@@ -1013,7 +1013,10 @@ impl InlineStack {
         count: usize,
         both: bool,
     ) -> Option<InlineEl> {
-        let lowerbound = min(self.stack.len(), self.get_lowerbound(c, count, both));
+        if self.stack.is_empty() {
```
raphlinus, line 1016: This doesn't seem right, and it feels like it might be possible to construct failing cases. One is that the extra lookback in the stack needs more than 1. Another is that it fails to call `set_lowerbound()` in the case that the stack is empty.

raphlinus:
Consider the following patch instead, which fixes the case I looked at:

```diff
diff --git a/src/parse.rs b/src/parse.rs
index 1573d31..4d4c624 100644
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -1062,7 +1062,14 @@ impl InlineStack {
         }
     }
 
+    fn trim_lower_bound(&mut self, ix: usize) {
+        self.lower_bounds[ix] = self.lower_bounds[ix].min(self.stack.len());
+    }
+
     fn push(&mut self, el: InlineEl) {
+        if el.c == b'~' {
+            self.trim_lower_bound(InlineStack::TILDES);
+        }
         self.stack.push(el)
     }
 }```

Martin1887:
Thanks, Raph, this patch solves the issue.

PR #654, opened 2023-06-01
Add a separate mode to parse footnotes the same way GitHub does

### LINKED ISSUE #20: Footnote definition does not end
``` rust
extern crate pulldown_cmark as cmark;

use cmark::{Options, Parser};

fn main() {
    for e in Parser::new_ext("Hello, [^goo]\n\n\
                              [^goo]: world.\n\n\
                              Follow [link][ref]",
                             Options::all()) {
        println!("{:?}", e);
    }
}

```

The above code outputs (footnote definition extending until the end):

```
Start(Paragraph)
Text("Hello, ")
FootnoteReference("goo")
End(Paragraph)
Start(FootnoteDefinition("goo"))
Start(Paragraph)
Text("world.")
End(Paragraph)
Start(Paragraph)
Text("Follow ")
Text("[link]")
Text("[ref]")
End(Paragraph)
End(FootnoteDefinition("goo"))
```

### LINKED ISSUE #530: Document footnote syntax
It's not in the CommonMark spec, and unlike the GitHub syntax extensions, the README does *not* link to anything. I've only figured out the syntax by finding an example; where can the README point to?

### LINKED ISSUE #623: Indented footnote definitions confuse the parser
The following:

```
Foo[^foo]
Bar[^bar]

[^foo]:
    FooDef1
    FooDef2

[^bar]:
    BarDef
```

renders like so on GitHub:

image

but `pulldown_cmark::Parser` gets confused:

1. It sees the definition text as indented code blocks.
2. It does not see the end of the first definition before the start of the second.

Running the following code:

```rust
static MARKDOWN: &str = r#"\
Foo[^foo]
Bar[^bar]

[^foo]:
    FooDef1
    FooDef2

[^bar]:
    BarDef
"#;

fn main() {
    let options = Options::empty().union(Options::ENABLE_FOOTNOTES);
    let parser = Parser::new_ext(MARKDOWN, options);
    for event in parser {
        eprintln!("{event:?}");
    }
}
```

yields the following events:

- Start(Paragraph)
- HardBreak
- Text(Borrowed("Foo"))
- FootnoteReference(Borrowed("foo"))
- SoftBreak
- Text(Borrowed("Bar"))
- FootnoteReference(Borrowed("bar"))
- End(Paragraph)
- Start(FootnoteDefinition(Borrowed("foo")))
- **Start(CodeBlock(Indented))**
- Text(Borrowed("FooDef1\n"))
- Text(Borrowed("FooDef2\n"))
- End(CodeBlock(Indented))
- Start(FootnoteDefinition(Borrowed("bar")))
- Start(CodeBlock(Indented))
- Text(Borrowed("BarDef\n"))
- End(CodeBlock(Indented))
- End(FootnoteDefinition(Borrowed("bar")))
- **End(FootnoteDefinition(Borrowed("foo")))**

I would expect it to:
- treat the indented definitions as part of the definition, and
- end the first definition before starting the second.

There is a workaround:

```
Foo[^foo]
Bar[^bar]

[^foo]: FooDef1
    FooDef2

[^bar]: BarDef
```

This yields: 

- ...
- Start(FootnoteDefinition(Borrowed("foo")))
- Start(Paragraph)
- Text(Borrowed("FooDef1"))
- SoftBreak
- Text(Borrowed("FooDef2"))
- End(Paragraph)
- End(FootnoteDefinition(Borrowed("foo")))
- Start(FootnoteDefinition(Borrowed("bar")))
- Start(Paragraph)
- Text(Borrowed("BarDef"))
- End(Paragraph)
- End(FootnoteDefinition(Borrowed("bar")))

i.e. by putting content on the same line as the definition begins, the parser is happy. Unfortunately, at least one code formatter in VSCode for Markdown (maybe the built-in one?) likes to format long definitions **indented below** the `[^foo]:` opener.

notriddle:
Resolves #20

Resolves #530

Resolves #623

This change is similar to, but a more limited change than, <https://github.com/raphlinus/pulldown-cmark/pull/544>. It changes the syntax, but does not touch the generated HTML or event API.

Motivation
----------

This commit is written with usage in mdBook, rustdoc, and docs.rs in mind.

* Having a standard to follow, or at least a public test suite in [cmark-gfm] [^c], makes it easier to distinguish bugs from features.
* It makes sense to commit to following GitHub's behavior specifically, because mdBook chapters and docs.rs README files are often viewed in GitHub preview windows, so any divergence will be very annoying.
* If mdBook and docs.rs are going to use this syntax, then rustdoc should, too.
* Having both footnote syntaxes use the same API and rendering makes it more feasible for rustdoc to change the syntax over an [edition]. To introduce a syntax change in a new edition of Rust, we must make rustdoc warn anyone who writes code that will have its meaning change. To do it, run the parser twice in lockstep (with `ENABLE_FOOTNOTES` on one parser, and `ENABLE_GFM_FOOTNOTES` on the other), and warn if they diverge.
  * Alternatively, run a Crater build with this same code to check if this actually causes widespread breakage.
* In particular, using tree rewriting to push the footnotes to the end is not as useful as it sounds, since that's not enough to exactly copy the way GitHub renders footnotes. To do that, you also need to sort the footnotes by the order in which they are *referenced*, not the order in which they are defined. This type of tree rewriting is also a waste of time if you want "margin note" rendering instead of putting them all at the end.

[cmark-gfm]: https://github.com/github/cmark-gfm/blob/1e230827a584ebc9938c3eadc5059c55ef3c9abf/test/extensions.txt#L702
[edition]: https://doc.rust-lang.org/edition-guide/editions/index.html

[^c]: cmark-gfm is under the MIT license, so incorporating parts of its test suite into pulldown-cmark should be fine.


Martin1887:
Thanks for your contribution, I have to carefully review this and the other related pull requests before merging, as well as checking the possible syntaxes.

It could be included in the 0.10 version but maybe it will not.

specs/gfm_footnotes.txt
Martin1887, line 1: This spec should be placed in `third_party/GitHub`

build.rs
Martin1887, line 1: Why using the spec filename here instead of using a new type of example as with other extensions (`example_metadata_blocks` for instance)?

Doing everything in the same way increases consistency and maintenance.

specs/gfm_footnotes.txt
Martin1887, line 1: Examples should be marked as `example_gfm_footnotes` as commented in `build.rs`.

tests/lib.rs
```diff
@@ -14,14 +14,18 @@ use tendril::stream::TendrilSink;
 mod suite;
 
 #[inline(never)]
-pub fn test_markdown_html(input: &str, output: &str, smart_punct: bool, metadata_blocks: bool) {
+pub fn test_markdown_html(input: &str, output: &str, smart_punct: bool, metadata_blocks: bool, is_gfm: bool) {
```
Martin1887, line 17: There are several features related to GFM, I would name this `gfm_footnotes`.

src/lib.rs
```diff
@@ -358,5 +359,13 @@ bitflags::bitflags! {
         /// - `+++` line at start
         /// - `+++` line at end
         const ENABLE_PLUSES_DELIMITED_METADATA_BLOCKS = 1 << 8;
+        /// GitHub-compatible footnote syntax. Mutually-exclusive with `ENABLE_FOOTNOTES`.
+        const ENABLE_GFM_FOOTNOTES = 1 << 9;
```
Martin1887, line 363: I think that nothing prevents here from enabling both options, which is incorrect. This is a special case because there are two different options for the same thing with small differences.

Pulldown-cmark adheres to CommonMark and GFM extensions, so I think the best solution here is maintaining the `ENABLE_FOOTNOTES` flag enabling the new behaviour and a new `FOOTNOTES_OLD_STYLE` or something similar to enable the old behaviour not fully compatible with GFM footnotes.

This supposes a breaking change, but 0.10 release contains some of them, so I would prioritize GFM adhesion against not breaking changes here.

specs/gfm_footnotes.txt
Martin1887: Taking into account the comments of `lib.rs`, these should be the normal `example` examples and old-style footnotes should be `example_footnotes_old_style` or something similar.

src/lib.rs
Martin1887, line 1: To actually solving #530, the footnotes documentation should be pointed here.

I think the best solution is putting a link in the docstring and adding a file describing the specification of old-style footnotes and the differences with the new fully compatible GFM footnotes, also indicating the differences with GitHub behaviour if any (missing footnotes backlinks, footnotes order, ...).

Martin1887 (changes requested):

src/lib.rs
Martin1887: What do you think, @raphlinus?

Martin1887 (commented):
Changes should also be done in `main.rs` to use the new footnotes option.

Martin1887 (approved):

Martin1887:
Every seems right now, I will do some checks tomorrow and then it will be merged. Thanks for your work and fast responses!

PR #664, opened 2023-06-08
Add an example which shows raw Markdown events

mgeisler:
I’ve found this to be a very useful tool when looking at parse output. This is similar to parser-map-event-print.rs, but it shows the events using their debug representation and it avoids printing HTML.


Martin1887:
Great!, thanks for your contribution.

PR #665, opened 2023-06-10
Skip leading space on leading pipe in `scan_table_head`

### LINKED ISSUE #656: Indented table rendering as text
I'm not sure whether to call this a bug or not, as from my quick scan of the spec, I didn't find anywhere where it allows or disallows indenting a table.

Considering that pulldown-cmark allows for indented code blocks, then I assume intended tables should also be valid.

```rust
use pulldown_cmark::{html, Options, Parser};

// Notice the 2 spaces before the pipes
let text = "
  | A   | B   |
  | --- | --- |
  | 1   | 2   |

  ```rust
  println!("Foo Bar");
  ```
";

let mut options = Options::empty();
options.insert(Options::ENABLE_TABLES);
let parser = Parser::new_ext(text, options);

let mut out = String::new();
html::push_html(&mut out, parser);
println!("{}", out);
```

Which outputs the following:

```html
<p>| A   | B   |
| --- | --- |
| 1   | 2   |</p>
<pre><code class="language-rust">println!(&quot;Foo Bar&quot;);
</code></pre>
```

While I expected the following:

```html
<table><thead><tr><th>A</th><th>B</th></tr></thead><tbody>     
<tr><td>1</td><td>2</td></tr>
</tbody></table>
<pre><code class="language-rust">println!(&quot;Foo Bar&quot;);
</code></pre>
```

Here's the same table indented with 2 whitespace rendered here on GitHub:

  | A   | B   |
  | --- | --- |
  | 1   | 2   |

It also renders how I'd expect in VSCode's Markdown preview:

image

notriddle:
Fixes #656

### DIFF
--- a/examples/events.rs
+++ b/examples/events.rs
@@ -1,0 +1,22 @@
+use std::io::Read;
+
+use pulldown_cmark::{Event, Parser};
+
+/// Show all events from the text on stdin.
+fn main() {
+    let mut text = String::new();
+    std::io::stdin().read_to_string(&mut text).unwrap();
+
+    eprintln!("{text:?} -> [");
+    let mut width = 0;
+    for event in Parser::new(&text) {
+        if let Event::End(_) = event {
+            width -= 2;
+        }
+        eprintln!("  {:width$}{event:?}", "");
+        if let Event::Start(_) = event {
+            width += 2;
+        }
+    }
+    eprintln!("]");
+}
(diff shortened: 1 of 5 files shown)


Martin1887:
Thanks for your contribution! Test cases are great and your fix guided me to find the actual issue.

However, the actual fix is returning the actual offset from `calc_indent` (1 minus was returned), and your fix didn't take into account tabs, which must be taken into account.

PR #669, opened 2023-06-21
Fix fuzzing empty list items

mgeisler:
While playing some more with the fuzzer, I found that the `normalize` function was buggy: it would turn

    Start(List(None))
      Start(Item)
      End(Item)
      Start(Item)
      End(Item)
    End(List(false))

into

    Start(List(None))
      Start(Item)
        Start(Paragraph)
        End(Item)
        Start(Item)
          Start(Paragraph)
          End(Item)
        End(List(false))

by mistake.

This PR moves the logic to a library (inside the fuzzing crate) so we can test this and perhaps reuse it across different fuzzers.


Martin1887:
Thanks!, I will try to review it this week.

mgeisler:
> Thanks!, I will try to review it this week.

No stress, this is a fun side project for me! 🙂

PR #671, opened 2023-06-21
Update MSRV to 1.60

### LINKED ISSUE #670: Old Cargo fails to select version of aho-corasick
CI fails with

```
    Updating crates.io index
error: failed to select a version for the requirement `aho-corasick = "=1.0.1"`
candidate versions found which didn't match: 0.7.20, 0.7.19, 0.7.18, ...
location searched: crates.io index
required by package `regex v1.8.1`
    ... which satisfies dependency `regex = "=1.8.1"` of package `pulldown-cmark v0.9.2 (/home/vsts/work/1/s)`
```
I can reproduce this locally with Rust 1.56.0.

A `cargo update` is enough to get things working, but this changes the `Cargo.lock` file — and it changes it differently that what you get when running `cargo update` on nightly Rust.

Is the policy of the project to keep the code compatible with Rust 1.56? If so, we should probably commit the lock file from that version.

mgeisler:
This is the result of running `cargo update` with Rust 1.60.

Fixes #670.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.20"
+version = "1.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cc936419f96fa211c1b9166887b38e5e40b19958e5b895be7c1f93adec7071ac"
+checksum = "43f6cb1bf222025340178f382c426f13757b2960e89779dfcb319c32542a5a41"
 dependencies = [
  "memchr",
 ]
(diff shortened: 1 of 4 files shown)


mgeisler:
Ups, running `cargo test` gives me

```
error: package `clap v3.2.25` cannot be built because it requires rustc 1.56.1 or newer, while the currently active rustc version is 1.56.0
```

After that I get:

```
error: package `rayon-core v1.11.0` cannot be built because it requires rustc 1.59 or newer, while the currently active rustc version is 1.56.1
```

and finally

```
error: package `log v0.4.19` cannot be built because it requires rustc 1.60.0 or newer, while the currently active rustc version is 1.59.0
```

So I guess the MSRV should be bumped to 1.60 now.

Martin1887:
Perfect, merged. Thanks!

PR #673, opened 2023-06-22
Implement differential fuzzer for pandoc

notriddle:


Martin1887:
Thanks for your contribution.

The goal of the project is supporting CommonMark and Github Flavored Markdown, Pandoc target is far from the scope. May this fuzzer provide help to catch errors in CommonMark+GFM? The only case I find is when both pulldown-cmark and commonmark.js are wrong and Pandoc does the job well.

On the other hand, this code is independent of the final binary and only a dev tool. What do you think, @raphlinus?

fuzz/src/lib.rs
```diff
@@ -186,6 +664,42 @@ pub fn xml_to_events(xml: &str) -> anyhow::Result<Vec<Event>> {
     Ok(events)
 }
 
+pub fn normalize_pandoc(events: Vec<Event<'_>>) -> Vec<Event<'_>> {
```
mgeisler, line 667: I think this is cool! I would rename `normalize` below to `normalize_commonmarkjs` or similar.

fuzz/fuzz_targets/pandoc.rs
```diff
@@ -0,0 +1,170 @@
...
+    //
+    // commonmark-hs also trims footnote reference names.
+    // I'm not going to bother reporting it as a bug, since it's
+    // obviously on purpose and it's not gonna matter.
+    let mut footstack = vec![];
+    for event in &pulldown_cmark_events {
+        use pulldown_cmark::{Event, Tag, TagEnd};
+        match event {
+            Event::Start(Tag::FootnoteDefinition(id)) => {
+                if id.starts_with("\n") || id.ends_with("\n") || id.starts_with("\r") || id.ends_with("\r") || id.starts_with(" ") || id.starts_with("\t") || id.contains("  ") || id.contains("\t ") || id.contains(" \t") || id.contains("\t\t") || id.ends_with(" ") || id.ends_with("\t") { return };
```
mgeisler, line 64: Perhaps it would be simpler to use the slice variants of the `starts_with` and `ends_with` patterns:

```suggestion
                let whitespace = &['\n', '\r', ' ', '\t'];
                if id.starts_with(whitespace)
                    || id.ends_with(whitespace)
                    || id.contains("\t ")
                    || id.contains(" \t")
                    || id.contains("\t\t")
                {
                    return;
                };
```

PR #668, opened 2023-06-20
Allow lowercase letters in HTML declarations

notriddle:
This was changed in the commonmark spec in https://github.com/commonmark/commonmark-spec/issues/620 (included in commonmark 0.30)

Fixes #659

Issue #659, opened 2023-06-08
Parsing `"<!e -hhl"` gives text, not HTML


mgeisler:
Events from pulldown-cmark:

```rust
"<!e -hhl" -> [
  Start(Paragraph)
    Text(Boxed("<!e -hhl"))
  End(Paragraph)
]
```

Events from commonmark.js:

```rust
"<!e -hhl" -> [
  Html(Boxed("<!e -hhl"))
]
```

XML from commonmark.js:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE document SYSTEM "CommonMark.dtd">
<document xmlns="http://commonmark.org/xml/1.0">
  <html_block>&lt;!e -hhl</html_block>
</document>
```

This looks like [start condition 4](https://spec.commonmark.org/0.30/#html-blocks) in the specification. 

I found this by testing against revision 20b52e5 of https://github.com/commonmark/commonmark.js/ — I get the same result as pulldown-cmark when I test against the code in the [dingus](https://spec.commonmark.org/dingus/?text=%3C!e%20-hhl).

PR #677, opened 2023-06-24
Allow fuzzing all HTML declarations

mgeisler:
Issue #659 was closed via #668, so we can clean this up now.


Martin1887:
Perfect, merged.

PR #679, opened 2023-06-25
Fix table/paragraph interruption when nested

notriddle:
This fixes the "peek at next line" code to account for block quotes. It's a bit more complicated than just skipping ahead, because the old code would cause the block quote to get interrupted by a table nested inside it (it simulated reaching a blank line, which interrupts the block quote). Instead, this commit hooks into the normal paragraph-interrupt code.


Martin1887:
Perfect, thanks!

PR #681, opened 2023-06-27
Fix spec violation on empty-list-item-cannot-interrupt-paragrph rule

### LINKED ISSUE #680: Lists are parsed in places that cmark doesn't see them
Fuzzing result

Events from pulldown-cmark:

```rust
";\n*\n%\n" -> [
  Start(Paragraph)
    Text(Borrowed(";"))
  End(Paragraph)
  Start(List(None))
    Start(Item)
    End(Item)
  End(List(false))
  Start(Paragraph)
    Text(Borrowed("%"))
  End(Paragraph)
]
```

Events from pandoc:

```rust
";\n*\n%\n" -> [
  Start(Paragraph)
    Text(Boxed(";"))
    SoftBreak
    Text(Boxed("*"))
    SoftBreak
    Text(Boxed("%"))
  End(Paragraph)
]
```

Events from commonmark.js:

```rust
";\n*\n%\n" -> [
  Start(Paragraph)
    Text(Boxed(";"))
    SoftBreak
    Text(Boxed("*"))
    SoftBreak
    Text(Boxed("%"))
  End(Paragraph)
]
```

notriddle:
Fixes #680

This was definitely a spec violation. **Emphasis mine.**

> The following rules define [list items](https://spec.commonmark.org/0.30/#list-items):
>
>  1. Basic case. If a sequence of lines Ls constitute a sequence of blocks Bs starting with a character other than a space or tab, and M is a list marker of width W followed by 1 ≤ N ≤ 4 spaces of indentation, then the result of prepending M and the following spaces to the first line of Ls, and indenting subsequent lines of Ls by W + N spaces, is a list item with Bs as its contents. The type of the list item (bullet or ordered) is determined by the type of its list marker. If the list item is ordered, then it is also assigned a start number, based on the ordered list marker.
>
>     Exceptions:
>      1. When the first list item in a list interrupts a paragraph—that is, when it starts on a line that would otherwise count as paragraph continuation text—then (a) **the lines Ls must not begin with a blank line**, and (b) if the list item is ordered, the start number must be 1.


test.md
```diff
@@ -0,0 +1,4 @@
+**!*!**
+
+__!_!__
+
```
Martin1887, line 4: Should not this file be removed?

Martin1887:
Thanks for your contribution, it looks fine.

PR #682, opened 2023-06-28
Use correct `both` flag when pushing remaining delimiters to stack

### LINKED ISSUE #657: Strong with nested emphasis parses as "**"
The string `"****foo*bar*baz****"` is parsed as follows:

```rust
"****foo*bar*baz****" -> [
  Start(Paragraph)
    Text(Borrowed("*"))
    Text(Borrowed("*"))
    Start(Emphasis)
      Start(Emphasis)
        Text(Borrowed("foo"))
      End(Emphasis)
      Text(Borrowed("bar"))
    End(Emphasis)
    Text(Borrowed("baz"))
    Text(Borrowed("*"))
    Text(Borrowed("*"))
    Text(Borrowed("*"))
    Text(Borrowed("*"))
  End(Paragraph)
]
```

This differs from [the commonmark.js reference implementation](https://spec.commonmark.org/dingus/?text=****foo*bar*baz****) which gives the following AST:

```xml
  <paragraph>
    <strong>
      <emph>
        <emph>
          <text>foo</text>
        </emph>
        <text>bar</text>
      </emph>
      <text>baz</text>
    </strong>
    <text>**</text>
  </paragraph>
```

I found this while trying to make pulldown-cmark-to-cmark round-trip the Markdown text with a fuzzer. I'm trying to make a single paragraph of text work and the fuzzer looks like this:

```rust
#![no_main]

use libfuzzer_sys::fuzz_target;
use pulldown_cmark::Parser;
use pulldown_cmark::{Event, Tag};
use pulldown_cmark_to_cmark::cmark;

fn round_trip(text: &str) -> String {
    let mut result = String::with_capacity(text.len());
    let events = Parser::new(&text);

    let events = events.collect::<Vec<_>>();

    if std::env::var("EVENTS").is_ok() {
        let mut width = 0;
        eprintln!("{text:?} -> [");
        for event in &events {
            if let Event::End(_) = event {
                width -= 2;
            }
            eprintln!("  {:width$}{event:?}", "");
            if let Event::Start(_) = event {
                width += 2;
            }
        }
        eprintln!("]");
    }

    cmark(events.into_iter(), &mut result).unwrap();
    result
}

fuzz_target!(|text: String| {
    if text.contains(&['\n', '\r', '`']) {
        return;
    }

    let round_trip_1 = round_trip(&text);
    match Parser::new(&round_trip_1).collect::<Vec<_>>().as_slice() {
        [Event::Start(Tag::Paragraph), .., Event::End(Tag::Paragraph)] => {}
        _ => return,
    }

    let round_trip_2 = round_trip(&round_trip_1);
    let round_trip_3 = round_trip(&round_trip_2);
    assert_eq!(round_trip_2, round_trip_3);
});
```

notriddle:
Fixes #657

Here's the delimiter match-ups that pulldown-cmark made before this commit, annotated to make clear the both flag and the multiple-of-three matching that was erroneously applied:

    ****foo*bar*baz****
    ^^12   2   1   ^^^^ = 6
           -   - both

    ***R]*-*
    ^^1  1 ^ = 3
         - both

### DIFF
--- a/specs/regression.txt
+++ b/specs/regression.txt
@@ -1248,3 +1248,20 @@ Table interruption when nested inside paragraph.
 </tbody></table>
 </blockquote>
 ````````````````````````````````
+
+
+ISSUE 657
+
+```````````````````````````````` example
+***R]*-*
+.
+<p>*<em><em>R]</em>-</em></p>
+````````````````````````````````
+
+The below example adds a trailing space.
+
+```````````````````````````````` example
+****foo*bar*baz****
+.
+<p><strong><em><em>foo</em>bar</em>baz</strong>**</p>
+````````````````````````````````
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -620,7 +620,7 @@ impl<'input, 'callback> Parser<'input, 'callback> {
                                     start: el.start,
                                     count: el.count - match_count,
                                     c: el.c,
-                                    both,
+                                    both: el.both,
                                 })
                             }
                             count -= match_count;
--- a/tests/suite/regression.rs
+++ b/tests/suite/regression.rs
@@ -1481,3 +1481,23 @@ fn regression_test_98() {
 
     test_markdown_html(original, expected, false, false, false);
 }
+
+#[test]
+fn regression_test_99() {
+    let original = r##"***R]*-*
+"##;
+    let expected = r##"<p>*<em><em>R]</em>-</em></p>
+"##;
+
+    test_markdown_html(original, expected, false, false, false);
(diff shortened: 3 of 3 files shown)


Martin1887:
Thanks for your contribution.

I need more time to fully understand the emphasis algorithm, but it looks well.

Could you take a look, @raphlinus?

raphlinus (approved):
I believe this change is correct. Good catch!

Martin1887:
Perfect, thanks!

Issue #592, opened 2022-06-05
Heading parsing inconsistency.


crlf0710:
When parsing
```text
###
````
and
```text
### ###
```
One might expect the same event sequence be generated... Instead it doesn't, with the second case generating a blank text event between start and end. Maybe it'll be better if the blank event doesn't get generated?

PR #683, opened 2023-07-01
fix: empty text events in ATX headings with trailing `#`

Martin1887:
Fixes: #592


mgeisler (approved):
Nice, though this makes me wonder if it would be possible to suppress emitting empty `Text` events in general?

PR #688, opened 2023-07-04
Fix incorrect scan of bad email address

### LINKED ISSUE #687: `<@1>` parsed as inline link
Events from pulldown-cmark:

```rust
"<@1>" -> [
  Start(Paragraph)
    Start(Link { link_type: Inline, dest_url: Borrowed("@1"), title: Borrowed(""), id: Borrowed("") })
      Text(Borrowed("@1"))
    End(Link)
  End(Paragraph)
]
```

Events from commonmark.js:

```rust
"<@1>" -> [
  Start(Paragraph)
    Text(Boxed("<@1>"))
  End(Paragraph)
]
```

XML from [commonmark.js](https://spec.commonmark.org/dingus/?text=%3C%401%3E):

```xml
  <paragraph>
    <text>&lt;</text>
    <text>@1&gt;</text>
  </paragraph>
</document>
```

I looked at the [section on autolinks in the spec](https://spec.commonmark.org/0.30/#autolinks) and the problem seems to be that `scan_email` accepts `@1` as a valid email address. However, the spec says

> An [email address](https://spec.commonmark.org/0.30/#email-address), for these purposes, is anything that matches the [non-normative regex from the HTML5 spec](https://html.spec.whatwg.org/multipage/forms.html#e-mail-state-(type=email)):
>
>     /^[a-zA-Z0-9.!#$%&'*+/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?
>     (?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$/

There is a comment at the start of `scan_line` saying that the text is scanned by hand to avoid using a regex library. I suppose we could rewrite the code there to parse the regex more faithfully, or perhaps we could pull in a regex library?

raphlinus:
The old scanner mistakenly accepted an email address with an empty label before the @ symbol.

This commit also adds some hand-written tests. I also ran a bit of property testing against the regex.

Also renumbers regression tests that were incorrectly numbered (probably due to merge conflict).

Fixes #687


raphlinus:
If people want to replicate my property testing (which is admittedly not very sophisticated), here's a [gist](https://gist.github.com/raphlinus/2fdb1ec32521fde2a73be043c7c7bf68). It was able to detect the reported issue, but not artificially induced bugs in label length, likely because the probability of a randomly generated label being valid gets very small as the length increases. A better strategy would be to generate a good email address from the regex, then mutate one byte. I didn't do that, but perhaps somebody is motivated to.

Martin1887:
Great!

mgeisler:
Cool,  that was much easier than I thought!

PR #691, opened 2023-07-11
Add special handling for backslash-escaped pipes in codeblocks

### LINKED ISSUE #356: Problems with escaping pipes in tables
There doesn't seem to be a way to escape pipes inside a backtick string inside a table.  Example:

```markdown
| My Table |
|----------|
| `Foo \| Bar` |
```

Returns a code event with the literal backslash still in it: `Code(Borrowed("Foo \\| Bar"))`. 

However, if you don't include the backslash, it breaks the table (and indeed, that is what GitHub renders here):

image

This is complicated since inside a backtick string [per the commonmark spec](https://spec.commonmark.org/0.29/#example-338), "All backslashes are treated literally".  The GHFM spec has an escape hatch [here](https://github.github.com/gfm/#example-200) where they specify "Include a pipe in a cell’s content by escaping it, **including inside other inline spans**".  So there is a little conflict there, but I think if pulldown-cmark wants to follow GHFM tables, it should probably handle escapes the same way.

A workaround for now is to use HTML, such as `<code>Foo \| Bar </code>` or `<code>Foo &vert; Bar </code>`

EDIT: Original issue below, the rest of the issues below have been fixed.

----

I'm running into some issues with escaping pipes which makes it difficult to use them in tables.

**Escaping only works at the beginning inside backticks**

Given the text ``` `\|` ```  I get the expected event `Code(Borrowed("|"))`.

Given the text ``` `foo \|` ``` I get `Code(Borrowed("foo \\|"))`, notice the unexpected backslash in the output.  Escaping a pipe only works if it is the very first thing inside backticks.  I would expect it to work anywhere inside backticks.

**Escaping inside code block does not work**

Given the text ``` <code>\|</code> ``` results in exactly the same output (that is, with the backslash).  I would expect the backslash to be removed (and is how GitHub interprets it).  This is an issue because if you try to use ```<code>|</code>``` inside a table, the pipe is interpreted as a column separator.  Adding the backslash prevents that, but results in an unwanted backslash in the output.

For now I think I'll be able to work around this using `&vert;`.

notriddle:
Fixes #356

Compare with the rendering in [GitHub] and [Pandoc].

[GitHub]: https://gist.github.com/notriddle/c027512ee849f12098fec3a3256c89d3
[Pandoc]: https://pandoc.org/try/?params=%7B%22text%22%3A%22%7C+Description+%7C+Test+case+%7C%5Cn%7C-------------%7C-----------%7C%5Cn%7C+Single++++++%7C+%60%5C%5C%60+++++++%7C%5Cn%7C+Double++++++%7C+%60%5C%5C%5C%5C%60++++++%7C%5Cn%7C+Basic+test++%7C+%60%5C%5C%7C%60++++++%7C%5Cn%7C+Basic+test+2%7C+%60%5C%5C%7C%5C%5C%7C%5C%5C%60+++%7C%5Cn%7C+Basic+test+3%7C+%60x%5C%5C%7Cy%5C%5C%7Cz%5C%5C%60%7C%5Cn%7C+Not+pipe++++%7C+%60%5C%5C.%60++++++%7C%5Cn%7C+Combo+++++++%7C+%60%5C%5C.%5C%5C%7C%5C%5C%60+++%7C%5Cn%7C+Extra+++++++%7C+%60%5C%5C%5C%5C%5C%5C.%60++++%7C%5Cn%7C+Wait%2C+what%3F+%7C+%60%5C%5C%5C%5C%7C%60+++++%7C%5Cn%7C+Wait%2C+what%3F+%7C+%60%5C%5C%5C%5C%5C%5C%7C%60++++%7C%5Cn%7C+Wait%2C+what%3F+%7C+%60%5C%5C%5C%5C%5C%5C%5C%5C%7C%60+++%7C%5Cn%7C+Wait%2C+what%3F+%7C+%60%5C%5C%5C%5C%5C%5C%5C%5C%5C%5C%7C%60++%7C%5Cn%7C+Wait%2C+what%3F+%7C++++++++++%5C%5C%7C%5Cn%7C+Wait%2C+what%3F+%7C+++++++++%5C%5C%5C%5C%7C%5Cn%7C+Wait%2C+what%3F+%7C++++++++%5C%5C%5C%5C%5C%5C%7C%22%2C%22to%22%3A%22html5%22%2C%22from%22%3A%22commonmark_x%22%2C%22standalone%22%3Afalse%2C%22embed-resources%22%3Afalse%2C%22table-of-contents%22%3Afalse%2C%22number-sections%22%3Afalse%2C%22citeproc%22%3Afalse%2C%22html-math-method%22%3A%22plain%22%2C%22wrap%22%3A%22auto%22%2C%22highlight-style%22%3Anull%2C%22files%22%3A%7B%7D%2C%22template%22%3Anull%7D

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -812,6 +812,11 @@ impl<'input, 'callback> Parser<'input, 'callback> {
 
             while ix != close {
                 let next_ix = self.tree[ix].next.unwrap();
+                let end = if next_ix == close {
+                    span_end
+                } else {
+                    self.tree[next_ix].item.start
+                };
                 if let ItemBody::HardBreak | ItemBody::SoftBreak = self.tree[ix].item.body {
                     if drop_enclosing_whitespace {
                         // check whether break should be ignored
@@ -838,12 +843,37 @@ impl<'input, 'callback> Parser<'input, 'callback> {
                         new_buf.push(' ');
                         buf = Some(new_buf);
                     }
+                } else if self.tree.is_in_table() && self.text[self.tree[ix].item.start..end].contains("|") {
+                    for (i, c) in bytes[self.tree[ix].item.start..end].into_iter().copied().enumerate() {
+                        if c != b'|' {
+                            continue;
+                        }
+                        let pipe_position = self.tree[ix].item.start + i;
+                        let buf = if let Some(ref mut buf) = buf {
+                            buf.push_str(&self.text[self.tree[ix].item.start..pipe_position]);
+                            buf
+                        } else {
+                            let mut new_buf = String::with_capacity(pipe_position - span_start);
+                            new_buf.push_str(&self.text[span_start..pipe_position]);
+                            buf.insert(new_buf)
+                        };
+                        let mut backslash_count = 0;
+                        for c in buf.bytes().rev() {
+                            if c == b'\\' {
+                                backslash_count += 1;
+                            } else {
+                                break;
+                            }
+                        }
+                        assert!(
+                            backslash_count % 2 == 1,
+                            "if number of backslashes were even, pipe wouldn't be escaped",
+                        );
+                        let trim = (backslash_count / 2) + 1;
+                        buf.truncate(buf.len() - trim);
+                        buf.push_str(&self.text[pipe_position..end]);
+                    }
                 } else if let Some(ref mut buf) = buf {
-                    let end = if next_ix == close {
-                        span_end
-                    } else {
-                        self.tree[ix].item.end
-                    };
                     buf.push_str(&self.text[self.tree[ix].item.start..end]);
                 }
                 ix = next_ix;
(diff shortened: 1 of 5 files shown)


Martin1887:
This is really weird... maybe in this case we should prioritize consistence to mimic GitHub behavior, and with more reasons taking into account that it conflicts the normal behavior of CommonMark outside tables.

The behavior inside code blocks seems reasonable but outside code blocks I would use the scaping rules of CommonMark.

src/parse.rs
```diff
@@ -838,12 +843,37 @@ impl<'input, 'callback> Parser<'input, 'callback> {
...
+                        };
+                        let mut backslash_count = 0;
+                        for c in buf.bytes().rev() {
+                            if c == b'\\' {
+                                backslash_count += 1;
+                            } else {
+                                break;
+                            }
+                        }
+                        assert!(
```
Martin1887, line 868: This `assert!` should be removed to avoid panics at runtime, right?

Martin1887:
This is closed in favor of #699, but more discussion will be happen around this.

Issue #473, opened 2020-08-31
Separate inline html and html block output in output.


crlf0710:
It is useful to preserve these information for non-rendering processing of markdown document data.

raphlinus:
This seems like a valid request. I note that the [CommonMark dingus](https://spec.commonmark.org/dingus/) generates different AST nodes for inline and block html. I wonder if that changed, because in general I've tried to make the tag/event signature match that AST.

marcusklaas:
If I recall correctly, pulldown used to make the distinction as well.

raphlinus:
Interesting, looks like #343 unified the concepts. Definitely a simplification if all you're doing is rendering, but I can imagine cases where distinction is useful.

crlf0710:
Maybe consider adding the start and end of a html block as a kind of "tag", so for use cases doesn't care about it, people can just ignore it. and for use cases that cares about it, extra handling can be added.

Specially, i'm thinking about using this distinction in fixing https://github.com/Byron/pulldown-cmark-to-cmark/issues/16  .

PR #692, opened 2023-07-15
feat!: separate inline HTML and HTML blocks

Martin1887:
Fixes: #473.

I have replicated the #343 previous behaviour, please check if you find any issues.


raphlinus:
I didn't review it in detail but as previously stated this seems like a good direction. If there's a specific concern in the implementation, let me know and I can take a deeper look.

PR #693, opened 2023-07-18
Cargo.toml: Add rust-version for MSRV.

waywardmonkeys:


Martin1887:
Good point, thanks for your contribution!

PR #699, opened 2023-08-10
Add special handling for backslashed table pipes

### LINKED ISSUE #356: Problems with escaping pipes in tables

notriddle:
Fixes #356

This is an alternative version of https://github.com/raphlinus/pulldown-cmark/pull/691 that follows GitHub's behavior more closely (it gives `\|` higher precedence in the parser than any other backslash escape).

Compare with the rendering in [GitHub].

[GitHub]: https://gist.github.com/notriddle/c027512ee849f12098fec3a3256c89d3

### DIFF
--- a/specs/table.txt
+++ b/specs/table.txt
@@ -449,3 +449,100 @@ a\
 <tr><td>d</td><td>e</td></tr>
 </tbody></table>
 ````````````````````````````````
+
+
+As a special case, pipes in inline code in tables are escaped
+with backslashes.
+
+The parsing rule for CommonMark is that block structures are parsed before
+inline structures are. Normally, this means backslashes aren't allowed to have
+any effect on block structures at all. Tables do consider backslashes, but
+not the same way inline syntax does: they have higher precedence than anything
+else, as-if they were parsed in a completely separate pass.
+
+This rule should be identical to GitHub's. See
+<https://gist.github.com/notriddle/c027512ee849f12098fec3a3256c89d3>
+for what they do. This is totally different from Pandoc's `commonmark_x`,
+for example, and from many other Markdown parsers, because of some weird
+corner cases it creates.
+
+```````````````````````````````` example
+| Description | Test case |
+|-------------|-----------|
+| Single      | `\`       |
+| Double      | `\\`      |
+| Basic test  | `\|`      |
+| Basic test 2| `\|\|\`   |
+| Basic test 3| `x\|y\|z\`|
+| Not pipe    | `\.`      |
+| Combo       | `\.\|\`   |
+| Extra       | `\\\.`    |
+| Wait, what? | `\\|`     |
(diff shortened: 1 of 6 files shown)


Martin1887:
Thanks for your contribution!

I think #691 is more consistent with CommonMark and should be merged instead of this despite not following the GitHub behaviour, but it is a difficult decision.

raphlinus:
It is a difficult decision, as there are tradeoffs on both sides. I really wish there was a standardization effort that could address this.

Having wrestled with it, my personal feeling is to go with #699. It's arguably less clean, but it results in fewer distinct rendering possibilities for the same source markdown. But I'm also open to more discussion on the topic, and would encourage any effort to get this resolved more systematically, for example on the commonmark forums.

Martin1887:
OK, I merge this because it solves and issue targeted for 0.10 and we can try to get more feedback on the CommonMark forums.

In a future version the decision could be changed, but it will probably be also different to #691.

Thanks!

Issue #697, opened 2023-07-31
Making 'static BrokenLinkCallback


ollpu:
I'm having trouble constructing a `Parser` with a broken link callback such that `'callback: 'static`. My use case is to use a parser in a function with the following signature:

```rust
fn parse(input: &str) -> impl Iterator<Item = Event<'_>>
```

However, if I want to add a broken link callback, I am unable to satisfy that signature.

For example:

```rust
fn callback<'input>(_link: BrokenLink<'input>) -> Option<(CowStr<'input>, CowStr<'input>)> {
    todo!()
}

fn parse(input: &str) -> impl Iterator<Item = Event<'_>> {
    Parser::new_with_broken_link_callback(&input, Options::empty(), Some(&mut callback))
}
```

gives

```
error[E0716]: temporary value dropped while borrowed
  --> src/lib.rs:32:79
   |
31 | fn parse(input: &str) -> impl Iterator<Item = Event<'_>> {
   |                 - let's call the lifetime of this reference `'1`
32 |     Parser::new_with_broken_link_callback(&input, Options::empty(), Some(&mut callback))
   |     --------------------------------------------------------------------------^^^^^^^^--
   |     |                                                                         |
   |     |                                                                         creates a temporary value which is freed while still in use
   |     argument requires that borrow lasts for `'1`
33 | }
   | - temporary value is freed at the end of this statement

For more information about this error, try `rustc --explain E0716`.
```

I tried some tricks, but I couldn't come up with a way to construct a `&'static mut dyn FnMut()`. Going through `const _: fn() = callback` gives a similar error because of the temporary created at the use site. A `static` item cannot be directly borrowed mutably, so it can't be put in a `&mut` without some temporary wrapper.

Using `static mut` works but requires unsafe. If there's a way to do this without unsafe I'd be interested to know!

I believe a more flexible API would be to accept `Option<Box<dyn FnMut(..) + 'callback>>` instead of `Option<&'callback mut dyn FnMut(..)>`. A value of the latter can be converted to the former by simply boxing the mutable reference, as they transitively implement `FnMut`. Performance might be slightly worse, but that seems negligible.

The BrokenLinkCallback API was reworked previously in #509. Changing to boxes would be yet another breaking change unfortunately.

ollpu:
Reading #509 more carefully, the trait + generics approach suggested there would also address this issue.

Martin1887:
So can this be closed?

Thanks.

[...]

PR #701, opened 2023-08-15
Use generics for broken link callback

### LINKED ISSUE #508: BrokenLinkCallback: Ownership and Send + Sync?
I was looking around the codebase and noticed that Rust has determined that `Parser` was `!Send` and `!Sync`. I thought it was odd because from a very high level perspective, it seems reasonable to expect us to be able to send a parser object between threads without difficulty--synchronous access maybe not so much (I can't really imagine a use case for when a client might want to have a parser be access from multiple threads at the same time) but `Send` implies `Sync` so ¯\\\_(ツ)\_/¯.

It looks like the primary problem is that the type `BrokenLinkCallback` isn't Send nor Sync because we don't require the trait object to be send or sync:

```rust
pub type BrokenLinkCallback<'a> =
    Option<&'a mut dyn FnMut(BrokenLink) -> Option<(CowStr<'a>, CowStr<'a>)>>;
```

A quick fix is to just require the trait object reference to be Send + Sync:

```rust
pub type BrokenLinkCallback<'a> =
    Option<&'a mut (dyn Send + Sync + FnMut(BrokenLink) -> Option<(CowStr<'a>, CowStr<'a>)>)>;
```

but since `BrokenLinkCallback` is part of the public API, this would be a breaking change as it restricts what functions are accepted.

As a [refresher on Closure auto-impl rules][refresher]:
 - A closure is Sync if all captured variables are Sync.
 - A closure is Send if all variables captured by non-unique immutable reference are Sync, and all values captured by unique immutable or mutable reference, copy, or move are Send.

I might be wrong, but I don't think adding this requirement would be too difficult on users. I would expect most use cases to already be `Send` and `Sync` (our tests need no modification, for example), and I believe the current definition `&mut dyn FnMut` already forces some sort of Send/Sync primitive around the Parser itself, if they need to do so.

On that note, is there a particular reason why we accept a `&mut` to a callback, instead of an owned instance we `Box` instead or have `Parser` be generic over the trait?

[refresher]: https://doc.rust-lang.org/reference/types/closure.html

ollpu:
As discussed in #508, #509 and #697, the current broken link callback API is a little clunky. This PR changes it to use generics instead of dynamic dispatch.

I haven't considered the performance or binary size implications of this. There should be no difference when not using the callback (all instances use `DefaultBrokenLinkCallback`), but when using it, all the parser's methods will be monomorphized somewhat unnecessarily. Ultimately it's up to you (the maintainers) to decide.

On the flip side, this solution provides the most flexible API. `Parser` will now transitively implement `Send` & `Sync` if the callback does. It can also accept `'static` callbacks:

```rust
fn callback<'input>(link: BrokenLink<'input>) -> Option<(CowStr<'input>, CowStr<'input>)> {
    todo!()
}

fn parse(input: &str) -> impl Iterator<Item = Event<'_>> + '_ {
    Parser::new_with_broken_link_callback(input, Options::empty(), Some(callback))
}
```

Some finer details that I'm unsure of:

- `new_with_broken_link_callback` still takes an `Option`, though it could just accept the callback directly. Doing it like this reduces API breakage as the previous signature `Option<&mut dyn FnMut>` is still accepted.
- I chose for `BrokenLinkCallback` to be a trait with a method `handle_broken_link`. It could also be a "trait alias" with an `FnMut` bound so that the separate function name wouldn't be exposed. I feel this makes the no-op `DefaultBrokenLinkCallback` cleaner, but it could also be a function.
- `BrokenLinkCallback` could be more general – if it is decided to add other types of callbacks later, it would be convenient to add them to a broader "`Callback`" trait. In this case the blanket impl for `FnMut` would be questionable though.

Closes #508, #697

### DIFF
--- a/examples/broken-link-callbacks.rs
+++ b/examples/broken-link-callbacks.rs
@@ -6,7 +6,7 @@ fn main() {
 
     // Setup callback that sets the URL and title when it encounters
     // a reference to our home page.
-    let callback = &mut |broken_link: BrokenLink| {
+    let callback = |broken_link: BrokenLink| {
         if broken_link.reference.as_ref() == "my website" {
             println!(
                 "Replacing the markdown `{}` of type {:?} with a working link",
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -98,7 +98,9 @@ mod tree;
 
 use std::{convert::TryFrom, fmt::Display};
 
-pub use crate::parse::{BrokenLink, BrokenLinkCallback, LinkDef, OffsetIter, Parser, RefDefs};
+pub use crate::parse::{
+    BrokenLink, BrokenLinkCallback, DefaultBrokenLinkCallback, OffsetIter, Parser, RefDefs,
+};
 pub use crate::strings::{CowStr, InlineStr};
 pub use crate::utils::*;
 
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -132,12 +132,12 @@ pub struct BrokenLink<'a> {
 }
 
 /// Markdown event iterator.
-pub struct Parser<'input, 'callback> {
+pub struct Parser<'input, F = DefaultBrokenLinkCallback> {
     text: &'input str,
     options: Options,
     tree: Tree<Item>,
     allocs: Allocations<'input>,
-    broken_link_callback: BrokenLinkCallback<'input, 'callback>,
+    broken_link_callback: Option<F>,
     html_scan_guard: HtmlScanGuard,
 
     // used by inline passes. store them here for reuse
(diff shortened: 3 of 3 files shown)


Martin1887:
Thanks, I will try to review and analyze it in the next days.

ollpu:
Right, I should add that impl.

ollpu:
I added `impl BrokenLinkCallback<'i> for Box<dyn BrokenLinkCallback<'i>>` now.

I also considered replacing `DefaultBrokenLinkCallback` with this, so that if there are e.g. multiple crates linked together, but only some of them needing to define the callback, they would still automatically agree on a single monomorphization. Trouble is that the input lifetime needs to be carried around. This complicates signatures, so I didn't do that for now.

ollpu:
Turns out my use case naturally evolved into having two different instantiations of `Parser`, one with a callback and one without.

`pulldown-cmark` is the only dependency, with `default-features = false, features = ["html"]`. I also have `strip = "symbols"` set.

Release binary size before this PR was 0.725 MB, and 0.748 MB after. I don't know if I can meaningfully measure performance with my setup.

Martin1887:
Sorry for the delay, this seems good to me!

justjavac:
After merging this PR, it no longer works

Martin1887:
Thanks for the feedback, fixed in #746.

Issue #702, opened 2023-08-24
Parsing to normalize HTML in spec tests not always appropriate


ollpu:
Currently when normalizing HTML for the correct answer and library output, the HTML is parsed with `html5ever`. This doesn't seem appropriate because some tests explicitly pass through invalid HTML.

I enabled error reporting like this:

<details><summary>diff</summary>

```diff
diff --git a/tests/lib.rs b/tests/lib.rs
index c1c2181..583c8fb 100644
--- a/tests/lib.rs
+++ b/tests/lib.rs
@@ -1,6 +1,8 @@
 #![cfg(feature = "html")]
 
 use html5ever::serialize::{serialize, SerializeOpts};
+use html5ever::tokenizer::TokenizerOpts;
+use html5ever::tree_builder::TreeBuilderOpts;
 use html5ever::{driver as html, local_name, namespace_url, ns, QualName};
 use markup5ever_rcdom::{Handle, NodeData, RcDom, SerializableHandle};
 use pulldown_cmark::{Options, Parser};
@@ -38,7 +40,12 @@ pub fn test_markdown_html(input: &str, output: &str, smart_punct: bool, metadata
     let p = Parser::new_ext(input, opts);
     pulldown_cmark::html::push_html(&mut s, p);
 
-    assert_eq!(normalize_html(output), normalize_html(&s));
+    println!("normalizing correct HTML");
+    let correct = normalize_html(output);
+    println!("ok");
+    println!("normalizing result HTML");
+    let result = normalize_html(&s);
+    assert_eq!(correct, result);
 }
 
 lazy_static::lazy_static! {
@@ -114,7 +121,16 @@ lazy_static::lazy_static! {
 fn make_html_parser() -> html::Parser<RcDom> {
     html::parse_fragment(
         RcDom::default(),
-        html::ParseOpts::default(),
+        html::ParseOpts {
+            tokenizer: TokenizerOpts {
+                exact_errors: true,
+                ..Default::default()
+            },
+            tree_builder: TreeBuilderOpts {
+                exact_errors: true,
+                ..Default::default()
+            },
+        },
         QualName::new(None, ns!(html), local_name!("div")),
         vec![],
     )
@@ -128,7 +144,13 @@ fn normalize_html(s: &str) -> String {
     let mut ret_val = Vec::new();
     serialize(&mut ret_val, &body, opts)
         .expect("Writing to a string shouldn't fail (expect on OOM)");
[...]

PR #704, opened 2023-08-27
fix: html5ever may hide actual errors

Martin1887:
Fixes: #702.

html5ever could be removed if this change is approved (pending commit in this pull request).

Please, @ollpu, @raphlinus, review this pull request. This change makes more difficult creating tests but the avoiding of false negatives may pay off.


tests/lib.rs
```diff
@@ -38,7 +44,24 @@ pub fn test_markdown_html(input: &str, output: &str, smart_punct: bool, metadata
...
+    assert_eq!(html_standardize(output), html_standardize(&s));
+}
+
+fn html_standardize(s: &str) -> String {
+    s.trim()
+        .replace("<br>", "<br />")
+        .replace("<br/>", "<br />")
+        .replace("<hr>", "<hr />")
+        .replace("<hr/>", "<hr />")
+        .replace('\t', "")
```
ollpu, line 58: This seems to only affect `regression_test_83` where there is

```
| foo | bar  |
→|-----|------|
| baz | alef |
```
expected to render as
```
<p>| foo | bar  |
→|-----|------|
| baz | alef |</p>
```

But isn't that incorrect? The leading whitespace should be trimmed in the output.

tests/lib.rs
```diff
@@ -38,7 +44,24 @@ pub fn test_markdown_html(input: &str, output: &str, smart_punct: bool, metadata
...
+        .replace("    ", "")
```
ollpu, line 59: This could be a little more strict,
```suggestion
        .replace("\n    ", "\n")
```
since if there is indentation at the first line, that is taken care of by `trim()`.

This also eats indentation inside `<pre>` tags, but maybe that's an acceptable compromise.

ollpu (commented):
The standardization logic implemented here seems to be a good balance. It took me a while to understand the intent of the `>\n<` -> `><` replacement (newlines can be added between tags to aid legibility, but spaces around inline tags are not interchangeable with newlines), so maybe that could be clarified somewhere.

Couple comments about the replacements that eat (potentially meaningful) whitespace.

tests/lib.rs
Martin1887: Yes, this seems right.

tests/lib.rs
Martin1887: I think the test was incorrect, now should be right.

raphlinus:
Yes, this seems like it admits more precise tests. I didn't go through all the changes in detail, so am glad that @ollpu seems to have gone through it in more detail. I'd be quite happy for them to give a final approval if they're willing to spend the time.

tests/lib.rs
```diff
@@ -38,7 +44,24 @@ pub fn test_markdown_html(input: &str, output: &str, smart_punct: bool, metadata
...
+    s.trim()
+        .replace("<br>", "<br />")
+        .replace("<br/>", "<br />")
+        .replace("<hr>", "<hr />")
+        .replace("<hr/>", "<hr />")
+        .replace("\n    ", "\n")
+        .replace("  \n", "\n")
+        .replace(" \n", "\n")
+        .replace("\n  ", "\n")
+        .replace("\n ", "\n")
```
ollpu, line 62: On closer inspection, I'm also not sure about these. Extra spaces at the end of lines are probably not something we want to keep in tests for no good reason.

- `  \n` -> ` \n` doesn't affect any tests. It could maybe be removed.
- ` \n` -> `\n` might actually highlight a point where we go against the spec.
  In 6.8 Soft line breaks:
  > Spaces at the end of the line and beginning of the next line are removed:
  
  But we preserve the space (there's one after a):
  ```
  $ target/debug/pulldown-cmark -e
  a 
  b
  0..5: Start(Paragraph)
  0..2: Text(Borrowed("a "))
  2..3: SoftBreak
  3..4: Text(Borrowed("b"))
  0..5: End(Paragraph)
  EOF
  ```
  And in 4.8 Paragraphs:
  > The paragraph’s raw content is formed by concatenating the lines and removing initial and final spaces or tabs.
  
  But we don't do that:
  ```
  $ target/debug/pulldown-cmark -e
  a 
  0..3: Start(Paragraph)
  0..2: Text(Borrowed("a "))
  0..3: End(Paragraph)
  EOF
  ```
  (these spaces are also preserved in the HTML output)
- `\n  ` -> `\n` affects one test from the CommonMark spec, [`spec_test_191`](https://spec.commonmark.org/0.30/#example-191). Though it's not explicitly mentioned, [example 183](https://spec.commonmark.org/0.30/#example-183) suggests that indentation on the first line of block HTML should be preserved, as commonmark.js does.
- `\n ` -> `\n` also doesn't affect any tests.

ollpu: Given these, I'm also a little skeptical that `.trim()` is useful. Maybe just `.strip_suffix('\n')`, shouldn't that be enough?

Martin1887: Hm... I will dig into this, but all tests should pass so those issues must be resolved before changing this.

Thanks!

ollpu: Filed issue #715 on this

Martin1887:
Finally!

PR #721, opened 2023-09-27
puncttable.rs -> unicode 15.1.0 (+mk_puncttable.py -> python3)

caldwell:
This updates puncttable.rs to unicode 15.1.0.

To do this I updated the python script to python3 (since I don't have python2 installed).

I also adjust it slightly so that it matches rustfmt (which got run against the puncttable.rs file at some point).

The python3 stuff could be a separate PR since it's technically not really related.


Martin1887:
I see this right. What do you think, @raphlinus?

Thanks!

raphlinus (approved):
Yes, it looks good to me as well. Thanks! And it certainly makes sense to update to python3, that shows the age of the origin of this crate.

PR #722, opened 2023-09-27
puncttable: Use hex in PUNCT_TAB and binary in PUNCT_MASKS

caldwell:
This requires #721. It's only adds 2 more small patches. I'm not sure you'll find it useful, but I did as a third party reading the source and trying to understand how the puncttable stuff worked. 

Looking at the `PUNCT_TAB` data, it wasn't clicking because I wasn't visualizing the relationship between the decimal number in the array the hex unicode range in the comments. Turning it to hex makes it clear that it's the range just shifted by a nibble.

Similarly, I think the `PUNCT_MASKS` is bit clearer when it's binary. I found it pleasing have the 16 bits right there visually hammering home that each one represents a unicode codepoint.


Martin1887:
This also seems right, but @raphlinus has a better perspective about this.

Thanks!

raphlinus (approved):
I don't have a strong feeling about this, but I see the value. It makes the byte weight slightly larger, and I personally find hex almost as readable as binary for masks (I know this is unusual), but the readability improvement outweighs it.

So I'm fine with it being merged, but would also respect a decision to keep changes minimal - here there's no improvement in user visible behavior to justify it.

Martin1887:
I see, so we can wait for other discussions here before merging it.

Thanks!

Martin1887:
A long time without discussions about this, so I close this pull request until anyone suggests otherwise. Thanks anyway for your contribution.

PR #734, opened 2023-10-19
Yet another Math implementation

notriddle:
This feature is loosely based on what https://github.com/raphlinus/pulldown-cmark/commit/63a29a1d4a7981769d3f2b9655b63a099426d121 #733 described, but copies [commonmark-hs] more closely (the balanced braces feature is added).

[commonmark-hs]: https://github.com/nschloe/github-math-bugs

It largely ignores GitHub, because its math parsing [is very buggy].

[is very buggy]: https://github.com/nschloe/github-math-bugs


src/parse.rs
```diff
@@ -290,6 +295,64 @@ impl<'input, 'callback> Parser<'input, 'callback> {
...
+                        if let Some((scan_ix, true)) = math_delims.find(cur_ix, search_count, brace_context, is_display) {
+                            self.make_math_span(cur_ix, scan_ix);
+                        } else {
+                            self.tree[cur_ix].item.body = ItemBody::Text;
+                        }
+                    } else {
+                        // we haven't previously scanned all codeblock delimiters,
+                        // so walk the AST
+                        let mut scan;
+                        'search_count: loop {
```
ollpu, line 323: The second time this loops around, I think `math_delims` should already be used.  Otherwise this has quadratic time complexity in a case like `$$x$}$$x$}$$x$}...`

```
$ python -c 'print("$$x$}"*10000)' | time target/release/pulldown-cmark -M > /dev/null
7.53user 0.00system 0:07.62elapsed 98%CPU (0avgtext+0avgdata 6100maxresident)k
0inputs+0outputs (0major+1291minor)pagefaults 0swaps
$ python -c 'print("$$x$}"*20000)' | time target/release/pulldown-cmark -M > /dev/null
31.44user 0.01system 0:31.56elapsed 99%CPU (0avgtext+0avgdata 9760maxresident)k
0inputs+0outputs (0major+2507minor)pagefaults 0swaps
```

ollpu (commented):
This looks fairly solid! I dismissed the curly brace matching feature initially because it seemed too complex to implement compared to its utility. Your approach here makes it look quite straightforward though.

In my implementation (rhysd#1), I opted to emit each dollar as its own item in the first pass, instead of eagerly combining pairs as display tags. This seemed to make parsing simpler. It avoids the "parity avalanche" where a closing inline tag requires all subsequent dollars in the same run to be reconsidered. Our implementations differ in these corner cases:

```
$x$$$$$
yours: <p><span class="math inline">x</span>$$$$</p>
mine:  <p><span class="math inline">x</span><span class="math display"></span></p>

$x$$$$$$$y$$
yours: <p><span class="math inline">x</span>$<span class="math display"></span>
          <span class="math inline">y</span>$</p>
mine:  <p><span class="math inline">x</span><span class="math display"></span>
          <span class="math display">y</span></p>
```

Although I've just noticed commonmark-hs does not accept empty display math either, so it disagrees with both of us:

```
$x$$$$$$$y$$
<p><span class="math inline">x</span>$$$<span class="math display">$y</span></p>
```

src/firstpass.rs
```diff
@@ -636,6 +646,70 @@ impl<'a, 'b> FirstPass<'a, 'b> {
...
+                    let can_close = count > 1 || !self.text[..ix].as_bytes().last().copied().map_or(true, is_ascii_whitespace);
+
+                    // 0xFFFF_FFFF... represents the root brace context. Using None would require
+                    // storing Option<usize>, which is bigger than usize.
+                    //
+                    // These can't conflict, because that would require storing usize::MAX
+                    // bytes in memory, which is impossible.
+                    //
+                    // Unbalanced braces will cause the root to be changed, which is why it gets
+                    // stored here.
```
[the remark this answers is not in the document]
ollpu: I suppose it doesn't really matter, could also be the index of the current `$` being processed.

src/parse.rs
```diff
@@ -290,6 +295,68 @@ impl<'input, 'callback> Parser<'input, 'callback> {
...
+                    });
+                    let mut search_count = if is_display { 2 } else { 1 };
+                    if !can_open {
+                        self.tree[cur_ix].item.body = ItemBody::Text;
+                        prev = cur;
+                        cur = self.tree[cur_ix].next;
+                        continue;
+                    }
+
+                    let result = 'result: loop {
```
ollpu, line 310: This loop should not be necessary with the unmerged `MaybeMath` approach. The search for matching inline math is handled when we come around to the second `$` (or final one of a run) in the main `handle_inline_pass1` loop.

It might also simplify things to store a bool e.g. `followed_by_dollar` in `MaybeMath`. Then you don't need to separately check the next item from the AST to verify it is display mode (other than at the last step where all items have to be erased). `MaybeEmphasis` etc use a similar idea: each char emits its own item, but also records how many of the same char follow. Up to you.

ollpu: Agreed. The idea is that if the current $ is followed by another $, inline math _cannot_ start from that point. So we recognize that, and try to find a closing _display_ tag. Only if that fails, we move to the latter $ and search for a closing inline tag.

```
$$x$
^
try to find the next $$. none found, move on

$$x$
 ^
try to find the next $. it can_close, so inline math is recognized
```

The spec test could be better, yeah.

src/firstpass.rs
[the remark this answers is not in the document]
ollpu: I guess having a limit is good either way. However, it's a little suspicious to me that an order of 5000 hashmap accesses would take multiple seconds. Was there still quadratic behavior hidden somewhere?

src/parse.rs
```diff
@@ -290,6 +295,68 @@ impl<'input, 'callback> Parser<'input, 'callback> {
...
+                        continue;
+                    }
+                    let result = if math_delims.is_populated() {
+                        // we have previously scanned all math environment delimiters,
+                        // so we can reuse that work
+                        if let Some((scan_ix, can_close)) = math_delims.find(cur_ix, is_display, brace_context) {
+                            // can_close only applies to inline math
+                            // block math can always close
+                            if self.tree[cur_ix].next != Some(scan_ix) && (can_close || is_display) {
+                               Some(scan_ix)
```
rhysd, line 315: nit: indentation

ollpu:
Heads up: I changed #733 to also require display math to be non-empty. The behavior I described for escaped pipes in tables should be the same as in your implementation as far as I understand.

src/main.rs
```diff
@@ -58,6 +58,7 @@ pub fn main() -> std::io::Result<()> {
     opts.optflag("d", "dry-run", "dry run, produce no output");
     opts.optflag("e", "events", "print event sequence instead of rendering");
     opts.optflag("T", "enable-tables", "enable GitHub-style tables");
+    opts.optflag("M", "enable-math", "enable LaTeX-style math");
```
ollpu, line 61: This clashes with `enable-metadata-blocks`, defined below. Running out of letters :(

ollpu:
> I've rebased this PR onto the current mainline, and started fuzz testing it against commonmark-hs (and found a couple of problems).

Great! I'll clean up the spec PR and add brace handling to it hopefully next week, so we can move forward with this.

Martin1887:
Hi!, and sorry for the long time without a response about this.

I still have to review the code, but I have reviewed the possible specs and #733 and this are the way to go in my opinion.

The only thing I would change is the target of the pull request, so for major changes it is better to use the branch https://github.com/pulldown-cmark/pulldown-cmark/tree/branch_0.11 I created last week.

I will review the code of both pull requests and this will be merged soon, finally! The spec is mandatory in the crate docs though, I guess that we can convert #733 complete description into RustDoc comments and tests spec.

Thanks to both for this great work!

Martin1887:
The only thing in the spec I would change is using `math math-inline` and `math math-display` classes instead `math inline` and `math display`, to avoid possible collisions with user-defined or CSS framework classes.

Martin1887 (changes requested):
Some minor comments. Thanks for the great work!

pulldown-cmark/src/parse.rs
```diff
@@ -62,6 +62,8 @@ pub(crate) enum ItemBody {
 
     // repeats, can_open, can_close
     MaybeEmphasis(usize, bool, bool),
+    // is_display, can_open, can_close, brace context
```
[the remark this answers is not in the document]
ollpu: It's tempting to replicate the style of the surrounding code here. Maybe we should change all of them to use named parameters.

ollpu:
@Martin1887 
> The spec is mandatory in the crate docs though, I guess that we can convert #733 complete description into RustDoc comments.

So are you thinking this would be a brief explainer in an `Options::ENABLE_MATH` doc comment or something else / more elaborate?

Currently the `math.txt` spec is kind of long and it doesn't function very well as an explainer. It could use some rearranging (clearly separate the pathological test cases to the end) and a pass to make sure that the important parts are concise and understandable.

I could work on these – maybe in a new PR after this is merged.

pulldown-cmark/specs/math.txt
```diff
@@ -0,0 +1,858 @@
+Run this with `cargo run -- -m -s specs/math.txt`.
+
+## Haskell-CommonMark-Extensions (Pandoc commonmark_x) parity tests
```
ollpu, line 3: What does this header refer to? Including the respective spec from commonmark-hs plus maybe some extra parity tests sounds like a good idea, but it shouldn't be at the beginning of the file.

Martin1887:
> @Martin1887
> 
> > The spec is mandatory in the crate docs though, I guess that we can convert #733 complete description into RustDoc comments.
> 
> So are you thinking this would be a brief explainer in an `Options::ENABLE_MATH` doc comment or something else / more elaborate?
> 
> Currently the `math.txt` spec is kind of long and it's hard to read as a guide. It could use some rearranging (clearly separate the pathological test cases to the end) and a pass to make sure that the important parts are concise and understandable.
> 
> I could work on these – maybe in a new PR after this is merged.

I think the best option is writing a small summary with a link to the full documentation, and creating a mdBook with all specs and other tips and guides, as the issue #863 I created yesterday. Thanks!

pulldown-cmark/specs/math.txt
Martin1887: I'm also not sure if the first line is correct, since other specs say 'Run this with `cargo test --features gen-tests ...`.'

It would be better having coherence here (I has also noticed that my specs `metadata_blocks.txt` and `heading_attrs.txt` do not have this line, adding it would be a good point).

pulldown-cmark/src/parse.rs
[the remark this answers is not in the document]
Martin1887: I have tried to change that but sadly it makes dirtier other parts of the code, where it is clear the meaning of each field because there they have names, so I have just removed the first wrong field of the comment.

Martin1887 (approved):

Martin1887:
I think that to merge this pull request and to close the math issue there are only two pending tasks:

- Making clear the heading of the spec in this pull request.
- Updating the spec pull request and search a place for it. Maybe we can create the mdBook now although it is not completed yet, many things can be already included.

pulldown-cmark/specs/math.txt
ollpu: Good spot, that is the outdated command which I corrected at some point in my version of the spec.

ollpu:
One remaining divergence between #733 and this is that this PR accepts empty display math items:

```
$ cargo run -- -m
$$$$
<p><span class="math math-display"></span></p>
```

I changed this to be disallowed in https://github.com/pulldown-cmark/pulldown-cmark/pull/733/commits/d583a8d3690f72700ed3a37289e856d570a0c7dd to align with commonmark-hs.

Otherwise I think #733 can be closed in favor of this. No need to merge both.

Martin1887:
I think that the high level description of #733 can be a good start point for the math spec in the documentation, but as it is true it will not be merged, I have closed it. Thanks!

Martin1887:
So let's merge it. Thanks!

ollpu:
Bug:
```
$ cargo run -- -m
$x {$ $ } $x$
<p>$x {$ $ } $x$</p>
```
I'm working on a small refactor which should fix this as well.

PR #750, opened 2023-10-27
Parse `[^]` as a link

notriddle:
This is consistent with most other CommonMark parsers, even when they have support for footnotes turned on.


Martin1887:
A bit weird that Pandoc flavor handles it differently, but closer to CommonMark is better. Thanks!

PR #771, opened 2023-11-02
Parse inline nested footnote definitions

### LINKED ISSUE #770: [fuzz results] inline nested footnote defs don't work
This markdown:

```markdown
[^foo] [^bar]

[^foo]: [^bar]: 1
```

should work like this:

[^foo] [^bar]

[^foo]: [^bar]: 1

<details>

Events from pulldown-cmark:

```rust
"-*-FD\n[^+\\&~+ppppppppppp]:[^+\\&~+]:1          +\n\n*1    ppp]:[^+\\&~+]:*    D\n" -> [
  Start(Paragraph)
    Text(Boxed("-*-FD"))
  End(Paragraph)
  Start(Paragraph)
    Start(Emphasis)
      Text(Boxed("1 ppp]:[^+&~+]:"))
    End(Emphasis)
    Text(Boxed(" D"))
  End(Paragraph)
]
```

Events from pandoc:

```rust
"-*-FD\n[^+\\&~+ppppppppppp]:[^+\\&~+]:1          +\n\n*1    ppp]:[^+\\&~+]:*    D\n" -> [
  Start(Paragraph)
    Text(Boxed("-*-FD"))
  End(Paragraph)
  Start(Paragraph)
    Start(Emphasis)
      Text(Boxed("1 ppp]:"))
      FootnoteReference(Boxed("1"))
      Start(FootnoteDefinition(Boxed("1")))
        Start(Paragraph)
          Text(Inlined(InlineStr { inner: [49, 32, 43, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0], len: 3 }))
        End(Paragraph)
      End(FootnoteDefinition)
      Text(Boxed(":"))
    End(Emphasis)
    Text(Boxed(" D"))
  End(Paragraph)
]
```

</details>

notriddle:
Fixes #770

They look like this:

    [^foo]: [^bar]: test

They don't make much sense (foo is empty), but everyone else parses them correctly, so we should too.

### DIFF
--- a/tests/suite/regression.rs
+++ b/tests/suite/regression.rs
@@ -1889,12 +1889,57 @@ fn regression_test_120() {
 
 #[test]
 fn regression_test_121() {
-    let original = r##"[First try
-----------
-Second try]: https://rust-lang.org
+    let original = r##"The second hyphen should parse the same way in both samples.
+
+ - >*
+
+   -
+
+The second hyphen should parse the same way in both samples.
+
+ - >x
+
+   -
 "##;
-    let expected = r##"<h2>[First try</h2>
-<p>Second try]: https://rust-lang.org</p>
+    let expected = r##"<p>The second hyphen should parse the same way in both samples.</p>
+<ul>
+<li>
+<blockquote>
+<ul>
+<li></li>
+</ul>
+</blockquote>
+<ul>
+<li></li>
+</ul>
+</li>
+</ul>
+<p>The second hyphen should parse the same way in both samples.</p>
+<ul>
+<li>
+<blockquote>
+<p>x</p>
+</blockquote>
+<ul>
+<li></li>
+</ul>
+</li>
+</ul>
+"##;
+
+    test_markdown_html(original, expected, false, false, false);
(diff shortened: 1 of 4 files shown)


tests/suite/regression.rs
```diff
@@ -1889,12 +1889,57 @@ fn regression_test_120() {
 
 #[test]
 fn regression_test_121() {
```
Martin1887, line 1891: This commit seems to remove a test from the `master` branch. Since this commit should not modify any regression test, I think the cleanest solution is to remove `regression.rs` from the commit, and new changes will be merged in their correspondent issues. Thanks!

Martin1887 (changes requested):

tests/suite/regression.rs
Martin1887: Finally a conflict generated after merging another pull request has fixed the issue, thanks for your work.

Martin1887 (approved):

PR #776, opened 2023-11-03
Fix up corner cases in table and code unescaping

### LINKED ISSUE #716: Inline code loses backslashes on subsequent lines
Due to the way `make_code_span` only starts buffering after the first line, its behavior can be different on subsequent lines. Escaped punctuation doesn't cover the backslash in any item span, so it gets lost.

```
$ cargo run -- -e
`\!\"\#\$\%\&
\!\"\#\$\%\&`
0..28: Start(Paragraph)
0..27: Code(Boxed("\\!\\\"\\#\\$\\%\\& !\"#$%&"))
0..28: End(Paragraph)
EOF
```

Not sure if there are any other items emitted by the first pass whose spans don't cover all of the input.

notriddle:
Fixes #716


Martin1887:
This feels right to me, but it is a pretty big change and I would like the approval of @raphlinus.

This was a blocking issue for the 0.10 version, so it is closer. Thanks!

raphlinus (approved):
This is a dauntingly large change for a relatively edge-casey bug, but I can't quickly see a better way to fix it. I'm satisfied with the degree of testing.

Note, the merge conflicts will need to be fixed before merging.

PR #777, opened 2023-11-06
Correct instructions for running specs

ollpu:
Running specs with `cargo run -- -s` hasn't worked since 22aea4720f8c9b2ad78b584e710e35e1c5a21e48. Fix the instructions to work with the current harness.


Martin1887:
Good catch, thanks!

PR #785, opened 2023-11-24
Trim spaces off end of parsed lines

### LINKED ISSUE #715: Whitespace before EOL sometimes incorrectly preserved
In the CommonMark spec, 4.8 Paragraphs:

> The paragraph’s raw content is formed by concatenating the lines and **removing** initial and **final spaces or tabs**.

and in 6.8 Soft line breaks:

> Spaces at the end of the line and beginning of the next line are removed:

Some test cases for these (atop `tests/html.rs`):

```rust
// fails
#[test]
fn trim_space_and_tab_at_end_of_paragraph() {
    let original = "one\ntwo \t";
    let expected = "<p>one\ntwo</p>\n";

    let mut s = String::new();
    html::push_html(&mut s, Parser::new(&original));
    assert_eq!(expected, s);
}

// passes
#[test]
fn trim_space_tab_nl_at_end_of_paragraph() {
    let original = "one\ntwo \t\n";
    let expected = "<p>one\ntwo</p>\n";

    let mut s = String::new();
    html::push_html(&mut s, Parser::new(&original));
    assert_eq!(expected, s);
}

// fails
#[test]
fn trim_space_nl_at_end_of_paragraph() {
    let original = "one\ntwo \n";
    let expected = "<p>one\ntwo</p>\n";

    let mut s = String::new();
    html::push_html(&mut s, Parser::new(&original));
    assert_eq!(expected, s);
}

// fails
#[test]
fn trim_space_before_soft_break() {
    let original = "one \ntwo";
    let expected = "<p>one\ntwo</p>\n";

    let mut s = String::new();
    html::push_html(&mut s, Parser::new(&original));
    assert_eq!(expected, s);
}
```

I tried fixing this, but it didn't turn out so simple. `trim_space_tab_nl_at_end_of_paragraph` seems to pass because hard breaks are correctly trimmed out. However, no hard break is recognized if the input doesn't have a trailing newline. So that case will at least require some kind of special handling or modification to `parse_line`.

notriddle:
Fixes #715

### DIFF
--- a/src/firstpass.rs
+++ b/src/firstpass.rs
@@ -333,11 +333,6 @@ impl<'a, 'b> FirstPass<'a, 'b> {
             self.tree.push();
             let (next_ix, _brk) = self.parse_line(ix, None, TableParseMode::Active);
 
-            if let Some(cur_ix) = self.tree.cur() {
-                let trailing_whitespace = scan_rev_while(&bytes[..next_ix], is_ascii_whitespace);
-                self.tree[cur_ix].item.end -= trailing_whitespace;
-            }
-
             self.tree[cell_ix].item.end = next_ix;
             self.tree.pop();
 
@@ -644,7 +639,7 @@ impl<'a, 'b> FirstPass<'a, 'b> {
                         );
                     }
 
-                    self.tree.append_text(begin_text, ix);
+                    self.tree.append_text(begin_text, ix - trailing_whitespace);
                     LoopInstruction::BreakAtWith(
                         end_ix,
                         Some(Item {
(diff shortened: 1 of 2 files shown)


Martin1887:
Very interesting, this was a pending issue for the 0.10 release, thanks! I have to carefully check the changes, and @raphlinus review would also be very appreciated.

raphlinus (approved):
I haven't gone deep into the logic, but this seems fairly straightforward - the existing behavior is (subtly) buggy, there are tests to cover that, and the author is someone very familiar with the codebase. I'm comfortable with it going in, and if there are problems, hopefully they'll surface in time. Thanks for the work on this!

Martin1887:
Perfect, thanks!

PR #786, opened 2023-11-25
Fix parsing of spaces in code

### LINKED ISSUE #661: Parsing inline code with `"\n "`  loses both newline and space
Events from pulldown-cmark:

```rust
"`\n `f " -> [
  Start(Paragraph)
    Code(Borrowed(""))
    Text(Borrowed("f"))
  End(Paragraph)
]
```

Events from commonmark.js:

```rust
"`\n `f " -> [
  Start(Paragraph)
    Code(Boxed(" "))
    Text(Boxed("f "))
  End(Paragraph)
]
```

XML from [commonmark.js](https://spec.commonmark.org/dingus/?text=%60%0A%20%60f):

```xml
  <paragraph>
    <code> </code>
    <text>f</text>
  </paragraph>
</document>
```

This might be the same as #655, but here it looks like ``"`\n `"`` is turned into `""`, which means that both the space and the newline got lost.

notriddle:
Fixes #661

The changed case in regression.txt actually aligns our behavior with [commonmark.js](https://spec.commonmark.org/dingus/?text=%3E%20%60%0A%3E%20%60&smart=1). Check the "HTML" tab.

This change simplifies everything by having the code span maker read the text directly, instead of deciphering the tree with all the emphasis data and trimmed spaces (that still gets used if it later finds out that MaybeCode isn't code).

### DIFF
--- a/specs/regression.txt
+++ b/specs/regression.txt
@@ -450,7 +450,7 @@ EMPTY CODE IN BLOCKQUOTE
 > `
 .
 <blockquote>
-<p><code></code></p>
+<p><code> </code></p>
 </blockquote>
 ````````````````````````````````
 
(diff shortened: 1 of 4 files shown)


Martin1887:
I understand the same, it is probably a corner-case bug of the reference implementation.

PR #799, opened 2023-12-10
cargo zigbuild

phoepsilonix:
added workflow for binary builds.


Martin1887:
This is interesting, but I wonder if there is another easier way of doing it without installing zig, it seems a bit over-complicated to me.

Thanks!

Makefile
```diff
@@ -0,0 +1,38 @@
...
+test:
+	@cargo +nightly test
+
+.PHONY: shellspec
+shellspec:
+	@shellspec
+
+.PHONY: install
+install:
+	@cargo +nightly install --path .
```
alerque, line 21: I'm not quite sure what the goal is there but none of this smell right. As someone packaging this for distros and using it myself, this whole pile of changes would just confuse me about how the project was expected to be used. It doesn't need `+nightly`. This Makefile is a thin wrapper around other commands and doesn't support any of the tooling I'd expect an actual Makefile to handle. Most of this feels more like it should go in a Justfile or something for developer jobs, but most of it isn't even build stuff, and the bits that are don't seem to be general purpose.

PR #801, opened 2023-12-16
Add text generation via `text` feature.

trojanfoe:
This PR provides basic plain-text generation via a new `text` feature, which is disabled by default.

I added this as I needed to send e-mail content which is best sent as a multipart message containing both HTML and plain text and markdown seemed like the best format for this.

One issue with generating multiple strings from the same markdown input is that it's necessary to duplicate the parser as the iterator consumes the input:

```
let text_content = {
    let pulldown_parser = Parser::new(&markdown_content);
    let mut text_content = String::new();
    text::push_text(&mut text_content, pulldown_parser);
    text_content
};

let html_content = {
    let pulldown_parser = Parser::new(&markdown_content);
    let mut html_content = String::new();
    html::push_html(&mut html_content, pulldown_parser);
    html_content
};
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
-target
+target/
 
+.idea/
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -516,7 +516,7 @@ dependencies = [
 
 [[package]]
 name = "pulldown-cmark"
-version = "0.9.2"
+version = "0.10.0"
 dependencies = [
  "bincode",
  "bitflags 2.4.1",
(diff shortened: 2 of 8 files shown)


Martin1887:
Interesting, but I have to dig more into it before merging. What do you think about this feature, @raphlinus?

Thanks!

trojanfoe:
I will say it’s not particularly well implemented. It works for my use-case but it needs a decent review and probably more work.

rimutaka:
Additional use cases:
- extracting the content to generate a summary
- generating excerpts
- supplying context to LLMs

I could make use of it right now if available.

Martin1887:
It could be useful, for sure. but we need a good implementation before merging.

PR #804, opened 2024-01-08
add rstest as dev dependency; use it to recreate several parser regression tests

bobhy:
This PR adds package rstest as a dev dependency, which makes it easy to write lots of test cases for a given test scenario.  It also re-implements several parser tests using rstest style, as a proof of concept.  No additional functionality (or test coverage) is intended.

I'm working on an actual feature (around GFM task lists) that wants to use this style, but thought I'd factor out the rstest dependency for separate review.

This PR also has some extraneous code formatting changes, thanks to `cargo fmt`, which my IDE ran automatically. I hope this won't be an annoyance/distraction...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 target
 
+.vscode
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -216,6 +216,101 @@ dependencies = [
 ]
 
 [[package]]
+name = "futures"
+version = "0.3.30"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "645c6916888f6cb6350d2550b80fb63e734897a8498abe35cfb732b6487804b0"
+dependencies = [
+ "futures-channel",
+ "futures-core",
+ "futures-executor",
+ "futures-io",
+ "futures-sink",
+ "futures-task",
+ "futures-util",
+]
+
+[[package]]
+name = "futures-channel"
+version = "0.3.30"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "eac8f7d7865dcb88bd4373ab671c8cf4508703796caa2b1985a9ca867b3fcb78"
+dependencies = [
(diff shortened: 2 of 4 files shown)


Martin1887:
I think the current testing infrastructure is good and I cannot see any value in your reimplementation using rstest (sorry, but I find it more difficult to read).

However, discussion about this topic is welcome, and thank you for your interest in contributing to pulldown-cmark!

bobhy:
My point in adding rstest was that I wanted to use it in subsequent PRs I might submit, not to force other devs to use it  

But I do think it's a good tool, proven in many other projects and that it has advantages to repay the cost of learning (as with everything in Rust).  

Rstest makes it easy to factor out common test code, so you have only one implementation of a fixture or test scenario to vet (or to update if you subsequently decide you were doing the test wrong). 

In this case, you have a set of tests that all use the same general pattern: Parse a markdown input expecting the parser to not crash and to produce an HTML output; verify the actual matches the expected output.  You can see by inspection that all the test cases will run exactly the same test scenario.  (I actually added an expected output for some of the "don't crash" tests, on the theory that we *should* care what the parser produces in these cases because that's what the user  is going to see and so becomes an implicit part of the contract).

Rstest also makes it easy to inventory the test cases you have for a given test scenario, so you can decide whether you've probed enough and the right kinds of points in the input space.   With testing, even too much is never enough.

In this case, you can see at a glance that this scenario doesn't test non-ascii Unicode at all and doesn't cover the empty string input.  So if you're curious about those things, you know you have to look elsewhere.

Martin1887:
I see this is true in general cases, but have you looked the tests structure of this project?

We don't write tests code in Rust, but using spec files, which are more readable in my opinion than the rstests of this pull request, and also they are already highly integrated in the project and they allow to easily import other Markdown specifications. Rust test code files are automatically created when running `cargo test --all-features`.

Please check `specs/regression.txt` as an example on how the tests are created in this project.

PR #807, opened 2024-01-09
Add reject-broken-links CLI option

### LINKED ISSUE #788: CLI should have a `--reject-broken-links` option
When running the pulldown-cmark binary, there should be a `--reject-broken-links` option, so that if you try to render a file with broken links, the program outputs an error message and exits with an error code.

(I can implement it in a PR if you're interested)

PoignardAzur:
This also removes the "mem::forget at the end of the program" trick, both to simplify lifetime handling and because it's less applicable now that the 
pulldown_cmark function might be called several times.

Fixes #788

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -34,11 +34,7 @@ fn dry_run(text: &str, opts: Options, broken_links: &mut Vec<BrokenLink<'static>
         text,
         opts,
         Some(|link: BrokenLink<'_>| {
-            broken_links.push(BrokenLink {
-                span: link.span,
-                link_type: link.link_type,
-                reference: link.reference.into_string().into(),
-            });
+            broken_links.push(link.into_static());
             None
         }),
     );
@@ -51,11 +47,7 @@ fn print_events(text: &str, opts: Options, broken_links: &mut Vec<BrokenLink<'st
         text,
         opts,
         Some(|link: BrokenLink<'_>| {
-            broken_links.push(BrokenLink {
-                span: link.span,
-                link_type: link.link_type,
-                reference: link.reference.into_string().into(),
-            });
+            broken_links.push(link.into_static());
             None
         }),
     )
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -113,43 +140,60 @@ pub fn main() -> std::io::Result<()> {
...
     Ok(())
 }
 
-pub fn pulldown_cmark(input: &str, opts: Options) {
-    let mut p = Parser::new_ext(input, opts);
+pub fn pulldown_cmark<'i>(input: &str, opts: Options, broken_links: &mut Vec<BrokenLink<'static>>) {
+    let mut p = Parser::new_with_broken_link_callback(
+        input,
+        opts,
+        Some(|link: BrokenLink<'_>| {
```
Martin1887, line 187: This closure is written exactly equal three times, violating the DRY principle. Please try to fix this code duplication.

Martin1887 (changes requested):

src/main.rs
PoignardAzur: Each closure is a single instruction where values are piped as-is, with the only non-trivial transformation being the `reference.into_string().into()` part to clone the string.

I don't think factoring out the closure into a function would reduce semantic duplication, and it would make the code significantly more complex lifetime-wise.

PoignardAzur: I guess I could add a `to_static` method to BrokenLink, which would reduce each closure to a single line, which would be less duplicate-y.

PoignardAzur: Done.

Martin1887: Yeah, it looks cleaner now! But cannot the final `None` lines be removed?

PoignardAzur: Short version: not without making other code hard to read.

Martin1887: OK, then ready to be merged. Thanks!

Martin1887 (approved):

PoignardAzur:
Whew!! \o/

PR #811, opened 2024-01-15
Calculate correct list indent for ordered lists

### LINKED ISSUE #810: [fuzz result] off-by-one indent numbered list item with empty first line
When a numbered list has a blank first line in its item, the line after it is supposed to be indented to match the imaginary space after it.

For examples:

```markdown
1.
      pulldown-cmark thinks this is code, but commonmark.js and commonmark-hs don't.
      The dot in the list marker is at column two, so the child paragraph actually
      starts in column *four* (there's three spaces of indentation built into the list).
```

```markdown
1.
  This is not in the list at all. It's a paragraph after it.
```

<details>

Events from pulldown-cmark:

```rust
"0)\n  (\n" -> [
  Start(List(Some(0)))
    Start(Item)
      Text(Borrowed("("))
    End(Item)
  End(List(true))
]
```

Events from pandoc:

```rust
"0)\n  (\n" -> [
  Start(List(Some(0)))
    Start(Item)
    End(Item)
  End(List(true))
  Start(Paragraph)
    Text(Boxed("("))
  End(Paragraph)
]
```

Events from commonmark.js:

```rust
"0)\n  (\n" -> [
  Start(List(Some(0)))
    Start(Item)
    End(Item)
  End(List(true))
  Start(Paragraph)
    Text(Boxed("("))
  End(Paragraph)
]
```

</details>

notriddle:
Fixes #810

### DIFF
--- a/specs/regression.txt
+++ b/specs/regression.txt
@@ -2116,3 +2116,28 @@ baz`</li>
 baz](https://example.com)</li>
 </ul>
 ````````````````````````````````
+
+ISSUE 810
+
+```````````````````````````````` example
+1.
+      pulldown-cmark used to think this was code, but commonmark.js and commonmark-hs didn't.
+      The dot in the list marker is at column two, so the child paragraph actually
+      starts in column *three*.
+.
+<ol>
+<li>pulldown-cmark used to think this was code, but commonmark.js and commonmark-hs didn't.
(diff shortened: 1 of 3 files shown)


Martin1887:
Thanks!

PR #694, opened 2023-07-18
chore: Update bitflags from 1.3.x to 2.x

waywardmonkeys:

PR #825, opened 2024-01-26
Allow footnote definitions to not be separated by blank lines

GuillaumeGomez:
This is a change allowing to fix https://github.com/rust-lang/rust/issues/100638.

It's a bit strange that link definitions can follow each others but that footnote definitions can't so I think for consistency, allowing it would be more logical.

As for the code, it's not great as it forces to "look ahead" in case this is a footnote definition and then return and come back into `parse_block` to actually put this new footnote definition into the tree. If you have suggestions for improving it, it'd be very welcome! :)

### DIFF
--- a/examples/parser-map-event-print.rs
+++ b/examples/parser-map-event-print.rs
@@ -1,15 +1,17 @@
-use pulldown_cmark::{Event, Parser, html};
+use pulldown_cmark::{html, Event, Parser};
 
 fn main() {
     let markdown_input = "# Example Heading\nExample paragraph with **lorem** _ipsum_ text.";
-    println!("\nParsing the following markdown string:\n{}\n", markdown_input);
+    println!(
+        "\nParsing the following markdown string:\n{}\n",
+        markdown_input
+    );
 
-    // Set up the parser. We can treat is as any other iterator. 
+    // Set up the parser. We can treat is as any other iterator.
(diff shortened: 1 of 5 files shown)


Martin1887:
Thanks for your work. Footnotes are not included in the GFM spec but in the footnotes example of its guide to Markdown consecutive footnotes work.

Please not this pull request should be created against the `master` branch.

Also note that tests must be written into spec `.txt` files, and `.rs` files are created automatically when running `cargo test --all-features`. Also note the failed test in the CI, very probably caused because the `.rs` file has been regenerated from the outdated `.txt` spec.

GuillaumeGomez:
Isn't the `master` branch an API breaking change? Because if so, I'd still like to have this fix merged into a `0.9.4` release so I can update rustdoc dependency and fix the bug directly there. I can backport this change to the `master` branch though.

I'll try to see how to make the tests work.

GuillaumeGomez:
CI passed. \o/

Martin1887:
OK, 0.10 version will be released soon, and it contains many bugfixes and enhancements, but also some minor breaking changes. So, we can generate a 0.9.4 release with this change and #694.

Thanks.

GuillaumeGomez:
Thanks a lot! I'll backport the fix in the next days.

Issue #598, opened 2022-07-03
`escape_html`: Single quotes not escaped


BenjaminRi:
The HTML escape table does not escape single quotes (`'`): https://github.com/raphlinus/pulldown-cmark/blob/27f7cd588c6f6402c1cd50a000dd6f73452bd34e/src/escape.rs#L140

 However, the HTML spec explicitly allows these single quotes for HTML attributes as well: https://www.w3.org/TR/html4/intro/sgmltut.html#h-3.2.2

This creates potential security issues when escaping HTML sequences used in attributes. It seems like this library uses double quotes (`"`) everywhere but whenever `escape_html` is used for attributes wrapped in single quotes (`'`), e.g. by a user of this library, an attacker can break out and append HTML code. So it would probably be prudent to escape it as well, with `&#39;` or `&#x27;` (which is the same).

PR #830, opened 2024-01-27
feat(escape): separate attribute and text HTML

notriddle:
This adds a function to escape attribute text, which will escape both single and double quotes, and another function used only for text nodes that doesn't.

This is an alternative to https://github.com/raphlinus/pulldown-cmark/pull/823

Fixes https://github.com/raphlinus/pulldown-cmark/issues/598


Martin1887:
This is better than #823, thanks!

PR #831, opened 2024-01-27
Set range end to Paragraph if returned early because of footnote definition

GuillaumeGomez:
My deepest apologies, I didn't notice but the end range was not filled and it panicked when running rustdoc on the code I used for the test. I fixed it and added a regression test to ensure the range is always filled as expected.

Can you yank the `0.9.4` version and release the `0.9.5` please? Again, I'm very sorry for not noticing the bug earlier on.

### DIFF
--- a/src/firstpass.rs
+++ b/src/firstpass.rs
@@ -321,6 +321,7 @@ impl<'a, 'b> FirstPass<'a, 'b> {
         let mut ix = start_ix;
         loop {
             if self.options.contains(Options::ENABLE_FOOTNOTES) && self.get_footnote(ix).is_some() {
+                self.tree[node_ix].item.end = if ix > start_ix { ix - 1 } else { start_ix };
                 self.tree.pop();
                 if let Some(node_ix) = self.tree.peek_up() {
                     if let ItemBody::FootnoteDefinition(..) = self.tree[node_ix].item.body {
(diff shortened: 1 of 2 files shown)


Martin1887:
No problem, version 0.9.5 has just been released and 0.9.4 is yanked.

GuillaumeGomez:
Thanks a lot!

Martin1887:
0.9.6 just released updating Cargo.lock and 0.9.5 yanked to fix #833.

GuillaumeGomez:
Thanks! :)

Martin1887:
No need of 'frontporting' to 0.10, this was already implemented as the new footnotes in #654.

GuillaumeGomez:
Apparently not when looking at https://github.com/raphlinus/pulldown-cmark/pull/654/files#diff-243abda520cdeea7f9b927714d874cf1ffa167c126d84089cef20d697e97532dR143. This still needs to be fixed.

Martin1887:
It is fixed in the new footnotes type, fully compatible with GitHub, but it does not in the old type. Probably convenient to have the same behavior than 0.9.6...

GuillaumeGomez:
It is not based on what is in https://github.com/raphlinus/pulldown-cmark/blob/master/specs/footnotes.txt.

You can't do (as is, you won't have two footnote definitions):

```md
Lorem ipsum.[^a] [^missing]

[^a]: def1
[^missing]: def2
```

Martin1887:
OK, I had only checked the spec test with two consecutive footnotes without an empty line between them. So this could be applied to the two types of footnotes then.

Thanks!

GuillaumeGomez:
So I'll try to backport my PR as mentioned in the coming days (might be tricky with me needing to prep for FOSDEM) or next week. :)

PR #835, opened 2024-01-30
chore: apply clippy lints

Martin1887:
Fixes: #573


Martin1887:
Minor clippy fixes, this would probably not need approval but in case, @notriddle. Thanks!

PR #837, opened 2024-02-01
Implement missing cell cutoff to stop quadratic results

### LINKED ISSUE #832: Quadratic output size explosion with tables extension
With table extension, one may get awfully long output with input generated by this python one-liner:

``` bash
$ python -c 'N=100; print("x|" * N + "\n" + "-|" * N + "\n" + "x\n" * N)' | pulldown-cmark -T
```

The output size grows roughly as square of the input size. With `N=10000` you still have an input of decent size (~70kB), but the corresponding output is roughly ~900MB (`10000 * 10000 * strlen("<td></td>")`), so it's potentially usable as DoS attack vector.

(In MD4C I solved this by limiting column count of tables to 128. Anything bigger is refused to be recognized as a table. Probably the only alternative would be to stop emitting empty table cells if they're missing in the input, breaking compatibility with GFM.)

notriddle:
Fixes #832

Benchmarks:

- before: https://notriddle.com/benchmark-html-out/pulldown-cmark-table-old/report/index.html

  a line chart, that's kinda wobbly but seems to curve upward, topping around 160ms

- after: https://notriddle.com/benchmark-html-out/pulldown-cmark-table-new/report/index.html

  a line chart that rapidly hits a peak around 7.5ms and stays there

### DIFF
--- a/benches/lib.rs
+++ b/benches/lib.rs
@@ -4,6 +4,25 @@ mod to_html {
     use criterion::{BenchmarkId, Criterion, Throughput};
     use pulldown_cmark::{html, Options, Parser};
 
+    pub fn pathological_missing_table_cells(c: &mut Criterion) {
+        let mut group = c.benchmark_group("    pub fn pathological_missing_table_cells(c: &mut Criterion) {
+            ");
+        let mut buf = String::new();
+        for i in 1..20 {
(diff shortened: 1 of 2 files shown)


Martin1887:
Great, thanks!

Issue #526, opened 2021-05-16
Consider breaking html escaping into subcrate?


alexwennerberg:
I really like the html escaping code for this library:

https://github.com/raphlinus/pulldown-cmark/blob/master/src/escape.rs

I've used it before by copy/pasting the relevant bits, but I was wondering if you'd consider putting it into a sub-crate so that it can be reused by other libraries that need to do html escaping?

Keats:
It seems that every template engine/HTML renderer has their own method, is there an up to date benchmark for them?

alexwennerberg:
Probably this one is fastest just by reading the code, plus the SIMD optimizations. I could put together some benchmarks

Keats:
I was thinking of https://github.com/Kogia-sima/sailfish/tree/master/sailfish/src/runtime/escape that also does fancy things. It would be great to have benchmarks though.

PR #838, opened 2024-02-02
feat: separate escape.rs in another crate

Martin1887:
Fixes: #526.


Martin1887:
Finally!, the last pending issue for the 0.10 release. Please, @notriddle, review it if you have a chance :-).

laurmaedje:
The "Breaking changes" section in the changelog should maybe be updated to mention this. It states that the `escape` module is now optional, which was true between #519 and this, but is a bit misleading now. I was previously using the `pulldown_cmark::escape` module directly and it stopped working after updating to 0.10.

Martin1887:
Right, thanks, it was in new features but it is also a breaking change. It is already updated.

PR #845, opened 2024-02-17
Limit link def expansion

### LINKED ISSUE #844: pathologic input: O(n^2) with link ref. defs
Input generated with this leads to O(n^2) both in time and output size:
```
$ python -c 'N=1000; print("[x]: " + "x" * N + "\n[x]" * N)'
```

(same as https://github.com/mity/md4c/issues/238; cmark seems to limit number of link. ref. defs instantiations in some way so it is not vulnerable.)

notriddle:
Fixes #844

### Before

image

### After

image

### DIFF
--- a/pulldown-cmark/src/parse.rs
+++ b/pulldown-cmark/src/parse.rs
@@ -254,6 +254,78 @@ impl<'input, F: BrokenLinkCallback<'input>> Parser<'input, F> {
         &self.allocs.refdefs
     }
 
+    /// Use a link label to fetch a type, url, and title.
+    ///
+    /// This function enforces the [`link_ref_expansion_limit`].
+    /// If it returns Some, it also consumes some of the fuel.
+    /// If we're out of fuel, it immediately returns None.
+    ///
+    /// The URL and title are found in the [`RefDefs`] map.
+    /// If they're not there, and a callback was provided by the user,
+    /// the [`broken_link_callback`] will be invoked and given the opportunity
+    /// to provide a fallback.
+    ///
+    /// The link type (that's "link" or "image") depends on the usage site, and
+    /// is provided by the caller of this function.
+    /// This function returns a new one because, if it has to invoke a callback
+    /// to find the information, the link type is [mapped to an unknown type].
+    ///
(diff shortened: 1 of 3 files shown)


pulldown-cmark/src/parse.rs
```diff
@@ -225,6 +243,8 @@ impl<'input, F: BrokenLinkCallback<'input>> Parser<'input, F> {
             inline_stack,
             link_stack,
             html_scan_guard,
+            // always allow 100KiB
+            link_ref_expansion_limit: text.len().max(100_000),
```
ollpu, line 247: I guess I'd rather make this `text.len() + 100_000` but replicating cmark is good.

ollpu (dismissed):
Simple enough solution. Looks good!

Similar quadratic growth could happen with the broken link callback, although it's highly context dependent. Should we warn about that in the docs?

ollpu:
> This runs after the callback, so it should cover it, right?

Ah, I see. But running the callback already potentially involves generating long strings, so quadratic work could be done anyway.

I guess the same applies for ref defs: If the url/link are not borrowed from the input, then them being fetched and cloned from the map is already quadratic. An input like

```python
"[x]: " + "x\\." * N + "\n[x]" * N
```

is still quadratic (though only noticeably so with N > 20000 as it's just string cloning).

So the check for whether the limit has been exhausted should be before anything is cloned or the callback is called.

I'm also not sure that we should apply the same limit to the broken link callback. For ref defs, the url & title are included in the document itself once, so the limit will never be hit if each definition is only used once. This is not the case for links expanded by the callback. Of course the 100K minimum size will be hard to reach in practice.

ollpu:
> That would be solved by making `CowStr::Boxed` use an `Arc` instead of a `Box`, right?

Yeah, that could be considered. For this issue though, there's probably a reasonable solution that checks the limit before cloning? Not sure if it's worth blocking this on switching to `Arc`s everywhere.

ollpu (approved):
Looks good now! (Assuming you're ok with the callback using the same limit)

Martin1887:
I have to deep a bit more on this but it looks good, thanks!

PR #858, opened 2024-03-02
Don't exit `scan_attribute` with the ix pointing at block quote

### LINKED ISSUE #857: [fuzz result] false positive inline HTML within blockquote
In this sample doc, there definitely shouldn't be any inline HTML, because inline HTML tags are required to have an `>` at the end. The `>` in the document is part of the block quote, not the HTML:

```markdown
><span
title
>junk
```

Events from pulldown-cmark:

```rust
"><V\nT\n>T" -> [
  Start(BlockQuote)
    Start(Paragraph)
      Html(Boxed("<V\nT"))
      Text(Borrowed("T"))
    End(Paragraph)
  End(BlockQuote)
]
```

Events from commonmark.js:

```rust
"><V\nT\n>T" -> [
  Start(BlockQuote)
    Start(Paragraph)
      Text(Boxed("<V"))
      SoftBreak
      Text(Boxed("T"))
      SoftBreak
      Text(Boxed("T"))
    End(Paragraph)
  End(BlockQuote)
]
```

XML from [commonmark.js](https://spec.commonmark.org/dingus/?text=%3E%3CV%0AT%0A%3ET):

```xml
  <block_quote>
    <paragraph>
      <text>&lt;</text>
      <text>V</text>
      <softbreak />
      <text>T</text>
      <softbreak />
      <text>T</text>
    </paragraph>
  </block_quote>
</document>
```

notriddle:
Fixes #857

That would be ambiguous, because it can't easily be distinguished between the block quote `>` and the end of an HTML tag `>`. If this function bails out because it can't find an `=`, it should point at the spot immedatiely after the attribute name instead.

### DIFF
--- a/pulldown-cmark/specs/regression.txt
+++ b/pulldown-cmark/specs/regression.txt
@@ -2511,3 +2511,17 @@ ISSUE 853
 .
 <p><a href="((()))">linky</a></p>
 ````````````````````````````````
+
+ISSUE 857
+
+```````````````````````````````` example
+><span
+title
+>junk
+.
+<blockquote>
+<p>&lt;span
+title
+junk</p>
+</blockquote>
+````````````````````````````````
--- a/pulldown-cmark/src/scanners.rs
+++ b/pulldown-cmark/src/scanners.rs
@@ -915,18 +915,18 @@ fn scan_attribute(
     buffer_ix: &mut usize,
 ) -> Option<usize> {
     ix += scan_attribute_name(&data[ix..])?;
-    let n_whitespace =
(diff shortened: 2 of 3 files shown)


pulldown-cmark/specs/regression.txt
```diff
@@ -2511,3 +2511,27 @@ ISSUE 853
 .
 <p><a href="((()))">linky</a></p>
```
Martin1887, line 2512: It seems the issue 853 has been duplicated in regression.txt. Thanks!

Martin1887 (changes requested):

Martin1887 (approved):

PR #866, opened 2024-03-16
Adjust documentation on custom attributes

max-heller:
The docs seem to mistakenly show comma-separated custom heading attributes:

https://github.com/pulldown-cmark/pulldown-cmark/blob/4ba4738224a84de7447dbf85de7433f86caee15d/pulldown-cmark/src/lib.rs#L429

The actual behavior of the library, as verified by the tests, is to space-separate custom attributes:

https://github.com/pulldown-cmark/pulldown-cmark/blob/4ba4738224a84de7447dbf85de7433f86caee15d/pulldown-cmark/tests/suite/heading_attrs.rs#L14


Martin1887:
Good catch, thanks!

PR #871, opened 2024-03-25
Math refactor

ollpu:
There are three changes here. (1) could be done without the other two to fix the bug.

(2) & (3) might have some small performance implications in some cases. However, they should make it easier to disallow empty display math items `$$$$`, which I'm planning to do in a follow-up PR.

1. Populate `math_delims` fully instead of stopping at an invalid closing delimiter. Inline code already works like this, though it can't encounter invalid delimiters.

   Fixes a bug when parsing e.g. this: `$x {$ $ } $x$`

2. Disallow `$$` matching a closing `$` and then marching delimiters in `make_math_span`. Instead, retry scanning at the second position (really just a `math_delims.find()`).
   
   Before:
   ```
   $$x$
   ^--|    # make_math_span called with this span
   ```
   After:
   ```
   $$x$
   ^--|    # no match, move forward
    ^-|    # make_math_span called with this span

3. Remove the `seen_first` optimization from `MathDelims`. It doesn't work with the retry strategy (2).

### DIFF
--- a/pulldown-cmark/specs/math.txt
+++ b/pulldown-cmark/specs/math.txt
@@ -336,14 +336,14 @@ This is: $\}$
 
 Math environment contains 2+2: $}$2+2$
 
-Math environment contains x: $x {$ $ } $x$
+Math environment contains y: $x {$ $ } $y$
 .
 <p>This is not valid math: $}{$</p>
 <p>Neither is this: { $}{$ }</p>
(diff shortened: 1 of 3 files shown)


ollpu:
Ok, apparently I was mistaken about `commonmark-hs` disallowing empty display math items ($$$$). Let's not change that then.

Martin1887:
Perfect, thanks!

PR #874, opened 2024-03-27
Fix doc of Options::ENABLE_MATH

duskmoon314:
There is no `Tag::Math` but two events. This commit fixes this typo.


Martin1887:
Thanks!

PR #875, opened 2024-03-31
Add symlink to license file so it is included in the release tarball that is published to crates.io

blinxen:
This is required by most distributions, since most of them package crates directly from crates.io.


Martin1887:
OK, thanks!

PR #876, opened 2024-04-04
fix: fix dead code warning from nightly compiler

rhysd:
I got the following warning from nightly rustc.

```
warning: field `span` is never read
    --> pulldown-cmark/src/parse.rs:1439:9
     |
1438 | pub struct FootnoteDef {
     |            ----------- field in this struct
1439 |     pub span: Range<usize>,
     |         ^^^^
     |
     = note: `FootnoteDef` has derived impls for the traits `Debug` and `Clone`, but these are intentionally ignored during dead code analysis
     = note: `#[warn(dead_code)]` on by default

warning: `pulldown-cmark` (lib) generated 1 warning
```

The field is not a public API and it is actually unused. This PR fixes the warning by removing the field.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -331,7 +331,7 @@ dependencies = [
 
 [[package]]
 name = "pulldown-cmark"
-version = "0.10.1"
+version = "0.10.2"
 dependencies = [
  "bincode",
  "bitflags",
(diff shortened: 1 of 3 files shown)


Martin1887:
Merged, thanks!

PR #880, opened 2024-04-22
(Re)introduce simd feature to pulldown-cmark-escape

ollpu:
I noticed the simd portion in `pulldown-cmark-escape/src/lib.rs` was dead code, as there was no `simd` feature defined on the new crate. This introduces that feature and automatically enables it through `pulldown-cmark` as well, if requested.

Guess no one noticed a difference in performance? :D


Martin1887:
My bad :sweat: , thanks for noticing it! I will do it also in the master branch a release a new 0.10 version.

PR #883, opened 2024-05-04
Add guide book and deploy script for it

### LINKED ISSUE #863: Pulldown-cmark mdBook
The crate documentation of docs.rs is good for learning the API, but some things like examples and specs are difficult to find for end users, and a mdBook would improve a lot the user experience for these types of things.

notriddle:
Fixes #863

You can see what it looks like at <https://notriddle.github.io/pulldown-cmark/>.

The final version would probably be at <https://pulldown-cmark.github.io>.

To make this work, you need to change the branch restrictions on deploying to github pages:

image

The action will then push a new version of the page whenever a release is made, or whenever a push is made to `test-deploy`.


.github/workflows/book.yml
```diff
@@ -0,0 +1,38 @@
...
+        mdbook build
+    - name: Deploy to GitHub
+      working-directory: guide
+      env:
+        GITHUB_DEPLOY_KEY: ${{ secrets.DEPLOY_KEY }}
+      run: |
+        touch book/.nojekyll
+        rustc deploy.rs
+        cd book
+        ../deploy
```
ehuss, line 38: Just a suggestion: I would recommend using GitHub's native deploy action instead of using deploy keys.

```suggestion
    - name: Upload Artifact
      uses: actions/upload-pages-artifact@v1.0.8
      with:
        path: ./book

  deploy:
    needs: build

    permissions:
      pages: write
      id-token: write

    environment:
      name: github-pages
      url: ${{ steps.deployment.outputs.page_url }}

    runs-on: ubuntu-latest
    steps:
      - id: deployment
        uses: actions/deploy-pages@v2.0.0
```

Martin1887:
I have to read it, but on a fast look it seems great, thanks!

ehuss:
You may also want to remove `guide/deploy.rs`.

chriskrycho:
First of all, this is a fantastic addition! It would be awesome if shipping this kind of thing just became a normal part of libraries in the Rust community. 💙 

Second, content structure suggestion: move the CommonMark spec. My recommendation would be to either put it as an appendix at the end, or remove the content entirely in favor of a link. If someone is just coming to the library for the first time and wants a quick guide, the spec is… quite the second page to get to. 😂 More than that, though, I think for most people, the spec is actually irrelevant, and just linking to the spec (maybe links to both the copy in the repo and the upstream?) would be more than sufficient. Thoughts?

Martin1887:
I have just read it and I find it really good! The cheatsheet chapter is very useful for new users.

However, I miss an introduction chapter (maybe just importing the README and/or docs.rs main page despite repeating them?) and a chapter to include the examples of the `examples` directory of the repository.

Thanks!

Martin1887:
Sorry for the delay, I'm been busy this week.

I see perfect the introduction and the introduction of the Math spec, but:

- In the cheat sheet, I think Math mode should not be under GFM features because we use another spec.
- The chapter 2 links seem to redirect to another subdomain, which feels weird and remove the mdBook theme (burning your eyes if you are using a dark theme :stuck_out_tongue_winking_eye:), could not the examples be transformed to mdBook chapters? For the API docs I think we can create a mini-chapter explaining that the API documentation is in docs.rs and indicating the link, so it is clear it is an external source.

Thanks!

Martin1887:
We could do it, but examples are not currently in docs.rs and it seems a bit weird placing them there, no?

I think the best place for examples is the book, but other alternatives could be also good.

Martin1887:
Really interesting, I didn't know that. But this feature is not stable, could it be used in the official docs.rs documentation just publishing the crate as usual?

Martin1887:
I think that it is better to use docs.rs for a more standard use of docs, we can use scraped examples when that feature become stable.

I think the last commits make this pull request perfect, thanks!

I will wait a bit to merge it in case someone else has any comments.

rhysd:
@Martin1887 @notriddle

The deploy job seems to need fix: https://github.com/pulldown-cmark/pulldown-cmark/actions/workflows/pages/pages-build-deployment

ehuss:
The workflow seems to be using very old versions of `actions/upload-pages-artifact` and `actions/deploy-pages`. I would suggest updating those to the latest.

rhysd:
I confirmed it was fixed after updating the actions. I created a PR #920.

ollpu:
What's the status of this? https://pulldown-cmark.github.io/pulldown-cmark is currently giving 404.

What would be involved in changing the URL to `https://pulldown-cmark.github.io/`? I guess there needs to be a repo `pulldown-cmark/pulldown-cmark.github.io` in the org and then some (??) adjustments to the Actions workflow to get it deployed there.

ollpu:
I tried pushing to `test-deploy`. It says [this](https://github.com/pulldown-cmark/pulldown-cmark/actions/runs/10375617959/job/28725701397):

> Error: Failed to create deployment (status: 404) with build version [..]. Request ID [..] Ensure GitHub Pages has been enabled: [..]

Martin1887:
It is a bit weird, it was already configured but maybe it got reset after changing the version of the deployment module.

It is working again now. Thanks for report it.

ollpu:
The release deployment is giving this error:
```
Tag "v0.12.0" is not allowed to deploy to github-pages due to environment protection rules.
```
which is apparently related to https://github.blog/changelog/2023-10-06-actions-secure-deployment-rollouts-to-protected-environments-based-on-select-tag-patterns/

Martin1887:
The thing is that a new tag should not trigger a book deploy I think, but I don't see why it is triggered.

rhysd:
https://pulldown-cmark.github.io/pulldown-cmark/ looks working fine to me. What is the problem being discussed here?

> The thing is that a new tag should not trigger a book deploy I think, but I don't see why it is triggered.

Yes, it is triggered on `release` event, not tag or branch creation.

https://github.com/pulldown-cmark/pulldown-cmark/blob/ecf248d8ed6559ea322ed624fb5240f5fbc187ee/.github/workflows/book.yml#L5C5-L5C12

rhysd:
Since I don't have the permission, I can't ensure but I guess the setting on this repository is not correct. The steps are:

1. Go to 'Settings' tab
2. Select 'Environments' item in the menu
3. Find 'Deployment branches and tags' section
4. Add 'Add deployment branch or tag rule' link
5. Select 'Ref type: Tag' in the pulldown menu
6. Add a pattern matching to any release tags

image

Martin1887:
I had only set `test-deploy` and `gh-pages` branches, a pattern for all `v*` tags is already set.

rhysd:
Hmm, `v*` should matches to `v0.12.0` 🤔

rhysd:
I'm not sure, but this discussion may be related to this issue: https://github.com/orgs/community/discussions/39054

We deploy the page on `release` event. And the event is raised on the default branch (if my understanding is correct). How about adding `master` to the allowed branches?

Martin1887:
It works (I have just re-executed the last job), the problem was the `v*` tags were not allowed when the releases were created.

PR #885, opened 2024-05-07
fix: remove unnecessary end_newline set

tomcur:
`self.write` sets `self.end_newline = false`:
https://github.com/tomcur/pulldown-cmark/blob/de575f0f328eaf574fd747c8510ce98012a05ddd/pulldown-cmark/src/html.rs#L86-L88


Martin1887:
Thanks!

PR #901, opened 2024-06-04
feat: set `DefaultBrokenLinkCallback` as the default broken link callback of `OffsetIter`

rhysd:
`Parser`'s `F` type parameter has the default type `DefaultBrokenLinkCallback` so that users can use the type without caring about the type parameter. However `OffsetIter` doesn't have the default type so users need to specify `DefaultBrokenLinkCallback` manually knowing the API.

This PR sets `DefaultBrokenLinkCallback` to the default callback type of `OffsetIter` as well as `Parser`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -331,7 +331,7 @@ dependencies = [
 
 [[package]]
 name = "pulldown-cmark"
-version = "0.10.3"
+version = "0.11.0"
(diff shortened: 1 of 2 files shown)


ollpu (approved):
Makes sense.

Martin1887:
Yup, good catch!

PR #904, opened 2024-06-04
chore(doc): use `cargo add` to instruct how to install this crate as dependency

rhysd:
`cargo add` was introduced recently to Cargo and it works well to install the latest version of this crate as dependency.

Since the version is included in the TOML snippets in README.md, they need to be updated when the major version of this crate is bumped. By using `cargo add`, we no longer need to take care about it and users can easily install this crate as dependency by copy&paste the command line.


Martin1887:
Right, good point!

PR #905, opened 2024-06-04
feat: add `-G` CLI option to enable GFM support

rhysd:
This PR adds a new CLI option to enable the `Options::ENABLE_GFM` parser option like other parser options. It is useful to understand how the parser option works in terminal.


Martin1887:
Yeah, good catch!

PR #908, opened 2024-06-15
Fix parsing blocks inside alert body

### LINKED ISSUE #907: Blocks inside alert body is parsed as blocks outside the body
## Repro

Save this content as `foo.md`

```markdown
> [!Note]
> - Foo
```

Then run the following command in this repository

```sh
cargo run -- -G < foo.md
```

## Expected behavior

The list is parsed as the content inside the alert body:

```html
<blockquote class="markdown-alert-note">
<ul>
<li>Foo</li>
</ul>
</blockquote>
```

## Actual behavior

The list is parsed as the content outside the alert body:

```html
<blockquote class="markdown-alert-note">
</blockquote>
<ul>
<li>Foo</li>
</ul>
```

## Environment

- Rust: 1.79
- pulldown-cmark: 36d576ece45928fd2cc5532394110de5c86f2528

## Additional info

GitHub renders `- Foo` as a content inside the alert block as follows:

> [!Note]
> - Foo

I'll make a pull request to fix this issue.

rhysd:
Fixes #907

The root cause is that the parser eats `> ` at the next line after eating the `> [!Note]` line. So the content is parsed as if it is outside the body:

```markdown
> [!Note]
> - Foo
```

is parsed as

```markdown
> [!Note]
- Foo
```

The `> ` at the next line should not be eaten to know the block quote continues.

### DIFF
--- a/pulldown-cmark/specs/blockquotes_tags.txt
+++ b/pulldown-cmark/specs/blockquotes_tags.txt
@@ -191,3 +191,14 @@ Like lazy continuations, the blockquote marker is still needed to make nested bl
 <ul><li>sink ships</li></ul></blockquote>
 </li></ul>
 ````````````````````````````````
+
+When an unknown tag is found, parser backtracks and the block is parsed as a normal blockquote:
+```````````````````````````````` example
+> [!Hello]
+> This should be a normal block quote.
+.
+<blockquote>
+<p>[!Hello]
+This should be a normal block quote.</p>
+</blockquote>
+````````````````````````````````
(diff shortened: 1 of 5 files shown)


rhysd:
I also added one more test case to clarify how unknown blockquote tag is handled. This test case tests the parser's backtracking here:

https://github.com/pulldown-cmark/pulldown-cmark/blob/36d576ece45928fd2cc5532394110de5c86f2528/pulldown-cmark/src/scanners.rs#L260

Martin1887:
Merged, thanks!

PR #915, opened 2024-06-21
Implement commonmark-hs compatible definition lists

### LINKED ISSUE #67: Support for Description Lists

notriddle:
Fixes #67


Martin1887:
Awesome! But I think it would be better to merge it in the new `branch_0.12` branch I have just created, because the last changes probably deserve a new 0.11 minor release.

Also, I wonder if the number of calls to `self.finish_list` in `firstpass.rs` could be reduced by dividing it into a function for normal lists and another one for description lists or something else. It seems the only messier part of the code respect to the previous version.

Thanks!

Martin1887:
A lot of time since the last update and I don't get a cleaner manner of doing it, so let's merge it. Thanks!

alyjak:
It looks like the definition lists do not work if they are embedded in another block type. Both of these examples are failing for me on v0.12.2 with the following panic: 

```panicked at /home/alyjak/.cargo/registry/src/index.crates.io-6f17d22bba15001f/pulldown-cmark-0.12.2/src/firstpass.rs:1540:38:
called `Option::unwrap()` on a `None` value`.
```

``` markdown
>
> A definition
> : embedded in a quote
>
```

Or this

``` markdown
A definition
:   an embedded term
   :   embedd definition
```

Is this expected behavior? Thanks for the hard work. I really appreciate this library.

Martin1887:
Please open a new issue with this problem, thanks!

Martin1887:
Please open a new issue with this problem, thanks!

Martin1887:
Please open a new issue with this problem, thanks!

PR #916, opened 2024-06-21
Reduce MSRV to 1.71.1 by separating benchmarks into a new crate with CI enhancements

rhysd:
Closes #894

This PR includes the following changes:

- Reduce MSRV of pulldown-cmark to 1.71.1 with updating the version in README.md
- Describe 1.74 or later is necessary for development as we discussed in #894
- Separate `pulldown-cmark/benches` as [a new crate `bench`](https://github.com/rhysd/pulldown-cmark/tree/msrv/bench) so that we can avoid pulldown-cmark depending on criterion which requires 1.74. This change allows rustc 1.71.1 to compile the tests for pulldown-cmark on `cargo test`.
- Currently only MSRV is tested on CI. So I added stable and nightly toolchains to test targets. This is because:
  - Stable toolchain is the most popular toolchain used to compile this crate by users
  - Nightly toolchain is one of our use cases because this crate is depended by Rust toolchain (rustdoc)

I confirmed CI passed on my fork: https://github.com/rhysd/pulldown-cmark/actions/runs/9608452645


Martin1887:
Wonderful approach but, as it creates a new crate, I think it is better to merge it in the new `branch_0.12` branch I have just created.

Thanks!

rhysd:
Hi @Martin1887, thanks for sharing your thoughts.

In my opinion, this change can be a part of v0.11.x because this is not a breaking change for users.

- The new crate is for internal use. It is never released on crates.io and users will never use it. All APIs of pulldown-cmark crate are preserved as-is (since we just moved benchmarks and didn't touch any library code).
- The MSRV downgrade is not a breaking change in contrast to MSRV upgrades. pulldown-cmark v0.11 users are already using this crate with Rust 1.74 or later.

Anyway, I'm okay with merging this branch into the `branch_0.12` if you still think the next version should be major version bump. Would you let me hear your thoughts?

Martin1887:
OK, I was also not sure of the major change but I did prefer be on the safe side :smile: . But I think you are right and this change would be valuable for a new minor 0.11 release.

GuillaumeGomez:
Thanks for this PR! We are adding `pulldown-cmark` into [rinja](https://github.com/rinja-rs/rinja) and we are very interested into this MSRV reduction. Do you know when the new `0.11` version will be released? Thanks in advance!

Martin1887:
I hope that I can release it at the end of the week.

GuillaumeGomez:
Awesome, thanks!

Martin1887:
It has just been published.

GuillaumeGomez:
Awesome, thanks!

PR #924, opened 2024-08-10
Fix lone task list item bug

### LINKED ISSUE #882: Tasklists with nameless checkboxes move checkbox to next paragraph
Suppose you have the following markdown code, which contains a nameless checkbox:

```md
* [ ] 
---
bar
```

So, we run:

```sh
printf '* [ ] \n---\nbar' | cargo run -- --enable-tasklists
```

This results in:

```html
<ul>
<li></li>
</ul>
<hr />
<p><input disabled="" type="checkbox"/>
bar</p>
```

Which looks like a stray checkbox after the horizontal rule:

markdown_tasklist_stray_checkbox

The checkbox was erroneously added to the next paragraph element.

Additionally, if there is no paragraph following, the checkboxes disappear completely:

```sh
printf '* [ ] \n* [ ] \n* [ ] \n' | cargo run -- --enable-tasklists
```

```html
<ul>
<li></li>
<li></li>
<li></li>
</ul>
```

It seems to me like this behavior was introduced after #557 was fixed.

### LINKED ISSUE #921: [fuzz result] thread 'main' panicked: slice index starts at 8 but ends at 5
```rust
use pulldown_cmark::Options;

fn main() {
    let string = "- [x]\t\t\n\\\n-";

    let mut opts = pulldown_cmark::Options::empty();

    opts.insert(Options::ENABLE_TASKLISTS);
    opts.insert(Options::ENABLE_HEADING_ATTRIBUTES);

    for _ in pulldown_cmark::Parser::new_ext(string, opts) {}
}
```
```
thread 'main' panicked at pulldown-cmark-0.11.0/src/firstpass.rs:1965:49:
slice index starts at 8 but ends at 5
```

https://github.com/pulldown-cmark/pulldown-cmark/blob/d7632acbe066b81a7f4b23c72fdb18c85f9cb1ea/pulldown-cmark/src/firstpass.rs#L1964-L1965

Btw, this is a label:fuzzer_finding, but cannot add this myself :)

notriddle:
Fixes #882
Fixes #921

### DIFF
--- a/pulldown-cmark/specs/regression.txt
+++ b/pulldown-cmark/specs/regression.txt
@@ -2608,3 +2608,30 @@ We instead set the limit at 32, which is the same limit used by cmark-gfm and ma
 <p><a href="https://rust.org/something%3A((((((((((((((((((((((((((((((()))))))))))))))))))))))))))))))">30</a>
 [40](https://rust.org/something%3A((((((((((((((((((((((((((((((((((((((((())))))))))))))))))))))))))))))))))))))))))</p>
 ````````````````````````````````
+
+ISSUE #921
+
+```````````````````````````````` example_metadata_blocks
+- [x]		
+\
+-
+.
+<ul>
+<li><input disabled="" type="checkbox" checked=""/></li>
+</ul>
+<h2>\</h2>
+````````````````````````````````
+
+ISSUE #882
+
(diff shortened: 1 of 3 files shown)


Martin1887:
Merged, thanks!

PR #931, opened 2024-08-13
Refactor TextMergeStream

### LINKED ISSUE #922: `TextMergeWithOffsets` adds spaces in indented code blocks
`TextMergeWithOffsets` can merge text nodes that are not actually adjacent in the source document.

The default Parser parses an indented code block as a series of text events, one per line.  `TextMergeWithOffsets` merges all of these into a single event; but the text for that event is borrowed and contains the extra indentation characters in between those nodes. (These are usually spaces, but I think they could even contain `>` characters if the code block is in a block quote.)

I think this must be a bug. It's surprising and hard to work around, and I don't think `TextMergeStream` behaves this way.

 The code with the bug is [here](https://docs.rs/pulldown-cmark/latest/src/pulldown_cmark/utils.rs.html#141-143).

ollpu:
Fixes #922 

- Rewrite `TextMergeWithOffset`. The old implementation borrowed text from the source based on the offsets, instead of combining the content properly.
- `TextMergeWithOffset` now has the same iterator adapter API as `TextMergeStream`, i.e. `::new(iter)`, instead of replicating the parser constructors.
- Implement `TextMergeStream` in terms of `TextMergeWithOffset`. The `DummyOffsets` adapter is a little verbose, but oh well (can't use `std::iter::Map` because the type must be named).
- A couple simple tests

### DIFF
--- a/pulldown-cmark/src/utils.rs
+++ b/pulldown-cmark/src/utils.rs
@@ -9,16 +9,13 @@
 //! Its author proposed the solution in
 //! <https://github.com/raphlinus/pulldown-cmark/issues/708>.
 
-use crate::{
-    BrokenLinkCallback, CowStr, DefaultBrokenLinkCallback, Event, OffsetIter, Options, Parser,
-};
-use std::{iter::Peekable, ops::Range};
+use crate::{CowStr, Event};
+use std::ops::Range;
 
 /// Merge consecutive `Event::Text` events into only one.
 #[derive(Debug)]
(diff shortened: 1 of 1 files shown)


Martin1887:
Great! But the interface has been changed, right? Then it should be merged in the 0.12 branch.

Thanks!

ollpu:
Ok, I hadn't noticed it.

Martin1887:
Thanks!

PR #932, opened 2024-08-13
Add regression test for #655

### LINKED ISSUE #655: Parsing inline code span with a single newline loses the newline
According to the [Commonmark playground](https://spec.commonmark.org/dingus/?text=%60%0A%60%0A) and the [Commonmark spec](https://spec.commonmark.org/0.30/#example-335), the string

```rust
"`\n`"
```

should be parsed into

```rust
[
  Start(Paragraph)
  Code(" ")
  End(Paragraph)
]
```

since newline is normalized as a space.

However, when parsing the string with pulldown-cmark 0.9.3, I get

```rust
[
  Start(Paragraph)
  Code("")
  End(Paragraph)
]
```

instead. I used this little test program:

```rust
use pulldown_cmark::{Event, Parser};

fn main() {
    let text = "`\n`";
    eprintln!("{text:?} -> [");
    for event in Parser::new(&text) {
        match event {
            Event::Code(code) => eprintln!("  Code({:?})", &*code),
            _ => eprintln!("  {event:?}"),
        }
    }
    eprintln!("]");
}
```

Note that `` "`foo\n`" `` and `` "`\nfoo`" `` are parsed correctly with a trailing and leading space, respectively.

I found this while trying round-trip strings through pulldown-cmark-to-cmark (https://github.com/Byron/pulldown-cmark-to-cmark/pull/55).

ollpu:
Closes #655


Martin1887:
Thanks!

Issue #645, opened 2023-05-14
External HTML crate meta-issue


Martin1887:
The purpose of this issue is tracking all issues and pull requests proposing features for an external full-featured crate for HTML rendering.

#629 and #631 pull requests propose solving #584 in different manners, and they can inspire an external crate.

#524 asks for opening links in a new tab.

#408 asks for bidirectional text.

#637 asks for lazy loading images.

Please comment here if a new HTML renderer crate is created, since it could be recommended in docs.

Martin1887:
#525 implements the opening of links in a new tab.

Martin1887:
#933 proposed an option to replace paragraphs by spans.

systemsoverload:
Did you have anything in mind for how you'd like an external crate to expose itself for this lib? I have a janky custom iterator for overriding HTML (emitting custom HTML events rather than the built-in events) and its just gross enough that I am considering biting the bullet and spinning up such a crate.

Martin1887:
We don't have an opinionated design, but the external crate would not be used by this one. The idea is to have a crate with more options for people who like the performance of this crate but would like more customization.

My idea is a bin crate configured via a TOML file and a lib crate that receives a configuration object deserialized from the TOML with serde. The crate core would use a loop over events as the `html.rs` of this crate but taking into account the options to render content in the desired manner and adding more attributes where needed.

So, having a `HtmlWriter / push_html` is not required but a perfectly valid design option.

Thanks for your interest.

systemsoverload:
Ok I think I've got a reasonable start here:
[Repo](https://github.com/systemsoverload/pulldown-html-ext)
[Lib crate](https://crates.io/crates/pulldown-html-ext)
[CLI Crate](https://crates.io/crates/pulldown-html-ext-cli)

This is pretty rough, notably the configurable options were mostly arbitrarily selected and not all of the `pulldown-cmark` test that I copied over pass just yet. But I wanted to get some input on whether or not this is approximately what you were thinking before I went too much further, but im working on using this lib in my own static site generator library and it seems to be a reasonable alternative so far 🤞

Martin1887:
I have not reviewed it in depth but it seems a good first approach!

systemsoverload:
After iterating on this for a while longer, my opinion is starting to sway towards implementing some of what I have in the external crate directly in `pulldown-cmark` instead.  

[...]

PR #933, opened 2024-08-14
Add support for generating html for markdown which should be shown inline

jhoobergs:
I don't know if this is something you want in this crate, but I needed it so I implemented it :) 

Basically this PR adds an `inline: bool` argument for html generation which let's the generator add `span`-tags instead of `p`-tags around a markdown paragraph.

The current implementation is somewhat of a 'breaking' change as the functions to generate html now have an extra argument. An alternative could be to duplicate the methods for inline generation instead of adding an argument.

### DIFF
--- a/pulldown-cmark/examples/broken-link-callbacks.rs
+++ b/pulldown-cmark/examples/broken-link-callbacks.rs
@@ -23,7 +23,7 @@ fn main() {
 
     // Write to String buffer.
     let mut html_output: String = String::with_capacity(input.len() * 3 / 2);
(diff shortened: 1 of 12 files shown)


Martin1887:
Sorry, but the scope of the HTML renderer of this project is very limited and almost only for demonstration purposes.

However, we encourage people to create a new crate that uses pulldown-cmark to generate a HTML renderer with extended configuration and capabilities, and this sort of features matches perfectly that use case (see #645).

Thanks!

PR #935, opened 2024-08-15
Renovate the superlinear time fuzzer

ollpu:
Some fixes and adjustments to the superlinear time fuzzer to make it work again. Found this issue with it: https://github.com/pulldown-cmark/pulldown-cmark/issues/934

- Update crates, fix source code path, add `\n` as a token
- Rewrite the literal parser using `syn::visit`
  - This should be a little simpler to maintain, as the bulk of the AST walk is handled by syn itself, while we only redefine custom behavior for some nodes

By the way, I think this component (or at least the directory) should be renamed. We currently have both `fuzz` and `fuzzer`, which is a little confusing. Bikeshedding:
- `superlinear-fuzzer`
- `perf-fuzzer`

Thoughts on this?

### DIFF
--- a/fuzzer/Cargo.lock
+++ b/fuzzer/Cargo.lock
@@ -4,78 +4,70 @@ version = 3
 
 [[package]]
 name = "anstream"
-version = "0.6.11"
+version = "0.6.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6e2e1ebcb11de5c03c67de28a7df593d32191b44939c482e97702baaaa6ab6a5"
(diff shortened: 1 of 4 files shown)


ollpu:
@Martin1887 Any ideas about renaming the component?

Martin1887:
`perf-fuzzer` sounds good I think, maybe `antiddos-fuzzer` or something like that?

ollpu:
Oh how about just `dos-fuzzer`

Martin1887:
It searches for dos behaviours, so I think it is good :smile: .

PR #943, opened 2024-08-20
impl From<CowStr> for String

oconnor663:
This matches the corresponding Cow impl that was stabilized in Rust 1.14: https://doc.rust-lang.org/std/string/struct.String.html#impl-From%3CCow%3C'a,+str%3E%3E-for-String

The main benefit of this new impl is that you can now pass a `CowStr` to a function that takes `impl Into<String>`.


Martin1887:
Thanks! It seems that this has not any drawback.

PR #945, opened 2024-08-30
Make dos-fuzzer part of the workspace

kdarkhan:
This is possible because of the change introduced in cargo-fuzz
in https://github.com/rust-fuzz/cargo-fuzz/pull/357.

Using the same workspace enables reusing same deps and sharing cache between members. It also has better editor support.


kdarkhan:
I attempted to do the same for `fuzz` dir but faced compilation errors. I believe they are caused by outdated patched dependency on moz-js. If the project maintainers approve of the approach, I can replace [the existing JS based parser](https://github.com/pulldown-cmark/pulldown-cmark/blob/18161a0e968be5b6c64001cf5481f3cb46c69a8c/fuzz/src/lib.rs#L33) with [reference implementation in C](https://github.com/commonmark/cmark). The C library could be referenced using bindgen. Let me know what you think.

ollpu (approved):
This looks alright if all the scripts still work. Do note that `dos-fuzzer` doesn't use cargo-fuzz, it's a custom thing entirely.

ollpu:
```
warning: profiles for the non root package will be ignored, specify profiles at the workspace root:
```

The root already has `lto = true`. You can probably just remove it from `dos-fuzzer/Cargo.toml`.

kdarkhan:
> ```
> warning: profiles for the non root package will be ignored, specify profiles at the workspace root:
> ```
> 
> The root already has `lto = true`. You can probably just remove it from `dos-fuzzer/Cargo.toml`.

Removed the config and also adjusted the relative path in `dos-fuzzer/run` script.

kdarkhan:
> OK, so I think that this can be merged as it is and another pull request could be opened for the other fuzzer, right?

Yes, I can work on the other part separately.

Just to be sure, can somebody confirm if I can replace the commonmark JS based implementation with the one implemented in C? It will require anybody who wants to run fuzzers installing the compiled library on machines.

ollpu:
> Just to be sure, can somebody confirm if I can replace the commonmark JS based implementation with the one implemented in C? It will require anybody who wants to run fuzzers installing the compiled library on machines.

I would rather try to find a solution for moz-js or change how the JS is run, but I am not very familiar with the setup.

PR #948, opened 2024-09-05
Make `fuzz` dir part of the workspace

kdarkhan:
This introduces 2 changes:

**Make `fuzz` dir part of the workspace**

Previously this was not possible in `cargo-fuzz` but became possible after https://github.com/rust-fuzz/cargo-fuzz/pull/357. It has the benefit of sharing build cache with the other crates and better editor support.

**Switch to upstream `mozjs`**

Not sure why the fork was used by @notriddle in https://github.com/pulldown-cmark/pulldown-cmark/pull/842. I did run each fuzzer with upstream dep for a minute without triggering any issues. Please let me know if this change should not be done.


Martin1887:
So everything is fine to merge this?

PR #953, opened 2024-09-08
Add instructions on fixing fuzz build

kdarkhan:
When I was testing https://github.com/pulldown-cmark/pulldown-cmark/pull/948,
I built it directly using `cargo build` from within `fuzz` crate which worked as expected.
However, when using fuzzers with `cargo fuzz`, it overrides a lot of compilation parameters and that fails the built.

```bash
: rust-lld: error: undefined symbol: __sancov_gen_.1094
          >>> referenced by parse.3be71763e9de75d0-cgu.0
          >>>               /home/user/projects/pulldown-cmark/target/x86_64-unknown-linux-gnu/release/deps/parse-4bac226fcf249aac.parse.3be71763e9de75d0-cgu.0.rcgu.o:(asan.module_dtor.1168)
```

I spent some time investigating the issue and the easiest fix seems to be overriding `lto` config in release profile which could be done using environment variables:

```bash
CARGO_PROFILE_RELEASE_LTO=thin cargo fuzz build
```

This PR adds documentation on the issue and how to work around it.

### DIFF
--- a/fuzz/README.md
+++ b/fuzz/README.md
@@ -1,0 +1,25 @@
+# Fuzz targets
+
+This crate specifies fuzzing targets which are
+instrumented with [cargo-fuzz](https://github.com/rust-fuzz/cargo-fuzz).
+
+## Fixing fuzz build issue
+
+At the moment, building fuzz targets with default settings
+(`cargo fuzz build`) throws many errors like this:
+
+```
+: rust-lld: error: undefined symbol: __sancov_gen_.1094
+          >>> referenced by parse.3be71763e9de75d0-cgu.0
+          >>>               /home/user/projects/pulldown-cmark/target/x86_64-unknown-linux-gnu/release/deps/parse-4bac226fcf249aac.parse.3be71763e9de75d0-cgu.0.rcgu.o:(asan.module_dtor.1168)
+```
+
+The issue seems to be triggered during linking of the binaries
+with default `profile.release.lto=true` set at the workspace `Cargo.toml`
+file.
+
+To fix the build, you can override `lto` config using env variable:
+
+```bash
+$ CARGO_PROFILE_RELEASE_LTO=thin cargo fuzz run parse -- -only_ascii=1 -max_total_time=60
+```


Martin1887:
Thanks!

ollpu:
I wonder if this is cause to revert #948 - `cargo fuzz` just doesn't work if you've compiled something else before?

Are there any benefits to a shared workspace if everything has to be rebuilt under different flags anyway?

Martin1887:
Good point... especially when fuzzers are a different monster.

kdarkhan:
> I wonder if this is cause to revert https://github.com/pulldown-cmark/pulldown-cmark/pull/948 - cargo fuzz just doesn't work if you've compiled something else before?

`rustc` is smart enough to recompile everything if you override `lto` config. Failure to build fuzz does not depend previous builds done, it forces a new build. Reverting the whole PR is not worth it I think. The PR introduced 2 changes:

1. Switches `mozjs` to upstream instead of the fork maintained by @notriddle
2. Makes `fuzz` part of the workspace.

I just verified that before my PR `fuzz` crate was not buildable with default llvm deps in Debian but was buildable in Archlinux. Now, with the PR merged it builds correctly in both Debian/Archlinux. We can revert the workspace sharing changes but `mozjs` switch from `1` should still stay.

> Are there other benefits to a shared workspace considering everything has to be rebuilt under different flags anyway?

For regular devs who don't use `fuzz`, nothing should change. Because `fuzz` is not listed in [default-members](https://github.com/pulldown-cmark/pulldown-cmark/blob/64a38ecf6f1bfea69e734d56abb61fd7a0971b12/Cargo.toml#L2), all commands not explicitly stating the `--package` will ignore `fuzz`.
For people who want to work on  fuzzers, sharing the workspace has better editor support. I did face an issue with `rust-analyzer` with default settings, it would complain that `fuzz` files are not part of the workspace and diagnostics stops working. The fix was opening the editor from `fuzz` dir which adds some friction. Another benefit is sharing the build cache. I am planning to introduce build manifest caching to github action + building `fuzz` dir during PR build. Building `fuzz` from scratch is slow but if previous manifest are cached it should be much quicker. This will allow catching issues like https://github.com/pulldown-cmark/pulldown-cmark/pull/947 before merge. It is technically possible to do the same with separate `fuzz` workspace but will require setting up different caches for root and fuzz workspaces.

ollpu:
> mozjs switch from 1 should still stay.

Yes! I didn't think of that part.

Editor support is a good point. I had similar issues while working on dos-fuzzer.

I suppose then the current status is fine. Hopefully the issue can eventually be fixed in cargo-fuzz (or wherever in the stack).

PR #967, opened 2024-10-06
Implement into_static for CowStr and Event in pulldown-cmark

Atreyagaurav:
Fixes: https://github.com/pulldown-cmark/pulldown-cmark/issues/625

I found that we cannot implement ToOwned, but I found the following code:
```rust
impl<'a> BrokenLink<'a> {
    /// Moves the link into version with a static lifetime.
    ///
    /// The `reference` member is cloned to a Boxed or Inline version.
    pub fn into_static(self) -> BrokenLink<'static> {
        BrokenLink {
            span: self.span.clone(),
            link_type: self.link_type,
            reference: self.reference.into_string().into(),
        }
    }
}
```
So, I added similar implementations for other types with lifetimes so we can use them more freely.

### DIFF
--- a/pulldown-cmark/src/strings.rs
+++ b/pulldown-cmark/src/strings.rs
@@ -262,7 +262,14 @@ impl<'a> CowStr<'a> {
     }
 
     pub fn into_static(self) -> CowStr<'static> {
-        self.into_string().into()
+        match self {
+            CowStr::Boxed(b) => CowStr::Boxed(b),
+            CowStr::Borrowed(b) => match InlineStr::try_from(b) {
+                Ok(inline) => CowStr::Inlined(inline),
+                Err(_) => CowStr::Boxed(b.into()),
+            },
+            CowStr::Inlined(s) => CowStr::Inlined(s),
+        }
     }
 }
 
--- a/pulldown-cmark/src/strings.rs
+++ b/pulldown-cmark/src/strings.rs
@@ -260,6 +260,10 @@ impl<'a> CowStr<'a> {
             CowStr::Inlined(s) => s.deref().to_owned(),
         }
     }
+
+    pub fn into_static(self) -> CowStr<'static> {
+        self.into_string().into()
+    }
 }
 
 impl<'a> fmt::Display for CowStr<'a> {
--- a/pulldown-cmark/src/firstpass.rs
+++ b/pulldown-cmark/src/firstpass.rs
@@ -142,7 +142,8 @@ impl<'a, 'b> FirstPass<'a, 'b> {
                 }
             }
             let container_start = start_ix + line_start.bytes_scanned();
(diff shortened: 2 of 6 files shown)


Atreyagaurav:
Please ignore/let me know if my auto-formatter messed things up in the repo.

rhysd:
I'm not a maintainer but I usually try not to apply code formatter in order to avoid unnecessary code changes.

ollpu:
i think it's long overdue that we enforce rustfmt in CI, but that's something to do separately (and I don't know of other's opinions).

pulldown-cmark/src/strings.rs
```diff
@@ -260,6 +260,10 @@ impl<'a> CowStr<'a> {
             CowStr::Inlined(s) => s.deref().to_owned(),
         }
     }
+
+    pub fn into_static(self) -> CowStr<'static> {
+        self.into_string().into()
```
ollpu, line 265: I think this should take advantage of the fact that only the `Borrowed` variant needs to reallocate, others can be no-ops.

Atreyagaurav: Thank you.

The variants looked a bit complicated so I went an easy route previously. Fixed it now.

Atreyagaurav:
Yeah, but I didn't notice the auto-formatter until I commited. This pull request is just adding things, nothing on changing things, so it should not be hard to read, it's just some files. Please forgive me for that :)

pulldown-cmark/src/strings.rs
ollpu: One more improvement would be to utilize the Inline variant if possible. The Clone implementation has an example of doing this.

```rust
match InlineStr::try_from(s) {
    Ok(inline) => CowStr::Inlined(inline),
    Err(..) => CowStr::Boxed(s.into()),
},
```

Atreyagaurav: I don't need to do that for the `Boxed` variant as well, right? Since, we're consuming the `self` and passing current `Box<str>` to output.

ollpu:
> Yeah, but I didn't notice the auto-formatter until I commited.

No worries! If you want to though, you could restore (at least the files that have no other changes) & squash.

```
git reset --soft master
git restore --staged -- pulldown-cmark/src/firstpass.rs pulldown-cmark/src/linklabel.rs pulldown-cmark/src/scanners.rs
git commit
```

pulldown-cmark/src/strings.rs
ollpu: Correct. Only the Borrowed variant needs to be changed in this case. Cloning is a bit different.

Atreyagaurav:
I could, but can we leave them be? I think it's good to have them formatted anyway.

ollpu:
Yeah it's fine, we can leave it be.

ollpu (approved):

ollpu:
Thanks for the PR! This is simple yet quite useful.

PR #969, opened 2024-10-08
Respect line starts when trimming header endings

notriddle:
This prevents it from trying to trim whitespace that's part of a block quote or list marker.

Fixes https://github.com/pulldown-cmark/pulldown-cmark/issues/963

This is an alternative fix to https://github.com/pulldown-cmark/pulldown-cmark/pull/968 that handles a few corner cases better. As @rhysd figured out the root cause, the problem is that the code stripped a space that was actually part of the block quote marker, attempting to trim the text node beyond its starting point.

### DIFF
--- a/pulldown-cmark/src/firstpass.rs
+++ b/pulldown-cmark/src/firstpass.rs
@@ -724,8 +724,25 @@ impl<'a, 'b> FirstPass<'a, 'b> {
             let new_end = if has_trailing_content {
                 content_end
             } else {
+                let mut last_line_start = header_start;
+                loop {
+                    let next_line_start = last_line_start + scan_nextline(&bytes[last_line_start..content_end]);
+                    if next_line_start >= content_end {
+                        break;
+                    }
+                    let mut line_start = LineStart::new(&bytes[next_line_start..content_end]);
+                    if scan_containers(
+                        &self.tree,
+                        &mut line_start,
+                        self.options.has_gfm_footnotes(),
+                    ) != self.tree.spine_len()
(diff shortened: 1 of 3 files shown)


pulldown-cmark/src/firstpass.rs
```diff
@@ -724,8 +724,25 @@ impl<'a, 'b> FirstPass<'a, 'b> {
...
+                    if scan_containers(
+                        &self.tree,
+                        &mut line_start,
+                        self.options.has_gfm_footnotes(),
+                    ) != self.tree.spine_len()
+                    {
+                        break;
+                    }
+                    last_line_start = next_line_start + line_start.bytes_scanned();
+                }
```
rhysd, line 743:
```suggestion
                if attrs.is_some() {
                    loop {
                        let next_line_start = last_line_start + scan_nextline(&bytes[last_line_start..content_end]);
                        if next_line_start >= content_end {
                            break;
                        }
                        let mut line_start = LineStart::new(&bytes[next_line_start..content_end]);
                        if scan_containers(
                            &self.tree,
                            &mut line_start,
                            self.options.has_gfm_footnotes(),
                        ) != self.tree.spine_len()
                        {
                            break;
                        }
                        last_line_start = next_line_start + line_start.bytes_scanned();
                    }
                }
```

Thanks for the better fix. If I'm understanding the issue correctly, this corner case only happens when heading attributes exist (because an empty line is never valid for setext header). The scanning part is only necessary when `attributes` is not empty.

PR #971, opened 2024-10-09
Enforce cargo fmt

ollpu:
Let's enforce `cargo fmt` in CI so it no longer introduces (completely) unrelated changes. The codebase is pretty consistently formatted already, so this isn't a big change.


Martin1887:
Good idea, thanks!

PR #976, opened 2024-10-29
Factor duplicate code out of parsers

notriddle:
This PR has miscellaneous refactoring, to reduce duplicate source code and generated code size:

* It introduces a couple helper functions, `skip_container_prefixes` and `scan_next_line_or_lazy_continuation`, to reduce code duplication.
* It passes `Options` itself down instead of `bool`, to prevent some potentially very easy mixups.
* It lifts some bound checked slices outside the loop, particularly in cases where I checked (using `cargo-show-asm`) that the compiler couldn't see it.


Martin1887:
Thanks!

Issue #973, opened 2024-10-22
pulldown-cmark panics on first pass when enabling `definition-lists`


ytmimi:
I'm not sure if nested definition lists are valid, but using [pulldown-cmark 0.12.2](https://github.com/pulldown-cmark/pulldown-cmark/tree/v0.12.2) and the following input causes a panic:

```markdown
level one!
:
    level two!
    :
        level three!
        :

new level one!
:
```

Here's a small program that reproduces the issue:
```rust
use pulldown_cmark::{Options, Parser};

static INPUT: &str = "
level one!
:
    level two!
    :
        level three!
        :

new level one!
:
";

fn main() {
    let options = Options::ENABLE_DEFINITION_LIST;
    let parser = Parser::new_ext(INPUT, options);
    for _ in parser {}
}

```

<details>
<summary>Traceback</summary>

```
thread 'main' panicked at /Users/ytmimi/.cargo/registry/src/index.crates.io-6f17d22bba15001f/pulldown-cmark-0.12.2/src/firstpass.rs:1540:38:
called `Option::unwrap()` on a `None` value
stack backtrace:
   0: rust_begin_unwind
             at /rustc/e7c0d2750726c1f08b1de6956248ec78c4a97af6/library/std/src/panicking.rs:665:5
   1: core::panicking::panic_fmt
             at /rustc/e7c0d2750726c1f08b1de6956248ec78c4a97af6/library/core/src/panicking.rs:74:14
   2: core::panicking::panic
             at /rustc/e7c0d2750726c1f08b1de6956248ec78c4a97af6/library/core/src/panicking.rs:148:5
   3: core::option::unwrap_failed
             at /rustc/e7c0d2750726c1f08b1de6956248ec78c4a97af6/library/core/src/option.rs:2012:5
   4: core::option::Option<T>::unwrap
             at /Users/ytmimi/.rustup/toolchains/nightly-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/option.rs:972:21
   5: pulldown_cmark::firstpass::FirstPass::pop
             at /Users/ytmimi/.cargo/registry/src/index.crates.io-6f17d22bba15001f/pulldown-cmark-0.12.2/src/firstpass.rs:1540:22
   6: pulldown_cmark::firstpass::FirstPass::parse_block
[...]

PR #980, opened 2024-11-06
Ensure "parse" fuzz target covers all options

ollpu:
The fuzz target `parse.rs` (intended to find crashes/panics) was missing some newer options.

Change to use `Options::from_bits_truncate` to generate arbitrary combinations of all options, current and future.

Not sure if the field order in the struct makes much difference, but I figured it's better to have the fixed length field first, so it doesn't move around and change unintentionally.

Tried it out, and it found #973 in under a minute (on v0.12.2 before the fix was merged).

### DIFF
--- a/fuzz/fuzz_targets/parse.rs
+++ b/fuzz/fuzz_targets/parse.rs
@@ -2,60 +2,15 @@
 use libfuzzer_sys::fuzz_target;
 
 use libfuzzer_sys::arbitrary::{self, Arbitrary};
-use pulldown_cmark::Options;
 
 #[derive(Debug, Arbitrary)]
 struct FuzzingInput<'a> {
+    options: u32,
     markdown: &'a str,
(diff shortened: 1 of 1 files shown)


ollpu:
Forgot to mention this does preclude using the `-only_ascii=1` flag with libfuzzer (all option combinations were possible when they were bools but now they're not). Shouldn't make much difference, but I haven't played with it that much.

ollpu:
@kdarkhan in the README example you had put `-only_ascii=1`. Have you found that to be helpful?

If we really wanted, it could be supported while keeping this option-agnostic via something like `[bool; Options::all().bits().count_ones()]`.

kdarkhan:
I haven't really tested it much with fuzzing this project. I put the option only as an example of fixing the build and not as the suggested option. Feel free to remove it if it is misleading now.

PR #981, opened 2024-11-06
Add `utils::character_reference`

SabrinaJewson:
I am a consumer of this crate, but I’m not outputting HTML, and I would like to transform `InlineHtml` and `HtmlBlock` nodes according to my own needs. Writing a small HTML parser myself is easy enough, but `pulldown-cmark` already has [the large table required to resolved named HTML entities](https://github.com/pulldown-cmark/pulldown-cmark/blob/master/pulldown-cmark/src/entities.rs), and for binary size’s sake I’d prefer to avoid duplicating all of this in my own code.

Therefore, this feature request exposes the functionality of resolving HTML character references in the public API of the crate, via the `utils` module.

While it may be nice to expose more HTML parsing, such an API would be more subjective and would be hard to make fit everybody’s needs. By contrast, this function is minimal and self-contained, and has legitimate disadvantages to users rolling their own versions of.

### DIFF
--- a/pulldown-cmark/src/firstpass.rs
+++ b/pulldown-cmark/src/firstpass.rs
@@ -1093,7 +1093,7 @@ impl<'a, 'b> FirstPass<'a, 'b> {
                     LoopInstruction::ContinueAndSkip(0)
                 }
                 b'&' => match scan_entity(&bytes[ix..]) {
-                    (n, Some(value)) => {
+                    Some((n, value)) => {
                         self.tree.append_text(begin_text, ix, backslash_escaped);
                         backslash_escaped = false;
                         self.tree.append(Item {
@@ -1104,7 +1104,7 @@ impl<'a, 'b> FirstPass<'a, 'b> {
                         begin_text = ix + n;
                         LoopInstruction::ContinueAndSkip(n - 1)
                     }
-                    _ => LoopInstruction::ContinueAndSkip(0),
+                    None => LoopInstruction::ContinueAndSkip(0),
                 },
                 b'|' => {
                     if ix != 0 && bytes[ix - 1] == b'\\' {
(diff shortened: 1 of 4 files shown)


ollpu:
Fair enough, I agree that this is something you don't necessarily want to implement yourself.

Just for binary size it might be better that we split the data into a separate crate or use the existing https://github.com/p-jackson/entities.

The implementation is also rather specific to CommonMark and there are a couple cases that behave differently from HTML.. In some use cases it _could_ be that the function is not safe for parsing actual HTML. I think the relevant [chapter](https://spec.commonmark.org/0.31.2/#entity-and-numeric-character-references) of the spec should be mentioned in the doc comment, or even as part of the function name.

SabrinaJewson:
I see – you’re right about CommonMark and HTML differing. In that case, in the interests of not wanting to design a HTML parser, I don’t think this function is the best way forward. Are you okay with an `entities` dep, then?

SabrinaJewson:
Hmm. `entities`’ data isn’t sorted, meaning we can’t do a binary search like we currently do. I did some benchmarks:

- binary search is 83ns
- std hashmap is 16ns
- `rustc_hash` hashmap is 5ns
- `foldhash-f` and `foldhash-q` are 6.5 and 7ns respectively
- BTreeMap is 45ns
- PHF hashmap is 15ns

Building the `rustc_hash` map takes 38μs, and it takes up 128KB in memory (cf. `entities::ENTITIES` takes up 105KB in the array and 25KB in static strings; the current impl takes 68KB in the array and 20KB in static strings). The size of `entities::ENTITIES` could be reduced by feature-flagging (which I opened [an issue on that repo](https://github.com/p-jackson/entities/issues/13) for).

There’s a tradeoff, I suppose. I opened https://github.com/p-jackson/entities/issues/14 to see what the maintainer of that crate is interested in.

ollpu:
Yeah, on closer look it might not be a great fit for us. In addition to the different sorting and extra codepoint data, it contains some of the entities both with and without `;`, which isn't needed for CM. So binary size might be something like 3­–4x compared to the current table.

PR #985, opened 2024-11-14
Fix typo in README

osa1:


osa1:
If "drive" here isn't a typo then the paragraph structure becomes a bit weird.

It starts by talking about pull parsing:

> Pull parsing is in some sense the most versatile architecture.

Then abrubtly switches to a property of push parsing:

> It's possible to drive a push interface ...

Then back to pull parsing again:

> Another advantage is ...

With "drive" replaced as "derive", it talks about properties of pull parsing, so the paragraph becomes more coherent.

Feel free to close if you disagree, but I think the paragraph reads a bit weird right now.

ollpu:
Matter of taste, but the current form reads OK to me. It is a property of pull parsing that you can also use it to _drive_ a push interface. The converse is not true: you can't easily ask `.next()` from a push parser, without collecting all events first.

"derive" also makes sense (albeit with subtly different meaning) and I don't disagree the paragraph could be written more clearly overall.

I'll leave this up to another maintainer

Issue #745, opened 2023-10-23
Better developers' documentation


Martin1887:
The current docs for developers are basic and more details about each algorithm should be very useful for new contributors.

systemsoverload:
Got a starting point in mind? I've been staring at the source for a week now and feel like attempting to explain some if it would be a good way to further my own understanding

Martin1887:
Great! I think that the best approach is adding a chapter for developers in the mdBook, with an overview of the architecture and then explaining each aspect of the parser in a different chapter.

systemsoverload:
Threw my notes into markdown here https://github.com/pulldown-cmark/pulldown-cmark/pull/986

PR #986, opened 2024-12-05
#745 - Add basic skeleton for developer docs

systemsoverload:
Create a reasonable skeleton for important topics to cover and write a bit about the various subjects. This mostly note-taking levels of detail, but I think a reasonable starting off point. I'm open to any and all suggestions from copy-edits to missing/incomplete/incorrect data.


systemsoverload:
Whoops that was unintentional. I reset my fork for the HTML rendering PR 🤦

PR #987, opened 2024-12-09
refactor: split HTML renderer into processor and renderer components

systemsoverload:
This commit introduces a significant architectural improvement to the HTML rendering system by splitting it into two main components:

1. HtmlEventProcessor: Handles event processing and document structure
2. HtmlElementRenderer: Defines the interface for HTML element rendering

Key changes:
- Move HTML rendering logic into a trait for better abstraction
- Export all of these in the `html` module of the crate
- Create a default implementation (HtmlWriter) of the renderer interface
- HtmlWriter/push_html should behave exactly the same as before

The refactoring grants an easy way for third parties to create their own extended impl of `push_html` that provides a custom event iterator, html renderer, or both.

This will specifically make writing external HTML rendering crates much simpler as `pulldown-cmark` can continue to own the default html rendering implementation which these crates can use as their own defaults.


Martin1887:
Interesting, but this is a pretty big change and it requires a closer look about how to handle this (at a first look it seems too much new code, performance implications, etc.). Thanks!

systemsoverload:
Im relatively new to rust so Im definitely open to any kind of feedback that you might have that could get something like the end goal (an exposed trait) without the implications you've called out above. If it's just not that compelling to you though, feel free to close the issue - no sore feelings for an RFC to which the response is "no thank you" :)

PR #988, opened 2024-12-10
Add basic skeleton for developer docs

systemsoverload:
Create a reasonable skeleton for important topics to cover and write a bit about the various subjects. This mostly note-taking levels of detail, but I think a reasonable starting off point. I'm open to any and all suggestions from copy-edits to missing/incomplete/incorrect data.

### DIFF
--- a/guide/src/dev/html-generation.md
+++ b/guide/src/dev/html-generation.md
@@ -1,0 +1,177 @@
+# HTML Generation
+
+This chapter explains how pulldown-cmark generates HTML output from Markdown events.
+
+## Overview
+
+HTML generation is implemented in the `html` module and consists of two main components:
+
+1. The `HtmlWriter` struct which manages state and writes HTML tags
+2. Helper functions for converting events to HTML tags and handling special cases
+
+The HTML generation process works by:
+1. Taking an iterator of Markdown events 
+2. Converting each event into corresponding HTML tags
+3. Managing state for special cases like tables and tight lists
+4. Writing the HTML tags to the provided output
+
+## The HtmlWriter
+
+The core type responsible for HTML generation is `HtmlWriter`:
+
+```rust
+struct HtmlWriter<'a, I, W> {
+    iter: I,        // Iterator supplying events
+    writer: W,      // Writer to write to
+    end_newline: bool,  // Whether last write ended with newline
+    in_non_writing_block: bool,  // In metadata block (no output)
+    table_state: TableState,  // Current state for table processing
+    table_alignments: Vec<Alignment>,  // Column alignments for current table
+    table_cell_index: usize,  // Current cell index in table row
+    numbers: HashMap<CowStr<'a>, usize>,  // For footnote numbering
+}
+```
+
+The writer keeps track of:
+
+- The current table state (head vs body)
+- Table column alignments
+- Current cell index
+- Footnote numbering
+- Whether we're in a non-writing block like metadata
+- Whether the last write ended with a newline
+
+## Event Processing
+
+The main event processing loop lives in `HtmlWriter::run()`. For each event:
+
+1. The event is matched and dispatched to the appropriate handler
+2. HTML tags are written based on the event type
+3. State is updated as needed
+
+Key event handling patterns:
+
+### Block Elements
+
+Block elements like paragraphs, headings, lists etc. are wrapped in HTML tags:
+
+```rust
+match event {
+    Start(Tag::Paragraph) => write("<p>"),
+    End(EndTag::Paragraph) => write("</p>\n"),
+    // etc
+}
+```
+
+### Inline Elements
+
+Inline elements like emphasis and links are handled similarly but without newlines:
+
+```rust 
+match event {
+    Start(Tag::Emphasis) => write("<em>"),
+    End(EndTag::Emphasis) => write("</em>"),
+    // etc
+}
+```
+
+### Text Content
+
+Text content is HTML escaped and written directly:
+
+```rust
+match event {
+    Text(text) => escape_html_body_text(&mut writer, &text),
+    // etc
+}
+```
+
+### Complex Elements
+
+More complex elements like tables require managing state:
+
+```rust
+match event {
+    Start(Tag::Table(alignments)) => {
+        self.table_alignments = alignments;
+        self.write("<table>")?;
+    }
+    // etc
+}
+```
+
+## HTML Safety
+
+The functions `escape_html()` and ``escape_href()`` are used throughout the library for escaping special characters. The escaping functions live in the `pulldown-cmark-escape` crate.
+
+## Writer Interface
+
+The HTML writer is generic over the writer type `W`, allowing output to:
+
+- Strings via `fmt::Write`
+- Files/IO via `io::Write`
+
+This generic design lets users choose the most efficient output method for their use case. For example:
+- Using `String` is convenient for in-memory processing and testing
+- Using `BufWriter<File>` is efficient for writing directly to disk
+- Using a network socket allows streaming HTML over a connection
+- Using a custom writer enables special handling like compression or logging
+
+The `StrWrite` trait provides a common interface to abstract over these different writers:
+
+```rust
+pub trait StrWrite {
+    type Error;
+    fn write_str(&mut self, s: &str) -> Result<(), Self::Error>;  
+}
+```
+
+This abstraction over the writer type means the HTML generation code can focus on correct tag generation and structure without worrying about the specific output destination. It also allows users to easily integrate pulldown-cmark's HTML output into their existing I/O pipelines.
+
+## Public API
+
+The main public API consists of:
+
+```rust
+// Write HTML to a String
+pub fn push_html<'a, I>(s: &mut String, iter: I) 
+where I: Iterator<Item = Event<'a>>
+
+// Write HTML to an IO writer
+pub fn write_html_io<'a, I, W>(writer: W, iter: I) -> io::Result<()> 
+where I: Iterator<Item = Event<'a>>,
+      W: io::Write
+
+// Write HTML to a fmt writer
+pub fn write_html_fmt<'a, I, W>(writer: W, iter: I) -> fmt::Result
+where I: Iterator<Item = Event<'a>>,
+      W: fmt::Write
+```
+
+## Performance Considerations
+
+HTML generation aims to be efficient by:
+
+1. Minimizing string allocations
+2. Using buffered writers
+3. Avoiding recursion in the core loop
+
+Note:
+
+```rust 
+// Using unbuffered writers (like Files) will be slow
+// Wrap them in BufWriter for better performance
+let file = BufWriter::new(File::create("output.html")?);
+write_html_io(file, parser);
+```
+
+This ensures good performance even with large documents.
+
+## Customization
+
+The HTML output can be customized by:
+
+1. Using a custom writer implementation
+2. Preprocessing the event stream 
+3. Post-processing the HTML output
+4. Using the parser options to enable/disable features
--- a/guide/src/dev/index.md
+++ b/guide/src/dev/index.md
@@ -1,0 +1,62 @@
+# Developer Guide
+
+pulldown-cmark uses a two-pass parsing strategy with a pull parser architecture to efficiently parse Markdown into HTML. This guide explains the internal workings of the library for developers who want to contribute or better understand how it works.
+
+## High-Level Architecture
+
+The parser operates in two main passes:
+
+1. **First Pass (Block Structure)**: The first pass scans the document and builds a tree structure representing block-level elements like paragraphs, lists, code blocks, etc. This establishes the hierarchical structure of the document.
+
+2. **Second Pass (Inline Processing)**: The second pass processes inline elements like emphasis, links and code spans within the blocks identified by the first pass. This is done in a streaming fashion as events are requested.
+
+The library uses a pull parser design, which means:
+
+- Instead of pushing events to a callback or building a complete AST, it provides an iterator interface that lets consumers pull events as needed
+- It enables flexible transformation of the event stream before rendering
+
+Key components:
+
+- `Parser`: The main entry point that implements the Iterator trait for Events
+- `Tree`: A Vec-based data structure that holds the block structure 
+- `Event`: An enum representing the different Markdown elements
+- `HtmlWriter`: Renders the event stream as HTML
+
+## Performance Characteristics
+
+The parser is designed for high performance:
+
+- Performance is intended to be linear with respect to the size of the input text
+- String handling uses copy-on-write semantics to avoid unnecessary allocations
+- SIMD optimizations are available for scanning text on x86_64
+
+## Extending the Parser
+
+The parser can be extended in several ways:
+
+- New syntax extensions can be added by implementing new scan functions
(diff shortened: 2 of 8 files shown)


ollpu:
Thanks! This is quite comprehensive already and the structure makes sense. I'm about halfway through reviewing.

On writing style, some parts feel sales-pitch-y in a way that I don't think is relevant in developer documentation. There are some inaccuracies that can be adjusted but also some nonsensical statements. (Did you use an LLM to help writing? That's fine, but I'd appreciate transparency on that)

systemsoverload:
Yeah the outline was generated with my dev notes and claude so the tone is classically LLM'd (sales pitch-y is a nice way to put it). I did a few of slop cleanup but I'll do another to tighten it up.

ollpu:
Alright! In that case I'd also cut down on the length of some of the pages especially around the "recommendations" and "design principles". They're reasonable for the most part, but too generic to add much value IMO.

systemsoverload:
Gave it another once over and removed some of the overly flowery blocks. It reads as a reasonable first-time-dev resource to me, but please let me know what else you'd like to see changed. Thanks!

guide/src/dev/html-generation.md
```diff
@@ -0,0 +1,181 @@
...
+
+The functions `escape_html()` and ``escape_href()`` are used throughout the library for escaping special characters. The escaping functions live in the `pulldown-cmark-escape` crate.
+
+## Writer Interface
+
+The HTML writer is generic over the writer type `W`, allowing output to:
+
+- Strings via `fmt::Write`
+- Files/IO via `io::Write` 
+- Custom writer implementations
```
ollpu, line 114: Custom writers are not accepted by the public API of the HTML renderer, although the `StrWrite` abstraction is exposed in `pulldown-cmark-escape`. `io/fmt::write` should be sufficient for most use cases like this.

guide/src/dev/html-generation.md
```diff
@@ -0,0 +1,181 @@
...
+
+## Performance Considerations
+
+HTML generation aims to be efficient by:
+
+1. Minimizing string allocations
+2. Using buffered writers
+3. Avoiding recursion in the core loop
+4. Pre-allocating common strings
+5. Reusing string buffers where possible
```
ollpu, line 161: 4 & 5 don't make sense to me

guide/src/dev/performance.md
```diff
@@ -0,0 +1,206 @@
...
+Key optimizations in the tree structure include:
+
+- Using indices instead of pointers for node references
+- Maintaining a "spine" for fast access to ancestor nodes
+- Storing nodes contiguously in a vector for better cache usage
+- Using non-zero indices to save space in option types
+
+## Single-Pass Parsing
+
+The parser uses a single pass over the input text for both block and inline parsing, avoiding the need to rescan text multiple times. This is achieved through:
```
ollpu, line 88: Untrue - there are multiple passes (two principle ones but also many smaller ones depending on how you look at it)

guide/src/dev/performance.md
```diff
@@ -0,0 +1,206 @@
...
+    if scan_ch(&bytes[i..], b':') == 0 {
+        return None;
+    }
+    i += 1;
+    let (bytecount, link_def) = self.scan_refdef(start, start + i)?;
+    Some((bytecount + i, UniCase::new(label), link_def))
+}
+```
+
+## Lazy Text Merging
```
ollpu, line 113: This is not a performance optimization. It is to work around the performance optimization that Text events are often split into multiple parts for various reasons. (Zero-copy ranges may be disjoint & parsed special characters are cheaper to keep in `InlineStr` rather than concatenating into a new allocation.)

You may want to use `TextMergeStream` if you'd rather handle the consecutive events "merged" together, but it has a performance cost.

guide/src/dev/performance.md
```diff
@@ -0,0 +1,206 @@
...
+        }
+    }
+}
+```
+
+This avoids unnecessary string allocations and concatenations when processing text nodes.
+
+## Protection Against Pathological Input
+
+Several protections are in place to prevent exponential time or memory usage on malicious input:
```
ollpu, line 141:
```suggestion
Several protections are in place to prevent quadratic time or memory usage on malicious input:
```

guide/src/dev/performance.md
```diff
@@ -0,0 +1,206 @@
...
+
+3. Avoiding string copies when references can be used instead
+
+## Key Performance Considerations
+
+When using the library, keep in mind:
+
+1. SIMD optimizations require the `simd` feature and x86_64 platform
+2. Large documents benefit most from SIMD scanning
+3. String allocation can be minimized using `TextMergeStream`
```
ollpu, line 186: no...

guide/src/dev/performance.md
```diff
@@ -0,0 +1,206 @@
...
+## Memory Usage Optimization
+
+Several techniques are used to minimize memory usage:
+
+1. Using non-zero indices to save space in option types:
+```rust
+pub(crate) struct TreeIndex(NonZeroUsize);
+```
+
+2. Reusing allocations where possible:
```
ollpu, line 169: I don't think `cows` and `links` are really reused. The main purpose of these is to keep tree nodes small and lifetime-free (index over pointer). The latter is not really necessary, `Tree<Item<'input>>` would work too. Some other fields in `Allocations` are reused I suppose.

guide/src/dev/performance.md
```diff
@@ -0,0 +1,206 @@
...
+4. The parser is designed for streaming, allowing incremental processing
+5. Pathological input protection may limit processing of extremely nested or repetitive content
+
+## Benchmarking
+
+The library includes benchmarks to measure performance of key operations:
+
+- Character scanning with and without SIMD
+- String handling with different storage strategies
+- Tree operations
```
ollpu, line 196: These would make sense I guess but we don't have them.

guide/src/dev/index.md
```diff
@@ -0,0 +1,64 @@
...
+The parser operates in two main passes:
+
+1. **First Pass (Block Structure)**: The first pass scans the document and builds a tree structure representing block-level elements like paragraphs, lists, code blocks, etc. This establishes the hierarchical structure of the document.
+
+2. **Second Pass (Inline Processing)**: The second pass processes inline elements like emphasis, links and code spans within the blocks identified by the first pass. This is done in a streaming fashion as events are requested.
+
+The library uses a pull parser design, which means:
+
+- Instead of pushing events to a callback or building a complete AST, it provides an iterator interface that lets consumers pull events as needed
+- This allows for memory-efficient processing since only a small part of the document needs to be in memory at any time
```
ollpu, line 16: The whole input string and block structure tree do have to be kept in memory.

guide/src/dev/index.md
```diff
@@ -0,0 +1,64 @@
...
+- `HtmlWriter`: Renders the event stream as HTML
+
+## Performance Characteristics
+
+The parser is designed for high performance:
+
+- Memory usage is O(n) in the nesting depth of blocks, not document size
+- String handling uses copy-on-write semantics to avoid unnecessary allocations
+- SIMD optimizations are available for scanning text on x86_64
+- The pull parser design allows streaming without buffering the whole document
```
ollpu, line 33: First and last point are untrue

guide/src/dev/index.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Extending the Parser
+
+The parser can be extended in several ways:
+
+- New syntax extensions can be added by implementing new scan functions
+- The event stream can be transformed using Iterator adaptors
+- Custom renderers can be built by consuming events
+- The HTML renderer can be customized through options
```
ollpu, line 42: Not really, options only affect parsing ATM

guide/src/dev/index.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Directory Structure
```
ollpu, line 44: This might be a good place to mention the higher level directory structure too, not just the main crate under `pulldown-cmark/`.

ollpu (commented):
Tried not being too pedantic, but here we are :slightly_smiling_face:

systemsoverload:
Excellent! Thanks for the feedback, pedantry is always welcome when talking docs. Will take another pass at the branch this morning.

guide/src/dev/inline-processing.md
```diff
@@ -0,0 +1,233 @@
...
+
+Key performance features:
+
+1. SIMD-accelerated character scanning
+2. Efficient delimiter matching
+3. Minimal string allocation
+4. Lazy processing of inlines
+5. Smart caching of partial results
+
+The inline processor aims to be:
```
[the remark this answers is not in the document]
systemsoverload: Ive removed that block, but that feels worth mentioning in the performance and/or extensions guide

guide/src/dev/string-handling.md
```diff
@@ -0,0 +1,145 @@
...
+## Performance Considerations
+
+The string system is designed for the performance characteristics needed by a Markdown parser:
+
+- Minimal copying of input text
+- Efficient handling of many small string fragments
+- Quick concatenation for things like text merging
+- Fast comparisons for link matching
+
+The `TextMergeStream` utility showcases these optimizations by efficiently combining consecutive text events while preserving memory efficiency.
```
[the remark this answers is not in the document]
systemsoverload: I wanted to call out `TextMergeStream` somewhere in here, but maybe its fine to omit as it is documented pretty extensively elsewhere.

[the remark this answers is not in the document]
ollpu: Yeah, I think you could mention the "separate text events" as one optimization, which the user can opt out of with `TextMergeStream` if desired. F

guide/src/dev/performance.md
systemsoverload: Rewrote to integrate your suggestion WRT emphasizing linear performance:

``` ## Protection Against Pathological Input

It is important that the parser performance remain linear with respect to the input length, otherwise the parser would find itself vulnerable to potential DOS attacks. This may not be important for all consumers, but for anyone depending on the library to handle user generated content this is critical.

Several protections are in place to prevent quadratic time or memory usage on malicious input:```

File: guide/src/dev/index.md (hunk 11 above)
Comment (systemsoverload) [reply guide/src/dev/index.md:42]: This was one of several issues caused by me working (accidentally) on an ancient version of the library for an attempt at HTML parsing improvements. Removed.

File: guide/src/dev/html-generation.md (hunk 1 above)
Comment (systemsoverload) [reply guide/src/dev/html-generation.md:115]: How about something like:

` - Custom writer implementations via implementing the `StrWrite` trait`

This is the bit of the crate that I was personally the most interested int and it wasn't clear to me using the existing docs.

**Comment by systemsoverload** [comment]
Accepted your suggestions, made some more copy edits, and rebased everything together. I only have the one remaining question as to how (if at all) you think we should document the intent of the `StrWrite` trait.

**Comment by ollpu** [comment]
Thanks! I'm not seeing the rebase yet though, and it looks like the latest commit undoes the changes.

> How about something like:
>
> - Custom writer implementations via implementing the StrWrite trait
> 
> This is the bit of the crate that I was personally the most interested int and it wasn't clear to me using the existing docs.

That isn't possible right now though. The only public entry points for the HTML module are:

```rust
pub fn push_html<'a, I>(s: &mut String, iter: I) {...}
pub fn write_html_fmt<'a, I, W: std::fmt::Write>(writer: W, iter: I) {...}
pub fn write_html_io<'a, I, W: std::io::Write>(writer: W, iter: I) {...}
```

`StrWrite` is mainly meant to be an adapter trait. An implementation detail we'd rather not have to expose. If you look at the definition for [`std::fmt::Write`](https://doc.rust-lang.org/stable/core/fmt/trait.Write.html), it is very close to `StrWrite`:

```rust
pub trait Write {
    // Required method
    fn write_str(&mut self, s: &str) -> Result;

    // Provided methods
    fn write_char(&mut self, c: char) -> Result { ... }
    fn write_fmt(&mut self, args: Arguments<'_>) -> Result { ... }
}
```

So for all intents and purposes, a user of pulldown-cmark should be able to implement `fmt::Write` for their custom type and use that instead of `StrWrite`.

The reason we want `StrWrite` at all is it makes the `Error` type customizable as an associated type, so we can pass through both `std::fmt::Result` and `std::io::Result` without losing anything.

Reading the [docs](https://doc.rust-lang.org/std/fmt/struct.Error.html) for `std::fmt::Error`, I notice it mentions that the way IO errors are passed through formatting in the standard library is [with an adapter that stores the error](https://doc.rust-lang.org/src/std/io/mod.rs.html#1821):

```rust
struct Adapter<'a, T: ?Sized + 'a> {
    inner: &'a mut T,
    error: Result<()>,
}
```

We could consider something similar for pulldown-cmark in the future as well.

Previous discussions:
https://github.com/pulldown-cmark/pulldown-cmark/issues/492
https://github.com/pulldown-cmark/pulldown-cmark/pull/493
https://github.com/pulldown-cmark/pulldown-cmark/pull/870
https://github.com/pulldown-cmark/pulldown-cmark/pull/881

TL;DR: `StrWrite` is an implementation detail to abstract over the `fmt` and `io` machinery, since they have different error types. Users can provide custom writers by implementing either `fmt::Write` or `io::Write`, whichever is more appropriate.

systemsoverload:
Unrelated to this PR, that actually cleared up some mystery around the `W` generic 👍

ollpu:
Indeed, I had to read through those prior discussions again to remind myself of the nuances

systemsoverload:
Actually pushed the rebase now ;)

guide/src/dev/inline-processing.md
```diff
@@ -0,0 +1,149 @@
+# Inline Processing
+
+The second pass of pulldown-cmark's parsing process handles inline elements like emphasis, links, and code spans.
```
ollpu, line 3: Forgot this, for completeness
```suggestion
The second pass of pulldown-cmark's parsing process handles inline elements like emphasis, links, and code spans.

Corresponds to Phase 2 of ["A Parsing Strategy"](https://spec.commonmark.org/0.31.2/#appendix-a-parsing-strategy).
```

guide/src/dev/performance.md
```diff
@@ -0,0 +1,114 @@
...
+
+The SIMD optimization is implemented in `scanners.rs` and operates by:
+
+1. Creating a lookup table of special characters (like `*`, `_`, etc.)
+2. Loading 16 bytes at a time into a SIMD register
+3. Performing parallel lookups to identify special characters
+4. Generating a bitmask indicating which bytes matched
+
+```rust
+// Example from simd.rs showing the core scanning logic
```
ollpu, line 17: There is no simd.rs

ollpu (approved):
Couple notes still, but I won't hold this up too much more

guide/src/dev/inline-processing.md
ollpu: Actually, we don't really quite use the described algorithms for phase 2, even though it's conceptually doing the same things. I'll just close this thread

Martin1887:
Thanks for the writing and reviewing work! I will try to review this in the next days.

Martin1887:
It looks good to me! I think it can be merged when the last olipu's comment is fixed. Thanks!

systemsoverload:
Corrected that reference to be `firstpass.rs` 👍

PR #991, opened 2024-12-14
Implement extension WikiLinks; `Options::ENABLE_WIKILINKS`

frostu8:
> This was originally a single feature for a highly opinionated flavor of this crate, but after doing some research it seems [there *is* a demand](https://github.com/rambip/pulldown-cmark-wikilink) for MediaWiki-style wikilinks. It still may be out of the scope of this project, though.

This PR implements MediaWiki-style wikilinks. More specifically, it supports the syntax offered by [Obsidian](https://obsidian.md/) and [Python-Markdown WikiLinks](https://python-markdown.github.io/extensions/wikilinks/). The syntax and implementation was chosen to be as close to it as possible, though it may differ in certain edge cases. See the [`pulldown_cmark` wikilinks spec](https://github.com/frostu8/pulldown-cmark/blob/impl-wikilinks/pulldown-cmark/specs/wikilinks.txt) on what specifically this is being checked and tested against.

Because this changes the meaning of certain syntax, this is a exclusive feature gate, culminating in most of the line changes in this PR. The following text:

```markdown
[[Ambiguous]]

[Ambiguous]: https://example.com/
```

happily renders in HTML as:

```html
<p>
[<a href="https://example.com/">Ambiguous</a>]
</p>
```

Enabling `Options::ENABLE_WIKILINKS` makes this render as:

```html
<p>
<a href="Ambiguous">Ambiguous</a>
</p>
```

I'm open to changing this behavior. I only opted to implement it like this because that's how it appears Obsidian handles this syntax, and this extension was originally made for 100% compatibility with Obsidian.

The actual implementation details are a bit squirrely ~~and involve using the `LinkStack` in unintuitive ways~~. This is mostly because I wanted to avoid running another inline pass in `parse.rs` before `handle_inline_pass1` specifically to extract wikilinks before normal Markdown links were processed, while also maintaining the link precedence stated above.


Martin1887:
This seems a good addition to pulldown-cmark, thanks! We have to review the code and look at alternative formats before merging it, though.

frostu8:
> This seems a good addition to pulldown-cmark, thanks! We have to review the code and look at alternative formats before merging it, though.

Definitely could use a code review. There are some strange implementation decisions I wanna work back on before this gets merged, some from not being comfortable with the codebase initially. Looking to see if I have some time to work some things out before the holiday season ends where I'm at.

pulldown-cmark/src/parse.rs
```diff
@@ -2091,6 +2176,22 @@ impl<'a, F: BrokenLinkCallback<'a>> Iterator for OffsetIter<'a, F> {
     }
 }
 
+fn format_wikilink<'a>(text: &'a str) -> CowStr<'a> {
```
notriddle, line 2179: I'm not sure it's a good idea to do this in the parser. If I compare this with my two go-to reference implementations:

- commonmark-hs doesn't normalize wikilinks in its parser, as we can see in the [spec file](https://github.com/jgm/commonmark-hs/blob/master/commonmark-extensions/test/wikilinks_title_before_pipe.md). URL percent-encoding happens in the HTML renderer.
- Gollum (the software behind GitHub's Wiki tab) mixes parsing with normalization, but, because of the way it works, it actually [has to consult its database](https://github.com/gollum/gollum-lib/blob/master/lib/gollum-lib/filter/tags.rb#L250), not just string replace. Meaning you would need more elaborate hook handling to do it anyway.

pulldown-cmark/specs/wikilinks.txt
```diff
@@ -0,0 +1,86 @@
...
+.
+<p>This is <a href="/WikiLink/">a <strong>strong</strong> pothole</a>.</p>
+````````````````````````````````
+
+Or images:
+
+```````````````````````````````` example_wikilinks
+This is a cute dog, linked to the page "/WikiLink/"
+
+[[WikiLink|![dog](dog.png)]]
```
notriddle, line 56: How about a test case for `[[WikiLink|[cat](cat.html)]]` and `[[WikiLink|[[fish]]]]`?

pulldown-cmark/src/parse.rs
```diff
@@ -617,14 +618,97 @@ impl<'input, F: BrokenLinkCallback<'input>> Parser<'input, F> {
...
-                ItemBody::MaybeLinkClose(could_be_ref) => {
+                ItemBody::MaybeLinkClose(double, could_be_ref) => {
                     self.tree[cur_ix].item.body = ItemBody::Text {
                         backslash_escaped: false,
                     };
                     if let Some(tos) = self.link_stack.pop() {
                         if tos.ty == LinkStackTy::Disabled {
                             continue;
                         }
+                        if tos.ty == LinkStackTy::Link(true) {
```
notriddle, line 629: Ordinary markdown links are allowed to start or end with `[[` and `]]`, without the other side also being doubled up (meaning it's not a wikilink).

For example:

```markdown
[[squid] calamari is considered a delicacy](https://en.wikipedia.org/wiki/Squid)

[calamari [squid]](https://en.wikipedia.org/wiki/Squid)
```

> [[squid] calamari is considered a delicacy](https://en.wikipedia.org/wiki/Squid)
>
> [calamari [squid]](https://en.wikipedia.org/wiki/Squid)

This works in commonmark-hs with wikilinks turned on ([playground](https://pandoc.org/try/?params={%22text%22%3A%22[[squid]+calamari+is+considered+a+delicacy](https%3A%2F%2Fen.wikipedia.org%2Fwiki%2FSquid)\n\n[calamari+[squid]](https%3A%2F%2Fen.wikipedia.org%2Fwiki%2FSquid)%22%2C%22to%22%3A%22html5%22%2C%22from%22%3A%22commonmark%2Bwikilinks_title_after_pipe%22%2C%22standalone%22%3Afalse%2C%22embed-resources%22%3Afalse%2C%22table-of-contents%22%3Afalse%2C%22number-sections%22%3Afalse%2C%22citeproc%22%3Afalse%2C%22html-math-method%22%3A%22plain%22%2C%22wrap%22%3A%22auto%22%2C%22highlight-style%22%3Anull%2C%22files%22%3A{}%2C%22template%22%3Anull})), and should work in any implementation that parses wikilinks in a separate pass (assuming they correctly "bail out" if none are found).

I think that you should consider using a separate pass.

pulldown-cmark/src/parse.rs
```diff
@@ -617,14 +618,97 @@ impl<'input, F: BrokenLinkCallback<'input>> Parser<'input, F> {
...
+                                );
+                                self.tree[outer_el.node].item.body = ItemBody::Link(link_ix);
+                                self.tree[outer_el.node].child = Some(body_node);
+                                self.tree[outer_el.node].next = next_node;
+                                self.tree[outer_el.node].item.end = end_ix + 1;
+                                self.link_stack.disable_all_links();
+                            }
+                            // at this point, regardless of whether we were
+                            // successful the stack has been trashed, so use this
+                            // to bring the pass back in a valid state
```
notriddle, line 707: Ordinary markdown links are allowed to start or end with `[[` and `]]`, without the other side also being doubled up (meaning it's not a wikilink).

For example:

```markdown
[[squid] calamari is considered a delicacy](https://en.wikipedia.org/wiki/Squid)

[calamari [squid]](https://en.wikipedia.org/wiki/Squid)
```

> [[squid] calamari is considered a delicacy](https://en.wikipedia.org/wiki/Squid)
>
> [calamari [squid]](https://en.wikipedia.org/wiki/Squid)

This works in commonmark-hs with wikilinks turned on ([playground](https://pandoc.org/try/?params={%22text%22%3A%22[[squid]+calamari+is+considered+a+delicacy](https%3A%2F%2Fen.wikipedia.org%2Fwiki%2FSquid)\n\n[calamari+[squid]](https%3A%2F%2Fen.wikipedia.org%2Fwiki%2FSquid)%22%2C%22to%22%3A%22html5%22%2C%22from%22%3A%22commonmark%2Bwikilinks_title_after_pipe%22%2C%22standalone%22%3Afalse%2C%22embed-resources%22%3Afalse%2C%22table-of-contents%22%3Afalse%2C%22number-sections%22%3Afalse%2C%22citeproc%22%3Afalse%2C%22html-math-method%22%3A%22plain%22%2C%22wrap%22%3A%22auto%22%2C%22highlight-style%22%3Anull%2C%22files%22%3A{}%2C%22template%22%3Anull})), and should work in any implementation that parses wikilinks in a separate pass (assuming they correctly "bail out" if none are found), such as Gollum.

It doesn't work correctly in this PR, when wikilinks are turned on.

pulldown-cmark/src/parse.rs
```diff
@@ -1688,7 +1772,8 @@ struct LinkStackEl {
 
 #[derive(PartialEq, Clone, Debug)]
 enum LinkStackTy {
-    Link,
+    // if this is doubled up, could be a wikilink
+    Link(bool),
```
notriddle, line 1776: Obsidian and Gollum both have ways to create image-style wikilinks.

- https://help.obsidian.md/Linking+notes+and+files/Embed+files
- https://github.com/gollum/gollum/wiki#linking-internal-images

I suspect this is common in most wiki tools.

notriddle (changes requested):

pulldown-cmark/src/parse.rs
frostu8: Yeah, I think I do prefer normalization of wikilinks out of the parser after giving it some thought. Though I might draw the implementation closest to commonmark-hs and forgo any "batteries included," since how these links specifically are handled tends to be very particular for a user's usecase, to the degree that it'd serve most usecases if the parser simply emitted the link as-is with no transformation (exactly how the third-party wrote it).

pulldown-cmark/specs/wikilinks.txt
notriddle, line 1: Another couple of test cases that we probably want:

```markdown
[inline link]([[url]])

[inline link]([[url)]]

`[[code]]`

emphasis **cross [[over** here]]
```

[Pandoc playground](https://pandoc.org/try/?params=%7B%22text%22%3A%22%5Binline+link%5D%28%5B%5Burl%5D%5D%29%5Cn%5Cn%5Binline+link%5D%28%5B%5Burl%29%5D%5D%5Cn%5Cn%60%5B%5Bcode%5D%5D%60%5Cn%5Cnemphasis+**cross+%5B%5Bover**+here%5D%5D%5Cn%22%2C%22to%22%3A%22html5%22%2C%22from%22%3A%22commonmark%2Bwikilinks_title_after_pipe%22%2C%22standalone%22%3Afalse%2C%22embed-resources%22%3Afalse%2C%22table-of-contents%22%3Afalse%2C%22number-sections%22%3Afalse%2C%22citeproc%22%3Afalse%2C%22html-math-method%22%3A%22plain%22%2C%22wrap%22%3A%22auto%22%2C%22highlight-style%22%3Anull%2C%22files%22%3A%7B%7D%2C%22template%22%3Anull%7D)

pulldown-cmark/src/parse.rs
```diff
@@ -831,10 +878,117 @@ impl<'input, F: BrokenLinkCallback<'input>> Parser<'input, F> {
...
+        self.wikilink_stack.clear();
         self.code_delims.clear();
         self.math_delims.clear();
     }
 
+    /// Handles a wikilink.
+    ///
+    /// This function may bail early in case the link is malformed, so this
+    /// acts as a control flow guard. Returns the link node if a wikilink was
+    /// found and created.
```
notriddle, line 890: This seems like a good implementation strategy. I like it! :+1:

It makes it relatively easy to see where the main parser diverges for wikilinks, like having this as its own pass would have, but without potentially generating an improperly nested document (like `<a><a></a></a>`, or, worse, `<a><em></a></em>`).

pulldown-cmark/src/parse.rs
```diff
@@ -831,10 +878,117 @@ impl<'input, F: BrokenLinkCallback<'input>> Parser<'input, F> {
...
+                        .iter()
+                        .position(|b| *b == b'#')
+                        .unwrap_or(wikiname.len())
+                        + start_ix;
+                    // bail early if the wikiname would be empty
+                    if display_ix >= display_end_ix {
+                        return None;
+                    }
+                    // TODO: wikitext should not be styled, might
+                    // need a more experienced contributor's help
```
notriddle, line 954: Can you give an example of where the wrong thing happens? Because it seems like you *have* ignored styling. In the other branch, you seem to have found the existing tree node after the pipe (probably a text node), and made it the child of the wikilink. If there's a MaybeEmphasis in there, it will be retained. In this branch, however, you've completely replaced the children of this tree node with plain text.

As a result, this test case passes on the PR as it stands:

~~~
```````````````````````````````` example_wikilinks
[[**not**]]

[[yes|**yes**]]
.
<p><a href="**not**">**not**</a></p>
<p><a href="yes"><strong>yes</strong></a></p>
````````````````````````````````
~~~

**Review by notriddle — APPROVED** [review]

**Comment by notriddle** [comment]
> it might involve some sort of extra buffer

Unfortunately, I'm not sure a correct version of trimming can be done without an extra buffer *within* pulldown-cmark, either. At least, if you want to have one, you need to account for other features that layer onto it.

~~~markdown
Escaped pipe in wikilink.

```````````````````````````````` example_wikilinks
This is a [[Very\|Wonderful/WikiLink#Secret]].
.
<p>This is a <a href="Very|Wonderful/WikiLink#Secret">WikiLink</a>.</p>
````````````````````````````````

```````````````````````````````` example_wikilinks
This is a [[Very&#33;Wonderful/WikiLink#Secret]].
.
<p>This is a <a href="Very|Wonderful/WikiLink#Secret">WikiLink</a>.</p>
````````````````````````````````
~~~

```console
$ cargo test
...
failures:

---- suite::wikilinks::wikilinks_test_13 stdout ----
thread 'suite::wikilinks::wikilinks_test_13' panicked at pulldown-cmark/src/parse.rs:2328:57:
begin <= end (18 <= 16) when slicing `This is a [[Very\|Wonderful/WikiLink#Secret]].
`

---- suite::wikilinks::wikilinks_test_14 stdout ----
thread 'suite::wikilinks::wikilinks_test_14' panicked at pulldown-cmark/tests/lib.rs:53:5:
assertion `left == right` failed
  left: "<p>This is a <a href=\"Very|Wonderful/WikiLink#Secret\">WikiLink</a>.</p>\n"
 right: "<p>This is a <a href=\"Very&amp;#33;Wonderful/WikiLink#Secret\">Very&amp;</a>.</p>\n"
```

There's a meaningful debate we could have about exactly what the href of these two samples should be, but the first one definitely shouldn't index out-of-bounds, and the second one definitely shouldn't produce `Very&` as the link text.

> and the end user wouldn't know if the wikilink was piped

So include it in the event?

```rust
pub enum LinkType {
...
    /// Wikilink link like `[[destination]]` or `[[destination|pothole]]`
    WikiLink { has_pothole: bool },
}
```

frostu8:
> Unfortunately, I'm not sure a correct version of trimming can be done without an extra buffer _within_ pulldown-cmark, either. At least, if you want to have one, you need to account for other features that layer onto it.

True; I personally believe this is out of scope for the parser, especially because this kind of pulling transformation is what pulldown-cmark was designed in mind with, but just wanted to draw attention to it as a sanity check.

notriddle:
Also, this spec should mention the way backslashed pipes behave: `[[first\|second]]`

The current behavior of this PR is equivalent-ish to `[second](first%5C)`, since the pipe is not escaped and the backslash is treated as part of the URL.

* commonmark-hs (pandoc) [does the same thing](https://pandoc.org/try/?params=%7B%22text%22%3A%22%5B%5Bfirst%5C%5C%7Csecond%5D%5D%22%2C%22to%22%3A%22html5%22%2C%22from%22%3A%22commonmark%2Bwikilinks_title_after_pipe%22%2C%22standalone%22%3Afalse%2C%22embed-resources%22%3Afalse%2C%22table-of-contents%22%3Afalse%2C%22number-sections%22%3Afalse%2C%22citeproc%22%3Afalse%2C%22html-math-method%22%3A%22plain%22%2C%22wrap%22%3A%22auto%22%2C%22highlight-style%22%3Anull%2C%22files%22%3A%7B%7D%2C%22template%22%3Anull%7D).
* Standalone Gollum is roughly equivalent, though its syntax is backwards (pothole first, then destination, meaning it parses like `[first\\](second)`).
* [GitHub Wiki](https://github.com/notriddle/notriddle.github.io/wiki) doesn't treat the backslash as escaping the pipe, either, but drops it on the floor completely, like `[first](second)`.
* Obsidian treats the backslash as a path separator, treating `[[first\|second]]` like `[second](first)`, but when I try a slightly different tactic, like `[[first\,second|third]]`, I get the same result as `[third](first/,second)`.
* python-markdown doesn't seem to support piped pothole-style links at all, so none of this matters.

So, yeah. I don't think we need to change this, but we should definitely mention it in the spec, as well as all this testing I did.

frostu8:
> So, yeah. I _**don't**_ think we need to change this, but we should definitely mention it in the spec, as well as the way other implementations behave.

Got it. The behavior is now defined in the spec as of cac0e51. Thank you for the help!

I also went ahead and tested it on Obsidian Publish since I already use it, and yeah, the backslash is treated as a path. In HTML, `[[first\,third|second]]` renders bizarrely as `<a href="first%5C%2Cthird">second</a>`, with extra attributes and CSR data omitted for clarity. The app correctly links you, but the web publish version does not. [See here](https://publish.obsidian.md/facsimile/Sandbox), although this link will rot after this PR closes/merges.

notriddle (approved):
If @Martin1887 is happy with this, then so am I.

Martin1887:
Good addition to pulldown-cmark, and thanks to both for the great review and implementation cycle!

PR #993, opened 2024-12-19
Change subscript CLI flag to -B

ollpu:
-D conflicts with definition lists

I considered adding a test so this doesn't keep happening, but `getopts` doesn't allow any kind of introspection on the flags, and a separate array might be a little overkill. https://github.com/rust-lang/getopts/pull/116 is the ideal solution but unclear if it's getting merged :(


Martin1887:
Thanks!

PR #1001, opened 2025-01-08
Refactor footnote-rewrite.rs examlpe into using footnote filter

blandger:
An example is refactored into using Footnote filter sctuct for more compact and less verbose usage.


rimutaka:
@blandger , can this converted to a draft until ready for review?
It may be a good idea to squash the commits.

blandger:
@rimutaka Sure, Max. What should I do ??

PR #1007, opened 2025-01-13
docs: Added a doc-comment for ENABLE_SMART_PUNCTUATION option.

rimutaka:
`ENABLE_SMART_PUNCTUATION` setting was not documented.
I added a brief description after looking at the source code in https://github.com/pulldown-cmark/pulldown-cmark/blob/1f36c334d958b6f91eeabefc405c7f86d81c300c/pulldown-cmark/src/firstpass.rs#L1120

Please, double-check if it is full and correct.


Martin1887:
Thanks, good catch, but please fix the `fmt` errors before merging.

PR #1010, opened 2025-01-24
Document more Events and Tags

### LINKED ISSUE #1009: Expand documentation on `Event` e.g. link to reference
As a user of this library only used to writing markdown, some Event kinds can be confusing like `InlineHtml` vs `Html` vs `Start(HtmlBlock)` would help me a lot if there were, e.g., examples of markdown syntax resulting in these events and/or a link to the commonmark spec.

If you are interested in adding these, I'd be willing to write a PR once I fully understand how they work.

ModProg:
fixes #1009

Feel free to close, I mainly looked into it for my self, but maybe this is useful for others as well.

### DIFF
--- a/pulldown-cmark/src/lib.rs
+++ b/pulldown-cmark/src/lib.rs
@@ -160,7 +160,7 @@ pub enum Tag<'a> {
     /// list is chosen, classes are prefixed with `.` and custom attributes
     /// have no prefix and can optionally have a value (`myattr` or `myattr=myvalue`).
     ///
-    /// `id`, `classes` and `attrs` requires [`Options::ENABLE_HEADING_ATTRIBUTES`].
+    /// `id`, `classes` and `attrs` are only parsed and populated with [`Options::ENABLE_HEADING_ATTRIBUTES`], `None` or empty otherwise.
     Heading {
         level: HeadingLevel,
         id: Option<CowStr<'a>>,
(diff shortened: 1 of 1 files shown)


ModProg:
> Nice! The only thing is that fields and variants don't actually require enabling options, but they are always `None` or not emitted when the options are not enabled.

Yes, but if you don't enable them, you don't need to handle them, that is what I meant by that. But I'm happy to adjust that wording if you have a better one!

Martin1887:
I would say something like 'Only relevant if this flag is enabled, always `None` otherwise (or 'Only triggered if this flag is enabled').

ModProg:
Used the wording "Only parsed and emitted with ....`

Martin1887:
Great!, the only thing is the comment for metadata blocks, it seems that you forgot to update it.

pulldown-cmark/src/lib.rs
```diff
@@ -224,6 +282,8 @@ pub enum Tag<'a> {
     },
 
     /// A metadata block.
+==== BASE ====
```
Martin1887, line 285: Wrong doc comment here.

ModProg: ups, not sure how that happened, fixed it.

Martin1887 (commented):
Merged, thanks!

Issue #919, opened 2024-06-29
Custom blockquote tags


Martin1887:
Custom blockquote tags can be a nice addition for some users. Related to https://github.com/pulldown-cmark/pulldown-cmark/issues/718#issuecomment-2113460565.

BrandonDyer64:
Bumping this. My website isn't in English and I'd like to allow people to put custom tags that map into something `pulldown-cmark` understands.

```rust
let parser = parser.map(|event| match event {
    Event::Start(Tag::BlockQuote(kind)) => {
        Event::Start(Tag::BlockQuote(kind.map(|kind| match kind {
            BlockQuoteKind::Unknown("NOTO") => BlockQuoteKind::Note,
            BlockQuoteKind::Unknown("KONSILO") => BlockQuoteKind::Tip,
            BlockQuoteKind::Unknown("GRAVA") => BlockQuoteKind::Important,
            BlockQuoteKind::Unknown("ADMONO") => BlockQuoteKind::Warning,
            BlockQuoteKind::Unknown("AVERTO") => BlockQuoteKind::Caution,
            _ => BlockQuoteKind::Note,
        })))
    }
    _ => event,
});
```

Martin1887:
For this use case, a preprocessing step that translates the tags with a simple `match` makes more sense to me.

BrandonDyer64:
@Martin1887 Isn't `pulldown-cmark` already doing all the parsing? I don't want to have to import another parsing lib just to rewrite the blockquote kind.

I'm also realizing now that the BlockQuote would need to have some way to change the output text as well, to ensure it isn't the auto-generated English.

Martin1887:
The scope of the pulldown-cmark HTML renderer is a simple and working example of HTML rendering, but not a full-fledged renderer.

The creation of another renderer that uses pulldown-cmark in another crate is welcome, though, as reflected in the issue #645.

At the same time, the purpose of the parser is being standard-compatible, with some other standards like Github Flavoured Markdown, but not a parser that supports all custom use cases as this. However, the custom blockquote tags is something easy to consider and it will be probably added soon.

PR #1012, opened 2025-01-25
Add support for more blockquote kinds

cestef:
Another commonly used blockquote kind, I feel like we could even add support for all of them, as listed on obsidian's documentation: https://help.obsidian.md/Editing+and+formatting/Callouts#Supported+types

### DIFF
--- a/pulldown-cmark/specs/blockquotes_tags.txt
+++ b/pulldown-cmark/specs/blockquotes_tags.txt
@@ -7,6 +7,14 @@ Blockquotes can optionally have one of the following tags:
 - [!IMPORTANT]
 - [!WARNING]
 - [!CAUTION]
+- [!FAQ]
+- [!INFO]
+- [!QUOTE]
+- [!SUCCESS]
+- [!FAILURE]
+- [!DANGER]
+- [!BUG]
+- [!EXAMPLE]
 
 Using one of these tags adds a class with the same name but in lowercase
 (note, tip, etc.).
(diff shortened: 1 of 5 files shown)


Martin1887:
Thanks, but some tests for the new kind are required before merging :slightly_smiling_face:.

Martin1887:
You have to add more examples in `pulldown-cmark/specs/blockquotes_tags.txt` and then test with the `--all-features` flag to update the Rust test files.

cestef:
I have added tests as recommended, here is the test output:
[test-results.txt](https://github.com/user-attachments/files/18703745/test-results.txt)

cestef:
Blockquote kinds added in the PR:

- `[!INFO]`
- `[!QUOTE]` 
- `[!SUCCESS]`
- `[!FAILURE]`
- `[!DANGER]`
- `[!BUG]`
- `[!EXAMPLE]`

Martin1887:
Uhm... nice addition, but I'm not sure if some of them are out of the scope of this project, since it aims to only support standard (or semi-standard) features.

cestef:
Thanks for the feedback! Would it make sense to make support for these blockquote kinds optional via a feature flag? That way, the core parser remains focused on standard features while still allowing users who need these kinds to enable them.

Martin1887:
The thing is blockquote tags are already under a feature flag, so nested or complementary flags may be cumbersome.

This change is very small, so it may be included, let's see what the rest of maintainers think about this.

notriddle:
I would really prefer to add support for *custom* blockquote kinds that the host application can postprocess with a callback function (like the Unknown link callback we have now). It would fix #919, work better for i18n, and allow supporting even more complex directives like [DocFx videos](https://github.com/dotnet/docfx/blob/e4d06da4d68ca9d49ad9488a1a124974a885d1e1/test/Docfx.MarkdigEngine.Extensions.Tests/QuoteSectionNoteTest.cs#L170) and [MDN callouts](https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Howto/Markdown_in_MDN#notes_warnings_and_callouts).

Martin1887:
It sounds much better, I agree.

cestef:
This clearly sounds like a better approach indeed

cestef:
The Unknown approach seems good, but there's a problem:

```rs
pub enum BlockQuoteKind<'a> {
	// ...
    Unknown(CowStr<'a>),
}
```

implies `BlockQuoteKind` can't be `Copy` anymore because of the string contained. Which also means that TagEnd can't be either. Would it make sense to implement a "full" enum for `Tag`(non-copy) and keep a basic one for `TagEnd` (copy) ?

Something like:

```rs
pub enum BlockQuoteKind {
    Note,
    Tip,
    Important,
    Warning,
    Caution,
    Unknown,
}

pub enum FullBlockQuoteKind<'a> {
    Standard(BlockQuoteKind),
    Unknown(CowStr<'a>),
}
```

But then what would `ItemBody` hold ? Full or basic ? it seems a bit tricky lol

notriddle:
https://github.com/pulldown-cmark/pulldown-cmark/pull/1013

PR #1013, opened 2025-02-09
Add support for custom block quote admonition kinds

notriddle:
This adds a callback handler that can parse any arbitrary block quote kind, as long as square brackets are correctly nested. You can choose your own data type for the payload, as long as it can be copied, and it should cover obsidion's stuff.


Martin1887:
Interesting, thanks! However, fmt errors should be fixed.

Also, we initially integrated admonitions in a miscellaneous GFM flag, changing it in favor of an admonition flag would be a breaking change (maybe a better approach, we have to discuss it).

notriddle:
- okay, I've fixed the fmt errors
- as for this being a breaking change, I'm pretty sure the new generic param is a breaking change anyway, so it's kind of moot
- now that it supports non-GFM admonition tags, it's not really a GFM feature any more

cestef:
@notriddle `push_html` is missing a type parameter for the underlying `Event<'a>`:

https://github.com/pulldown-cmark/pulldown-cmark/blob/152148e00eeb109e6dd7e4613cbdb16155a6606a/pulldown-cmark/src/html.rs#L550-L555

Which I think forces you to use the default implementation of Event.

cestef:
My idea was to implement a simple `ToClass` trait for admonitions that will be used with `HtmlWriter`:

```rs
pub trait ToClass<'a> {
    fn to_class(&self) -> &'a str;
}

impl<'a, I, W, AD> HtmlWriter<'a, I, W>
where
    I: Iterator<Item = Event<'a, AD>>,
    W: StrWrite,
    AD: ToClass<'a>,
// ...
```` 

The default html conversion then becomes:
```rs
Tag::BlockQuote(kind) => {
    let class_str = kind.map(|k| k.to_class()).unwrap_or("");
    if self.end_newline {
        self.write(&format!("<blockquote{}>\n", class_str))
    } else {
        self.write(&format!("\n<blockquote{}>\n", class_str))
    }
}
```

as for the implementation for the GFM `BlockquoteKind`:

```rs
impl ToClass<'_> for BlockQuoteKind {
    fn to_class(&self) -> &'static str {
        match self {
            BlockQuoteKind::Note => "markdown-alert-note",
            BlockQuoteKind::Tip => "markdown-alert-tip",
            BlockQuoteKind::Important => "markdown-alert-important",
            BlockQuoteKind::Warning => "markdown-alert-warning",
            BlockQuoteKind::Caution => "markdown-alert-caution",
        }
    }
}
```

ollpu:
Apologies for not getting to this sooner.

I'm not really a fan of making `Tag` generic and adding further generics to `Parser`.

Could the callback traits be combined? So `BrokenLinkCallback` is replaced with something like:

```rust
trait ParserCallbacks<'input> {
    fn handle_broken_link(...) -> ...;
    fn handle_admonition_tag(...) -> ...;
}
```

As for avoiding the generic on `Tag`, could we do something like this?

```rust
pub enum BlockQuoteKind<'input> {
    Note,
    Tip,
    Important,
    Warning,
    Caution,
    Custom(CowStr<'input>),
}
```

I imagine putting the raw text (`WARNING`, `MY_CUSTOM_ADMONITION`) in the custom field is usable enough for the majority of use cases.

notriddle:
I like the idea of combining the traits.

I don't like the "Note, Tip, Important, Warning, Caution, Custom(CowStr)" design. If making it generic isn't acceptable, then it seems simpler to pass the string through, even when it's one of GitHub's standard tags.

ollpu:
> pass the string through, even when it's one of GitHub's standard tags.

That sounds alright too

Martin1887:
But if we use just a string, the renderer behaviour to display built-in icons and so would have to match strings instead of enum variants, that does not seem good...

notriddle:
Because the set of supported tags is fixed, we don't actually have to allocate anything. The `'static` lifetime is a subtype of `'input`. And we can define the return value as the class, not just the tag, making the renderer simpler.

```rust
pub trait ParserCallbacks<'input> {
    fn handle_broken_link(
        &mut self,
        link: BrokenLink<'input>,
    ) -> Option<(CowStr<'input>, CowStr<'input>)> {
        None
    }
    fn handle_admonition_tag(
        &mut self,
        tag: &'input str,
    ) -> Option<CowStr<'input>> {
        Some(CowStr::Borrowed(if tag.eq_ignore_ascii_case("note") {
            "markdown-alert-note"
        } else if tag.eq_ignore_ascii_case("tip") {
            "markdown-alert-tip"
        } else if tag.eq_ignore_ascii_case("important") {
            "markdown-alert-important"
        } else if tag.eq_ignore_ascii_case("warning") {
            "markdown-alert-warning"
        } else if tag.eq_ignore_ascii_case("caution") {
            "markdown-alert-caution"
        } else {
            return None;
        }))
    }
}
```

The biggest problem with involving strings in this data structure at all, even if we don't ordinarily use it, comes from producing a valid [TagEnd](https://docs.rs/pulldown-cmark/0.13.0/pulldown_cmark/enum.TagEnd.html). We have to keep that string around to be able to emit it on both events.

ollpu (changes requested):
I suggest we move forward with what @notriddle suggested above, i.e., have the callback return the class name directly.

> The biggest problem comes from producing a valid [TagEnd](https://docs.rs/pulldown-cmark/0.13.0/pulldown_cmark/enum.TagEnd.html). We have to keep that string around to be able to emit it on both events.

But do we have to include the string in `TagEnd`? It may be a ltitle asymmetric, but there is precedent in that `Tag::List(Option<..>)` turns into `TagEnd::List(bool)`. Similarly, we could have `TagEnd::BlockQuote(bool)`. That is enough for the our renderer at least, though it is potentially a challenge in some custom rendering situations.

In any case, I don't think the user-provided callback should be called twice (for both begin and end events). It could be annoying if you want to make it stateful or some kind of expensive lookup. Better to store it in the tree (via `CowIndex`).

PR #1015, opened 2025-02-10
Use an explicit node for tight paragraphs

### LINKED ISSUE #1002: `>` counts as both the start of a blockquote and the end of inline HTML?
Using commit: 2f183d71d12c810754ade2138488f3e2551f40b8.

I'm a little confused about this case, and I think this is an issue with identifying inline HTML. Given the following input:

```markdown
* <c
  >

```

pulldown-cmark considers the `\n>` as part of the `InlineHtml` text and as starting a blockquote.

running `cargo run -- -e` produces these events:

```
0..10: Start(List(None))
0..10: Start(Item)
2..8: InlineHtml(Boxed("<c\n>"))
8..9: Start(BlockQuote(None))
8..9: End(BlockQuote(None))
0..10: End(Item)
0..10: End(List(false))
EOF
```

notriddle:
This cleans up a lot of hacks, because our tree didn't quite match the way commonmark actually works, and we need to avoid letting a block appear as a child of an inline.

It also fixes a bug in HTML parsing that could only be hit with tight lists.

Fixes #1002

### DIFF
--- a/pulldown-cmark/specs/regression.txt
+++ b/pulldown-cmark/specs/regression.txt
@@ -2788,3 +2788,13 @@ the trailing space after the &gt; should be stripped
 .
 <p><a href="Wiki%3C">Link</a></p>
 ````````````````````````````````
+
+```````````````````````````````` example_wikilinks
+* <c
+  >
+.
+<ul>
+<li>&lt;c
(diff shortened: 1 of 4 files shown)


Martin1887:
Thanks!

PR #1026, opened 2025-03-04
Add `no_std` support

CrazyboyQCD:


Martin1887:
Great addition, thanks! However, I miss a specification of the features missing when the `no_std` flag is enabled in the `README.md`.

SabrinaJewson:
You should add this commit as well: https://github.com/SabrinaJewson/pulldown-cmark.rs/commit/4203ce5bd8ef79f67357717d2b728579f3dec294

It sets `default-features = false` for `memchr`, meaning it won’t depend on std anymore

notriddle (approved):

notriddle:
It looks like you’ve mostly just changed import lines.

If it’s helpful, then this change seems okay to me.

Martin1887:
Merged, thanks!

PR #1043, opened 2025-06-11
spoiler with summary, colon fenced

jim-taylor-business:
hello again

this is an attempt at an implementation of spoiler markdown syntax. here's one of the tests from the PR as an example: 

```
::: spoiler Is this expandable?
Is this collapsable?

Is this **bold**?
:::
.
<details>
<summary>Is this expandable?</summary>
<p>Is this collapsable?</p>
<p>Is this <strong>bold</strong>?</p>
</details>
```

it is one of many ideas discussed here:

https://talk.commonmark.org/t/what-could-a-spoiler-tag-extension-look-like/767/58

i am using it myself and wondering if it would be useful to the project. i definitely am not 100% certain that i have implemented it correctly or idiomatically. it does pass all my current tests, of course

thanks and regards


notriddle:
[markdown-it](https://github.com/markdown-it/markdown-it-container?tab=readme-ov-file#example) and [commonmark-hs](https://github.com/jgm/commonmark-hs/blob/master/commonmark-extensions/test/fenced_divs.md) both use this syntax for general purpose custom fenced blocks, not just spoilers.

I think this extension should do the same thing: create a `<div>` with the attribute string as a class by default, and allow an application-defined filter to turn it into whatever other element they want.

pulldown-cmark/src/scanners.rs
```diff
@@ -269,6 +269,26 @@ impl<'a> LineStart<'a> {
         }
     }
 
+    pub(crate) fn scan_spoiler_fence(&mut self) -> bool {
+        if self.scan_case_insensitive(b":::") {
```
notriddle, line 273: commonmark-hs allows you to use three or more opening delimiters, not just three. Please try to implement extensions in a compatible way: we really want to avoid inventing our own syntax wherever feasible.

jim-taylor-business:
> [markdown-it](https://github.com/markdown-it/markdown-it-container?tab=readme-ov-file#example) and [commonmark-hs](https://github.com/jgm/commonmark-hs/blob/master/commonmark-extensions/test/fenced_divs.md) both use this syntax for general purpose custom fenced blocks, not just spoilers.
> 
> I think this extension should do the same thing: create a `<div>` with the attribute string as a class by default, and allow an application-defined filter to turn it into whatever other element they want.

thanks for the information @notriddle 

do you think this very specific use case could exist behind a more specific feature flag?

is there any obvious code issues you can see with this implementation?

pulldown-cmark/src/scanners.rs
```diff
@@ -269,6 +269,26 @@ impl<'a> LineStart<'a> {
...
+            if self.scan_case_insensitive(b" spoiler ") {
```
notriddle, line 274: I think it's a mistake to implement this in a way that `:::spoiler` doesn't work.

jim-taylor-business:
@notriddle - so if i make this PR more generic similar to the  [markdown-it](https://github.com/markdown-it/markdown-it-container?tab=readme-ov-file#example) container extension, and implement the spoiler variation of the container extension, that would be acceptable?

notriddle:
Personally, I think so, but @Martin1887 needs to approve, too.

jim-taylor-business:
thanks for the reply @Martin1887

i think i would like to make it similar to the [markdown-it](https://github.com/markdown-it/markdown-it-container?tab=readme-ov-file#example) container extension and implement spoilers as a specific variant of that so that it fits an existing standard as per @notriddle 's suggestion

does that seem like a good idea to you too?

jim-taylor-business:
i've refactored things as i hope we discussed @notriddle @Martin1887 

please let me know if i need to make changes

pulldown-cmark/src/lib.rs
```diff
@@ -158,6 +158,14 @@ pub enum BlockQuoteKind {
     Caution,
 }
 
+/// ContainerBlock kind (Spoiler only).
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub enum ContainerKind {
+    Spoiler,
+    Example,
+}
```
notriddle, line 167: This isn't what markdown-it and pandoc have.

They allow *any string at all* to appear in the container block line. It's an extension point for the application using this library; we don't want to limit it to two modes.

jim-taylor-business: ah yes i missed the default behaviour in the spec

have changed my `Example` to be a `Default` behaviour

jim-taylor-business:
added an example to the PR summary for the default behavior

ollpu (commented):
Disregard my previous comment. I think it's fine that we accept all blocks (as the "Default" variant), so this feature doesn't need a callback.

Note for merging: we don't have a set policy, but IMO new `Options` should be considered breaking changes (-> 0.14).

Parsing with more than three colons still seems a bit buggy:

```md
:::: spoiler
content
::::
```
gives
```html
<div class="">
<p>content
:</p>
</div>
```

I think it should work more or less the same as fenced code blocks do, i.e. there can be more than than 3 characters, and the closing delimiter must have at least as many as the opening one did. This allows nesting blocks, if desired:

```md
:::: outer
::: inner
content
:::
::::
```

Below are some other divergences I noticed when comparing to commonmark-hs ([Try pandoc](https://pandoc.org/try/) commonmark_x mode) and markdown-it-container. Granted, we can iron out some of these in follow-up PRs as well.

- commonmark-hs does not permit the info string to be empty on the starting line:
  
  <table>
  <tr><th>Input</th><th>commonmark-hs</th><th>This PR</th>
  <tr><td>

  ```
  :::

  content     

  :::
  ```

  </td><td>

  ```
  <p>:::</p>
  <p>content</p>     
  <p>:::</p>
  ```

  </td><td>

  ```
  <div class="">
  <p>content</p>     
  </div>
  ```

  </table>

- commonmark-hs does not permit there to be any extra text on the closing line (other than whitespace). It is instead treated as opening a new block.

  <table>
  <tr><th>Input</th><th>commonmark-hs</th><th>This PR</th>
  <tr><td>

  ```
  ::: block     

  content

  ::: end
  ```

  </td><td>

  ```
  <div class="block">     
  <p>content</p>
  <div class="end">

  </div>
  </div>
  ```

  </td><td>

  ```
  <div class="block">     
  <p>content</p>
  </div>
  <p>end</p>
  ```

  </table>

- Non-meaningful indentation (less than 4) on the closing line starts a new block instead in this implementation

  <table>
  <tr><th>Input</th><th>commonmark-hs</th><th>This PR</th>
  <tr><td>

  ```
   ::: block     
  :::

  ::: block
   :::
  ```

  </td><td>

  ```
  <div class="block">     

  </div>
  <div class="block">

  </div>
  ```

  </td><td>

  ```
  <div class="block"></div>     
  <div class="block">
  <div class=""></div>
  </div>
  ```

  </table>

jim-taylor-business:
> Parsing with more than three colons still seems a bit buggy:
> 
> ```
> :::: spoiler
> content
> ::::
> ```

ah yeah i missed the reference to 3 or more in the spec

jim-taylor-business:
i have updated the PR to reflect all the changes suggested

@ollpu @Martin1887 @notriddle 

with the exception of this test

```
::: block     

content

::: end
```

i'm ok with what this PR generates but i know you like to be compatible

ollpu:
Cool, thanks!

Do you think you can make nesting work as well?

```
:::: a
::: b

:::
::::
```
gives
```
<div class="a">
<p>b</p>
</div>
```
while I expected nested divs

ollpu (commented):
Thanks, and apologies again about the delay. This is starting to look quite robust, but there are still some issues. Nesting is tricky and I guess we don't have any other feature to take inspiration from directly.

Consider this:
```md
::: container
> shouldn't close, right?
> :::
> x  
:::

<div class="container">
<blockquote>
<p>shouldn't close, right?</p>
</blockquote>
</div>
<blockquote>
<p>x
:::</p>
</blockquote>
```

Another thing we try to stay conscious of is time complexity, to make sure untrusted input cannot cause parsing to become too slow. The current implementation seems to be quadratic when nesting containers.

With input generated like this (replace N with some number):
```sh
python -c 'print("\n".join(":"*k + "x" for k in range(N, 2, -1)))'
```
If I double the value of N, the size of the input quadruples (more or less). However, the runtime seems to slow down by ~16x:
| N | time (s) |
| - | - |
| 1000 | 0.32 |
| 2000 | 4.99 |
| 4000 | 85.80 |

That would indicate quadratic run time growth with respect to input size, which we generally want to avoid.

One solution to this is to limit the amount of nesting (or simply colons recognized, before saturating, similar to how `{}` are handled in math spans).

Don't stress these too much. The runtime issue can definitely be fixed later. For now it would be enough to have something that works and is well specified by the spec tests. If nesting proves too difficult, we can also leave it out and try to get it working later.

jim-taylor-business:
thanks for the feedback @ollpu 

will get stuck into it when i can :smiley:

pulldown-cmark/src/scanners.rs
```diff
@@ -507,7 +525,9 @@ pub(crate) fn scan_blank_line(bytes: &[u8]) -> Option<usize> {
 }
 
 pub(crate) fn scan_nextline(bytes: &[u8]) -> usize {
-    memchr(b'\n', bytes).map_or(bytes.len(), |x| x + 1)
+    memchr(b'\n', bytes).map_or(memchr(b'\r', bytes).map_or(bytes.len(), |x| x + 1), |x| {
+        x + 1
+    })
 }
```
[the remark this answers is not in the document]
jim-taylor-business: i have reverted this change so it can be addressed properly as you suggest

jim-taylor-business:
amazing feedback @ollpu  thanks!

> These tests are getting a little abstract aren't they...

the more the merrier :smile: 

i have tried to implement everything you suggested

let me know if there's anything else

notriddle (approved):
This pretty much looks good to me. The test case covers most of the possibilities, and the implementation fits in with the rest of the code.

Other than some low-hanging optimization fruit, it looks fine.

PR #1044, opened 2025-06-15
Remove unused dependency

Expyron:


ollpu (approved):
Good catch, thanks!

ollpu:
Looks like there's an unrelated new warning in nightly

ollpu:
Sorry about the delay. We normally have a policy of requiring two maintainers to approve a PR, but this is so minor that I'll just go ahead and merge.

PR #1045, opened 2025-06-24
Extract non-generic ParserInner

sharnoff:
The current `Parser`'s generics can impact compile times of dependent crates due to each instantiation needing to compile and optimize separate copies of very large and complex functions. That has a [particularly notable effect](https://sharnoff.io/blog/why-rust-compiler-slow#what-are-we-optimizing) when using something like [`cargo-chef`](https://github.com/LukeMathWalker/cargo-chef) — because generics will get compiled as part of the crate where they're monomorphized, rather than in the dependencies they come from.

There are good reasons to keep the `BrokenLinkCallback` as a generic argument (see #701), so to provide the same API while reducing compile times, extract the non-generic bits of the parser into a separate struct. The new `ParserInner`'s methods take a `dyn BrokenLinkCallback`.

Along the way, also unify the iterator impls for `Parser` and `OffsetIter`, so that the logic is not duplicated.

### DIFF
--- a/pulldown-cmark/src/parse.rs
+++ b/pulldown-cmark/src/parse.rs
@@ -191,11 +191,17 @@ pub struct BrokenLink<'a> {
 
 /// Markdown event iterator.
 pub struct Parser<'input, F = DefaultBrokenLinkCallback> {
+    broken_link_callback: Option<F>,
+    inner: ParserInner<'input>,
+}
+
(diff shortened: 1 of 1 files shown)


notriddle (approved):
All of this code seems reasonable, and I agree that using `dyn` for the broken link callback won't cause performance problems in most use cases.

Martin1887:
OK, merged. We will have to create a major release for this breaking change.

PR #1051, opened 2025-08-10
Move benchmark assets to `bench/third_party/`

rhysd:
## Problem

`pulldown-cmark` package delivers the [third party assets in `third_party` directory](https://github.com/pulldown-cmark/pulldown-cmark/blob/7a6179f17c731eaea0fad82b714329985a311a72/pulldown-cmark/Cargo.toml#L17) however some of them are only used by benchmark and unnecessary for users.

Since third party assets have their licenses, users need to audit them.

## Solution

This PR newly creates `bench/third_party` directory and moves the benchmark assets to the directory so that we can reduce the number of licenses in the `pulldown-cmark` package.


ollpu:
> `pulldown-cmark` package delivers the [third party assets in `third_party` directory](https://github.com/pulldown-cmark/pulldown-cmark/blob/7a6179f17c731eaea0fad82b714329985a311a72/pulldown-cmark/Cargo.toml#L17)

That's in the exclude table, so these should not have been delivered to the crates.io package. This change seems ok otherwise, though.

PR #1060, opened 2025-11-14
Refactor test generation to reduce diff noise when adding new options

uimataso:
## Problem

While exploring how to implement my own extension, I noticed that adding a new argument to `tests::test_markdown_html` causes 1000+ lines of changes, because every test cases requires an additional `false` argument. For example, in PR #991, more than half of the diff is just adding an extra `false` to `test_markdown_html`.

So I think it is a good idea to refactor the test generation to reduce the diff noise.

## Solution

Introduce a configuration struct (with `#[derive(Default)]`) to hold all test options. This allows new options to be added without modifying every single test invocation, only the tests that actually depend on the new option need to be updated.


uimataso:
> There's actually already a new class of tests merged into branch-0.14, which would be the next release. Could you change your merge target to that and add the new class introduced in https://github.com/pulldown-cmark/pulldown-cmark/pull/1043?

sure! I'll do that.

> I'm wondering if this needs a separate type, or if we could get away with using Options directly, maybe with a function that gives the "test defaults"?

Yeah, I could use `Options` directly in the generated test, I think this is better too.

ollpu:
Sorry, it's been a while. I have an alternative approach in #1122, which requires no changes to the test generator when adding new Options, and there are no global defaults.

uimataso:
No worries! Feel free to close this PR if this is no longer needed.

PR #1066, opened 2026-01-24
Replace U+0000 with U+FFFD

### LINKED ISSUE #1065: U+0000 must be replaced with U+FFFD to conform CommonMark spec
## Overview

https://spec.commonmark.org/0.31.2/#insecure-characters

> 2.3 Insecure characters
> For security reasons, the Unicode character U+0000 must be replaced with the REPLACEMENT CHARACTER (U+FFFD).

§2.3 specifies that U+0000 must be replaced with U+FFFD. However pulldown-cmark doesn't conform the specification.

## Repro

Check out this repository and run the following command.

```sh
echo -n "\0" | cargo run -- -e
```

## Expected behavior

U+0000 is replaced with U+FFFD

## Actual behavior

The above command outputs

```
0..1: Start(Paragraph)
0..1: Text(Borrowed("\0"))
0..1: End(Paragraph)
EOF
```

U+0000 is not replaced in the output.

rhysd:
Fix #1065

With this patch the following command

```sh
echo -n "\0" | cargo run -- -e
```

outputs

```
0..1: Start(Paragraph)
0..1: Text(Inlined(InlineStr { inner: [239, 191, 189, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0], len: 3 }))
0..1: End(Paragraph)
EOF
```

`239 191 189` is `0xEF 0xBF 0xBD` that is UTF-8 encoded form of U+FFFD.

### DIFF
--- a/pulldown-cmark/specs/regression.txt
+++ b/pulldown-cmark/specs/regression.txt
@@ -2836,3 +2836,16 @@ Link refdef split from paragraph with line with spaces.
 .
 <p>:</p>
 ````````````````````````````````
+
+U+0000 must be replaced with U+FFFD: https://spec.commonmark.org/0.31.2/#insecure-characters
+See https://github.com/pulldown-cmark/pulldown-cmark/issues/1065
+
+```````````````````````````````` example
+ 
+   
+hello world
+.
+<p>�
+���
+hello�world</p>
+````````````````````````````````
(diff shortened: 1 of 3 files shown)


notriddle:
1. In the rare case that someone actually uses a NUL byte, it's probably fine to accept a performance penalty when processing code spans.
2. Aren't we already doing that when replacing CR and LF with spaces?

ollpu:
Merged as part of #1087, thanks!

Issue #1068, opened 2026-02-08
blockquotes that are... not really blockquotes


orangecms:
Hi! I'm coming a long way. please let me elaborate. 🥺

The extended block quotes in GitHub Flavored Markdown (GFM) are being sorta-abused to mark other kinds of blocks, too, such as for various info boxes (note, tip, warning...) - which is what `BlockQuoteKind` stems from, IIUC. 
I'm not sure where this comes from, but I want to take it for granted, and improve the docs and rendering behavior.
At least I don't find it in https://github.github.com/gfm/#block-quotes, though I do know that GitHub supports this.

Currently:
https://github.com/pulldown-cmark/pulldown-cmark/blob/287fcf9c117ca4eb555ecf2cc5e7c131c01d7b24/pulldown-cmark/src/html.rs#L247-L262

i.e., if no "kind" is provided, it renders as a `<blockquote>`, which usually semantically fits.

If a kind (i.e., note, tip, etc., starting with a specific marker, say, `> [!Info]`) _is_ provided, then it is still rendered as a `<blockquote>` in pulldown-cmark, which I find semantically irritating, especially with assistive tooling.
I suggest changing the name of the not-really-quote-blocks to something carrying those semantics, such as _Admonition_, and for rendering, use a `<div>` with `role="note"` (see also https://github.com/rust-lang/mdBook/issues/3026).

For ref, on GitHub, it turns into this, using a `<div>` (see browser dev tools):

> [!Note]
> interesting thing

Martin1887:
It seems reasonable to me, although it would be a breaking change.

orangecms:
I've given it a shot. 🥳

miikkas:
`<aside>`[^1] also seems like it could fit nicely instead of `<div>`[^2]. Especially when content gets nested (e.g. something other than text is inside an admonition), having more `<div>`s increases the risk of having a "div soup".

[^1]: <https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/aside>
[^2]: <https://stackoverflow.com/questions/7579826/which-semantic-html-tag-for-displaying-side-notes-and-admonitions>

PR #1069, opened 2026-02-09
reword BlockQuoteKind -> AdmonitionKind, render as div

orangecms:
The extended GFM reuses blocks that are usually for quote also for admonitions, so while the notation is ambiguous, we can distinguish the semantics. This is an API breakage for library users.

The rendering as a `<div role="note">` is semantically more adequate and more suitable for assistive tooling. There is no semantic HTML element for this purpose at this point. For reference, see
- <https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/note_role>
- <https://www.w3.org/TR/WCAG21/#name-role-value>

via #1068

### DIFF
--- a/pulldown-cmark/src/lib.rs
+++ b/pulldown-cmark/src/lib.rs
@@ -151,7 +151,7 @@ impl<'a> CodeBlockKind<'a> {
 /// BlockQuote kind (Note, Tip, Important, Warning, Caution).
 #[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
-pub enum BlockQuoteKind {
+pub enum AdmonitionKind {
     Note,
     Tip,
     Important,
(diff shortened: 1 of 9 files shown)


orangecms:
Admonition support in the guide (I think that's what you mean by "the mdBook"?) is currently mentioned in [`index.md`](https://github.com/pulldown-cmark/pulldown-cmark/blob/287fcf9c117ca4eb555ecf2cc5e7c131c01d7b24/guide/src/index.md?plain=1#L7-L8):

> In addition to the base CommonMark language, it supports tables, task lists, strikethrough, footnotes, admonitions, and LaTeX-style math from GFM.

Martin1887:
Yes, I meant `specs/blockquotes_tags.txt`. It should work but it is not in https://pulldown-cmark.github.io/pulldown-cmark, probably it just have to be deployed.

pulldown-cmark/src/lib.rs
```diff
@@ -151,7 +151,7 @@ impl<'a> CodeBlockKind<'a> {
 /// BlockQuote kind (Note, Tip, Important, Warning, Caution).
 #[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
-pub enum BlockQuoteKind {
+pub enum AdmonitionKind {
```
Jules-Bertholet, line 154:
```suggestion
pub enum CalloutKind {
```

I think "admonition" has the wrong connotation here—it's a good fit for warnings, but a tip block is not really an "admonition". "Alert" is another option. (GitHub's docs use all 3: https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts)

Martin1887: Admonition is the typical name for this, but I agree. Maybe we can use the name used by Beamer: `block`.

And then, other kinds of blocks typically used in Beamer become more relevant: `definition`, `theorem`, etc.

orangecms: So, how about `CalloutBlock` here?

e.g. also what Quarto calls it; I think calling it a block makes sense to distinguish from inline things https://quarto.org/docs/authoring/callouts.html

Martin1887:
It seems that we merge this pull request for the next major release after deciding the name and rendering (`div` or `aside`).

Sorry for the long delay and thanks!

PR #1087, opened 2026-04-24
Filter NUL bytes as required by the commonmark spec

### LINKED ISSUE #1065: U+0000 must be replaced with U+FFFD to conform CommonMark spec

notriddle:
Fixes https://github.com/pulldown-cmark/pulldown-cmark/issues/1065

Obsoletes https://github.com/pulldown-cmark/pulldown-cmark/pull/1066

### DIFF
--- a/pulldown-cmark/specs/regression.txt
+++ b/pulldown-cmark/specs/regression.txt
@@ -2918,3 +2918,117 @@ hello world
 ���
 hello�world</p>
 ````````````````````````````````
+
+Replace NUL with U+FFFD in code spans
+
+```````````````````````````````` example
+`first second`
+`start `
+` end`
+` `
+.
+<p><code>first�second</code>
+<code>start�</code>
+<code>�end</code>
+<code>�</code></p>
+````````````````````````````````
+
+Replace NUL with U+FFFD in math spans
+
+```````````````````````````````` example
+$$first second$$
+$$start $$
+$$ end$$
+$$ $$
+.
+<p><span class="math math-display">first�second</span>
+<span class="math math-display">start�</span>
+<span class="math math-display">�end</span>
+<span class="math math-display">�</span></p>
+````````````````````````````````
+
+Replace NUL with U+FFFD in link titles
+
+```````````````````````````````` example
+[x](x "first second")
+[x](x "start ")
+[x](x " end")
+[x](x " ")
+.
+<p><a href="x" title="first�second">x</a>
+<a href="x" title="start�">x</a>
+<a href="x" title="�end">x</a>
+<a href="x" title="�">x</a></p>
+````````````````````````````````
+
+Replace NUL with U+FFFD in link destinations
+EF BF BD is UTF-8 for U+FFFD
+
+https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=064c457a9cbae22d41e355686f9a09ae
+
+```````````````````````````````` example
+[x](<first second>)
+[x](<start >)
+[x](< end>)
+[x](< >)
+.
+<p><a href="first%EF%BF%BDsecond">x</a>
+<a href="start%EF%BF%BD">x</a>
+<a href="%EF%BF%BDend">x</a>
+<a href="%EF%BF%BD">x</a></p>
+````````````````````````````````
+
+Replace NUL with U+FFFD in link reference definitions
+
+```````````````````````````````` example
+[first second]: https://example.com/a
+[start ]: https://example.com/b
+[ end]: https://example.com/c
+[ ]: https://example.com/d
+
+[first�second][]
+[start�][]
+[�end][]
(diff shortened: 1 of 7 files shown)


Martin1887:
Looks good to me!

ollpu (changes requested):
I tried adding a fuzz target that asserts there are no NULs left in the HTML output.

~~Instead, I seem to have found a regression:~~

```
echo -en '<!S\r:' | cargo run

thread 'main' (939493) panicked at pulldown-cmark/src/parse.rs:2412:50:
byte range starts at 4 but ends at 3
<!S
```

~~Does not panic on main.~~

Edit: My bad, not a regression. Filed https://github.com/pulldown-cmark/pulldown-cmark/issues/1112

ollpu (approved):
Ok, this looks good, but it's hard to be confident that it covers everything.

This is a bit of an architectural issue, and is also related to other preprocessing normalizations like CRLF. I don't have a great solution, but fuzzing could be helpful (once the existing failures are fixed).

I've been busy with other things in the last months, sorry about that.

Tests need to be regenerated (see CI).

ollpu:
Ah, here's something:

```
$ echo -e '```\na\0b\n```' | cargo run

<pre><code>ab
</code></pre>
```

ollpu:
And in HTML (I assume it should be replaced there too):

```
$ echo -e '<p>a\0b' | cargo run

<p>ab
```

```
$ echo -e '<span x="a\0b"></span>' | cargo run

<p><span x="ab"></span></p>
```

And heading attributes:

```
$ echo -e '# x { .a\0b }' | cargo run -- -H

<h1 class="ab">x</h1>
```

So yeah, this is tricky.

I'm leaning towards making this the renderer's responsibility instead. Internally, NULs are not a problem for us, so it should be fine to replace them only at the last moment. We just have to document that a renderer should perform this replacement alongside other escaping.

notriddle:
There are a few cases where we *do* have to process NUL into U+FFFD. For example, link refdefs need to be canonicalized:

```console
$ echo -e '[\0][]\n\n[\xef\xbf\xbd]: https://example.com\n' | ./build/src/cmark
<p><a href="https://example.com">�</a></p>
```

I'm wondering if we can implement this normalization in `CowStr::from`? That would cover all of these cases, and it would avoid creating a footgun for our dependents.

ollpu:
> For example, link refdefs need to be canonicalized:

For full compatibility, yes, but is it really mandated by the spec? Does it matter in practice? It's written so tersely that honestly pre/post processing both feel like valid readings.

> I'm wondering if we can implement this normalization in `CowStr::from`?

I was thinking that too, but it requires potentially reallocating long strings that happen to contain `\0`. Maybe if firstpass still tries to split text & HTML nodes around it? Other cases need reallocation anyway.

Performance-wise this requires an extra pass over borrowed strings, whereas doing it in firstpass or escaping, it would be integrated into the special char scanning pass.

Also, seems a bit counterintuitive for external users of CowStr. Maybe it's okay.

ollpu:
A separate `from` function on `CowStr` seems like a good middle ground, assuming that we can plug all (future) holes with fuzzing.

I don't love the name `normalize_from`. Does it imply that the function has a broader scope and should perform other normalizations as well? (Line endings?) If it's just this, maybe it could be e.g. `from_replace_nuls`.

ollpu (approved):

ollpu:
Tried to merge this with #1116 so I could get the fuzzer running. Found this:

```
$ echo -en '~~\0~~' | cargo run -- -e -B # subscript

0..5: Start(Paragraph)
0..5: Text(Borrowed("~~\0~~"))
0..5: End(Paragraph)
EOF
```

Edit: Apparently this is also a broader bug. When subscripts are enabled but strikethrough is not, text inside two tildes is not parsed correctly. #1118

ollpu:
The text of a normal wikilink:

```
$ echo -e '[[Z\0]]' | cargo run -- -W | cat -v

<p><a href="Z%EF%BF%BD">Z^@</a></p>
```

ollpu:
Ok, this is ready to merge now from my end.

PR #1088, opened 2026-04-29
Add an online demo to the book

SabrinaJewson:
This PR adds a WebAssembly-powered online dingus similar to the [commonmark.js dingus](https://spec.commonmark.org/dingus/) to the book.

It allows viewing the rendered HTML, the raw HTML, as well as the pulldown-cmark AST, which is displayed in `dbg!` fashion as a tree.

A screenshot is below:

image

<details><summary>AST view</summary>
image
</details>

I didn’t put much effort into browser support but it should work on all modern browsers.

### DIFF
--- a/guide/demo/src/lib.rs
+++ b/guide/demo/src/lib.rs
@@ -1,0 +1,93 @@
+#![no_std]
+#![expect(clippy::missing_safety_doc)]
+
+use alloc::string::String;
+use core::fmt::Write;
+use dlmalloc::GlobalDlmalloc;
+
+extern crate alloc;
+
+#[global_allocator]
+static ALLOC: GlobalDlmalloc = GlobalDlmalloc;
+
+#[panic_handler]
+#[cfg(target_arch = "wasm32")]
+fn panic(_info: &core::panic::PanicInfo) -> ! {
+    core::arch::wasm32::unreachable()
+}
+
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn alloc(size: usize, align: usize) -> *mut u8 {
+    let layout = alloc::alloc::Layout::from_size_align(size, align).unwrap();
+    unsafe { alloc::alloc::alloc(layout) }
+}
+
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn dealloc(ptr: *mut u8, size: usize, align: usize) {
+    let layout = alloc::alloc::Layout::from_size_align(size, align).unwrap();
+    unsafe { alloc::alloc::dealloc(ptr, layout) };
+}
+
+#[repr(C)]
+pub struct FfiString {
+    ptr: *mut u8,
+    len: usize,
+    cap: usize,
+}
+
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn render(
+    ptr: *mut u8,
+    len: usize,
+    capacity: usize,
+    options: u32,
+    ast: bool,
+    out: *mut FfiString,
+) {
+    let s = unsafe { String::from_raw_parts(ptr, len, capacity) };
+
+    let options = pulldown_cmark::Options::from_bits(options).unwrap();
+    let parser = pulldown_cmark::Parser::new_ext(&s, options);
+
+    let mut res = String::new();
+
+    match ast {
+        false => {
+            pulldown_cmark::html::push_html(&mut res, parser);
+        }
+        true => {
+            res.push_str("<ul>");
+            for event in parser {
+                match event {
+                    pulldown_cmark::Event::End(_) => res.push_str("</ul>"),
+                    _ => res.push_str("<li>"),
+                }
+                write!(HtmlEscape(&mut res), "{event:?}").unwrap();
+                match event {
+                    pulldown_cmark::Event::Start(_) => res.push_str("<ul>"),
+                    _ => res.push_str("</li>"),
+                }
+            }
+            res.push_str("</ul>");
+        }
+    }
+
+    let (ptr, len, cap) = res.into_raw_parts();
+    unsafe { *out = FfiString { ptr, len, cap } };
+}
+
+struct HtmlEscape<'a>(&'a mut String);
+
+impl Write for HtmlEscape<'_> {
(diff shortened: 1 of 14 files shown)


Martin1887:
It looks good, thanks!

Martin1887:
Please fix the dlmalloc version to `0.2` instead of `0.2.13` to update patches (it is currently in 0.2.14 for instance).

Everything else looks good!

SabrinaJewson:
I updated `dlmalloc`. (For what it’s worth, using `0.2` wouldn’t make much sense because that’s synonymous with `0.2.0`, which is earlier than `0.2.13`.)

guide/src/demo.md
```diff
@@ -0,0 +1,254 @@
...
+  oldFootnotes.addEventListener("input", () => {
+    if (oldFootnotes.checked) {
+      footnotes.checked = true;
+    }
+  });
+}
+
+// mdBook adds a document-level keydown listener for the arrow keys for page navigation. This
+// interferes with editing the textarea, so we stop propagation of these keydown events.
+form.elements["md"].addEventListener("keydown", e => e.stopPropagation());
```
notriddle: mdBook already has code that's supposed to take care of this. https://github.com/rust-lang/mdBook/commit/c5c31bb9f5dfb10d66f6b6478395e2068d593949

Are we on an outdated version of mdBook?

guide/demo/src/lib.rs
```diff
@@ -0,0 +1,93 @@
...
+            pulldown_cmark::html::push_html(&mut res, parser);
+        }
+        true => {
+            res.push_str("<ul>");
+            for event in parser {
+                match event {
+                    pulldown_cmark::Event::End(_) => res.push_str("</ul>"),
+                    _ => res.push_str("<li>"),
+                }
+                write!(HtmlEscape(&mut res), "{event:?}").unwrap();
```
notriddle, line 65: Since the `html` feature is enabled, you already transitively depend on [pulldown-cmark-escape](https://github.com/pulldown-cmark/pulldown-cmark/tree/main/pulldown-cmark-escape). You might as well use it.


```suggestion
                write!(pulldown_cmark_escape::FmtWriter(&mut res), "{event:?}").unwrap();
```

notriddle:
None of these are blockers.

guide/src/demo.md
SabrinaJewson: I didn’t know about that! That’s in 0.5.3, while the version used in CI is 0.4.40 and the version I was testing with locally is 0.5.2. I updated and removed this.

guide/demo/src/lib.rs
SabrinaJewson: I don’t think that particular suggestion does anything, because `FmtWriter` is meant to be a wrapper for the type where the escaped HTML is written, not a type that in and of itself does escaping. However, I was able to simplify the code by using this crate.

Martin1887:
> I updated `dlmalloc`. (For what it’s worth, using `0.2` wouldn’t make much sense because that’s synonymous with `0.2.0`, which is earlier than `0.2.13`.)

`0.2` does not mean 0.2.0 but the latest 0.2: https://doc.rust-lang.org/cargo/reference/resolver.html#version-requirements

> # Package A
> [dependencies]
> rand = "0.7"  # meaning `>=0.7.0,<0.8.0`

> # Package B
> [dependencies]
> rand = "0.6"  # meaning `>=0.6.0,<0.7.0`

So please use `0.2` instead :slightly_smiling_face:.

notriddle (approved):

SabrinaJewson:
> 0.2 does not mean 0.2.0 but the latest 0.2: https://doc.rust-lang.org/cargo/reference/resolver.html#version-requirements

It means both:
- `0.2`, as you linked, means `≥0.2.0,<0.3.0`, which results in:
  - `0.2.14` (the latest 0.2 after 0.2.0) when not building with `-Zminimal-versions`
  - `0.2.0` when building with `-Zminimal-versions`
- `0.2.13` means `≥0.2.13,<0.3.0`, which results in:
  - `0.2.14` (the latest 0.2 after 0.2.13) when not building with `-Zminimal-versions`
  - `0.2.13` when building with `-Zminimal-versions`

You can check this for yourself:

```
$ cargo new demo && cd demo
$ echo 'dlmalloc = "0.2.13"' >> Cargo.toml
$ cargo check
$ rg dlmalloc -A=1 Cargo.lock
…
19:name = "dlmalloc"
20-version = "0.2.14"
```

I specified `0.2.13`, but it uses `0.2.14`, because that’s the latest version available. The only difference between specifying `0.2`/`0.2.0` and `0.2.13` is that the former semantically promises that the package _also_ works with `0.2.0`. But I haven’t tested this, and have no intent to test this, so in my opinion it makes more sense to give a more conservative guarantee.

Example showing that `0.2` is the same as `0.2.0`:

```
$ cargo new demo && cd demo
$ echo 'dlmalloc = "0.2"' >> Cargo.toml
$ cargo +nightly check -Zminimal-versions
    Updating crates.io index
     Locking 2 packages to earliest Rust 1.97.0-nightly compatible versions
      Adding dlmalloc v0.2.0 (available: v0.2.14)
      Adding libc v0.2.0 (available: v0.2.186)
  Downloaded libc v0.2.0
  Downloaded dlmalloc v0.2.0
  Downloaded 2 crates (118.5KiB) in 0.08s
   Compiling dlmalloc v0.2.0
    Checking libc v0.2.0
```

This produces a bunch of errors, because `dlmalloc` specified `libc = "0.2"` in its `Cargo.toml` instead of specifying a patch number as is best practice.

See also: [PSA: Please specify precise dependency versions in Cargo.toml](https://users.rust-lang.org/t/psa-please-specify-precise-dependency-versions-in-cargo-toml/71277)

Martin1887:
Good to know, thanks for the detailed explanation!

I thought that at the patch level the last number was a hard requirement, and so `0.2.13``meant exactly 0.2.13. Also, I didn't know the `-Zminimal-versions` flag. And apologies for bothering you with a wrong change.

PR #1090, opened 2026-05-07
feat: add ==highlight== inline extension behind ENABLE_HIGHLIGHT

### LINKED ISSUE #1053: Highlighting support
Obsidian provides a way to highlight a piece of text using ==. It would be great if there was an option for this

mijowi:
Fix #1053 

Adds an opt-in `==highlight==` syntax behind `Options::ENABLE_HIGHLIGHT`, rendered as `<mark>highlight</mark>`. Originates from markdown-it / pandoc; not part of CommonMark or GFM, so it is not enabled by `ENABLE_GFM`.

The implementation mirrors strikethrough's two-character delimiter:
- only `==` (run length 2) forms a delimiter pair; single `=` and `===`+ stay literal
- standard CommonMark flanking rules apply; intraword `a==b==c` works (mirrors `~~`)
- a new `EQUALS` slot is added to `InlineStack::lower_bounds` so a failed `~~` lookup doesn't poison `==` matches via the shared fallback bucket
- new `--enable-highlight` CLI flag


notriddle (approved):

Martin1887:
Please fix the `cargo fmt` errors before merging. Thanks!

mijowi:
> Please fix the `cargo fmt` errors before merging. Thanks!

Should be gtg!

PR #1091, opened 2026-05-07
Recognize lone CR as a line ending in scan_nextline

### LINKED ISSUE #1056: scanners::scan_nextline does not recognize lone CR
```rs
pub(crate) fn scan_nextline(bytes: &[u8]) -> usize {
    memchr(b'\n', bytes).map_or(bytes.len(), |x| x + 1)
}
```

This works fine for CRLF and LF line endings, but a lone CR ("\r") is also valid, [per CommonMark](https://spec.commonmark.org/0.31.2/#line-ending).

This causes some bugs in the places it's used:

```
$ python -c 'print("```\rcode\rblock\r\n```")' | cargo run
<pre><code class="language-codeblock"></code></pre>
```

The fix should be simple, but how could we systematically test for proper handling of different line endings?

dtran24:
Fixes #1056.

## Problem

`scanners::scan_nextline()` currently only searches for `\n`, so it handles LF and CRLF but misses lone CR line endings. CommonMark treats LF, CRLF, and lone CR as valid line endings, so inputs containing lone `\r` can be parsed incorrectly.

One concrete failure case from #1056 is the input `"```\rcode\rblock\r\n```\n"`, which currently produces an empty fenced code block because the intermediate lone `\r` is not treated as a line break.

## Fix

Teach `scan_nextline()` to recognize both `\n` and `\r`, while still treating `\r\n` as a single line ending. This keeps the existing LF/CRLF behavior intact and brings lone CR handling in line with the CommonMark definition of a line ending.

## Changes

- `pulldown-cmark/src/scanners.rs`: switch `scan_nextline()` from `memchr()` to `memchr2()` so it recognizes `\n`, `\r\n`, and lone `\r`
- `pulldown-cmark/tests/html.rs`: add a regression test covering the fenced-code example from #1056
- keep the regression in `html.rs` instead of `regression.txt`, since newline normalization makes the lone-CR case hard to express there

## Test results

- `cargo test -q`
- `cargo test -q issue_1056`

### DIFF
--- a/pulldown-cmark/src/scanners.rs
+++ b/pulldown-cmark/src/scanners.rs
@@ -23,7 +23,7 @@
 use alloc::{string::String, vec::Vec};
 use core::char;
 
-use memchr::memchr;
+use memchr::{memchr, memchr2};
 
 pub(crate) use crate::puncttable::{is_ascii_punctuation, is_punctuation};
 use crate::{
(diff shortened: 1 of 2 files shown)


Martin1887:
This makes sense to me. Thanks!

PR #1092, opened 2026-05-09
fix: update the lifetime of returned RefDefs to match the input

el-pendeloco:
- The previous version erroneously ties the *contents* of the RefDefs to the lifetime of the parser/iterator.


Martin1887:
Good catch, thanks!

PR #1093, opened 2026-05-13
Feature/export link def

el-pendeloco:
- The RefDef struct is exposed, but not its value type (LinkDef). This prevent user code from storing the link definitions in a different data struct or use it in function arguments, for example (without using unsafe).


Martin1887:
Good catch, thanks!

PR #1096, opened 2026-05-19
fix: panic in parser iterator for `TightParagraph` in some cases

### LINKED ISSUE #1095: Parser panic on link-reference-like list item followed by blank tab line
pulldown-cmark 0.13.3 panics while iterating the parser for this 11-byte Markdown input:

```text
- [n]:Z\r\n\t\t
```

Minimal Rust reproducer:

```rust

fn main() {
    let s = "- [n]:Z\r\n\t\t";
    let parser = Parser::new_ext(s, Options::all());
    for _ in parser {}
}
```

Observed panic in a fuzz campaign through mdwright:

```text
thread panicked at .../pulldown-cmark-0.13.3/src/parse.rs:2367: called `Option::unwrap()` on a `None` value
```

The same panic reproduces when collecting events directly from `Parser::new_ext(..., Options::all())`, so it appears independent of mdwright's formatter/linter logic.

Expected: parsing arbitrary Markdown input should not panic; this input should produce some event stream or be handled internally without unwinding.

Martin1887:
Fixes #1095.

### DIFF
--- a/pulldown-cmark/src/parse.rs
+++ b/pulldown-cmark/src/parse.rs
@@ -2446,6 +2446,13 @@ mod test {
     }
 
     #[test]
+    fn issue_1095() {
+        let s = "- [n]:Z\r\n\t\t";
+        let parser = Parser::new_ext(s, Options::all());
+        for _ in parser {}
+    }
+
+    #[test]
     fn issue_1030() {
         let mut opts = Options::empty();
         opts.insert(Options::ENABLE_WIKILINKS);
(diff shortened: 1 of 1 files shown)


Martin1887:
Fixes #1097.

notriddle:
I've added the test case from 1097.

If you're okay with this, then r=me.

notriddle:
I checkout out the [v0.13.3](https://github.com/pulldown-cmark/pulldown-cmark/releases/tag/v0.13.3) tag, and was able to reproduce the problem with that test case.

For some reason, it doesn't actually reproduce with `main`. But at least we can be sure it's not going to reproduce.

Martin1887:
Thanks!

PR #1102, opened 2026-05-26
Fix: wikilink with pipe separator leaks child node's next chain into sibling content

wesc:
handle_wikilink creates a cycle/DAG in the tree when a pipe separator is present, causing exponential event-count growth.

When a wikilink with a pipe separator is parsed (e.g. [[name|display]]), handle_wikilink sets the wikilink node's child pointer to body_node without terminating body_node's next chain. Because body_node is a reused node whose next chain still threads through cur_ix (the closing ] delimiter) and into the content that follows the ]], the EventIter visits those trailing nodes twice: once as children of the wikilink (via the child pointer) and once as siblings (via the wikilink node's own next pointer).

With repeated adversarial patterns like `[[[[^(\n|]]]]=]]]]]]]]\n`, this causes super-polynomial event-count growth effectively hanging the parser. The fix adds a short chain-termination walk starting from body_node that finds and severs the next link that would otherwise reach cur_ix, ensuring the child subtree is properly bounded before the wikilink node is installed into the tree.

This bug was discovered via fuzzing.

This PR contains two tests for the behavior and a fix. There are no regressions against other tests.

### DIFF
--- a/pulldown-cmark/specs/wikilinks.txt
+++ b/pulldown-cmark/specs/wikilinks.txt
@@ -146,13 +146,21 @@ Wikilinks cannot be empty. They will render as-is.
 <p>]] [[]] [[|]] [[|Symbol]] [[</p>
 ````````````````````````````````
 
+If the display text is empty then we omit it from the output anchor.
+
+```````````````````````````````` example_wikilinks
+[[link|]]
+.
+<p><a href="link"></a></p>
+````````````````````````````````
+
 When a wikilink with a pipe separator is nested inside extra brackets, the
 trailing `]]` after the wikilink must appear exactly once.
 
 ```````````````````````````````` example_wikilinks
-[[[[link|]]]]
+[[[[link|display]]]]
 .
-<p>[[<a href="link">]</a>]]</p>
+<p>[[<a href="link">display</a>]]</p>
 ````````````````````````````````
 
 Other interactions wikilinks have with other Markdown syntax:
--- a/pulldown-cmark/specs/wikilinks.txt
+++ b/pulldown-cmark/specs/wikilinks.txt
@@ -146,6 +146,15 @@ Wikilinks cannot be empty. They will render as-is.
 <p>]] [[]] [[|]] [[|Symbol]] [[</p>
 ````````````````````````````````
 
+When a wikilink with a pipe separator is nested inside extra brackets, the
+trailing `]]` after the wikilink must appear exactly once.
+
+```````````````````````````````` example_wikilinks
+[[[[link|]]]]
+.
+<p>[[<a href="link">]</a>]]</p>
+````````````````````````````````
+
 Other interactions wikilinks have with other Markdown syntax:
 
 ```````````````````````````````` example_wikilinks
--- a/pulldown-cmark/src/parse.rs
+++ b/pulldown-cmark/src/parse.rs
@@ -3003,41 +3003,4 @@ text
             n8 / n1.max(1)
         );
     }
-
-    #[test]
-    fn wikilink_pipe_no_duplicate_siblings() {
-        // `[[[[name|]]]]`
-        // The outer `[[` become Text; the inner `[[` form a wikilink.
-        // After the wikilink the remaining `]]` must appear exactly
-        // once as siblings.
-        let input = "[[[[^(\n|]]]]";
-        let events: Vec<_> = Parser::new_ext(input, Options::ENABLE_WIKILINKS)
-            .into_offset_iter()
-            .collect();
-
-        // Collect non-structural events (excluding Start/End wrappers
-        // whose ranges legitimately repeat).  Each Text event's byte
-        // range must appear at most once.
-        let text_ranges: Vec<_> = events
-            .iter()
-            .filter_map(|(e, r)| matches!(e, Event::Text(_)).then_some(r.clone()))
-            .collect();
-
-        let mut seen = HashMap::new();
-        for r in &text_ranges {
-            *seen.entry(r.clone()).or_insert(0usize) += 1;
-        }
-        let dups: Vec<_> = seen.iter().filter(|(_, &c)| c > 1).collect();
-        assert!(
-            dups.is_empty(),
-            "Text events for byte ranges {:?} were emitted more than once.\n\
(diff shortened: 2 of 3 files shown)


pulldown-cmark/src/parse.rs
```diff
@@ -2957,4 +2980,64 @@ text
...
+
+        // Collect non-structural events (excluding Start/End wrappers
+        // whose ranges legitimately repeat).  Each Text event's byte
+        // range must appear at most once.
+        let text_ranges: Vec<_> = events
+            .iter()
+            .filter_map(|(e, r)| matches!(e, Event::Text(_)).then_some(r.clone()))
+            .collect();
+
+        let mut seen = std::collections::HashMap::new();
```
notriddle: Need to either make this test `#[cfg(feature=std)]`or use hashbrown.

pulldown-cmark/src/parse.rs
```diff
@@ -2957,4 +2980,64 @@ text
         ];
         assert_eq!(&events, &expected);
     }
+
+    #[test]
```
notriddle, line 2978: Is it possible to write a test for this in https://github.com/pulldown-cmark/pulldown-cmark/blob/main/pulldown-cmark/specs/wikilinks.txt instead?

Does this change affect the resulting HTML in any circumstances? It seems like it should.

wesc: I pushed up a commit that moves one of the two tests into wikilinks.txt. I'd be happy to do more testing too if needed.

pulldown-cmark/specs/wikilinks.txt
```diff
@@ -146,6 +146,15 @@ Wikilinks cannot be empty. They will render as-is.
 <p>]] [[]] [[|]] [[|Symbol]] [[</p>
 ````````````````````````````````
 
+When a wikilink with a pipe separator is nested inside extra brackets, the
+trailing `]]` after the wikilink must appear exactly once.
+
+```````````````````````````````` example_wikilinks
+[[[[link|]]]]
+.
+<p>[[<a href="link">]</a>]]</p>
```
notriddle: I’m confused. Where did the “]” inside the a element come from?

wesc: great catch.. I've been staring at these brackets and am falling into a daze. The bug is in the code I submitted.. the expectation is that scan_nodes_to_ix points to the display text, but in this particular example it points to the first closing ]. I will write some more tests.

wesc: What is the correct rendering for:

```
[[link|]]
```

Should we produce `<p><a href="link"></a></p>` or is that considered an error? We expect

```
[[|Symbol]]
```

to pass through and render as-is, so should we also expect `[[link|]]` to render as is?

notriddle: [`<a href="link"></a>`](https://pandoc.org/try/?params=%7B%22text%22%3A%22%5B%5Blink%7C%5D%5D%22%2C%22to%22%3A%22html5%22%2C%22from%22%3A%22commonmark%2Bwikilinks_title_after_pipe%22%2C%22standalone%22%3Afalse%2C%22embed-resources%22%3Afalse%2C%22table-of-contents%22%3Afalse%2C%22number-sections%22%3Afalse%2C%22citeproc%22%3Afalse%2C%22html-math-method%22%3A%22plain%22%2C%22wrap%22%3A%22auto%22%2C%22highlight-style%22%3Anull%2C%22files%22%3A%7B%7D%2C%22template%22%3Anull%7D)

wesc: This should be addressed in a new commit https://github.com/pulldown-cmark/pulldown-cmark/pull/1102/changes/65e5fd6c2e072fbc7d50c6935dd0d0685b159127

Two tests in wikilinks.txt, one for `[[[[foo|bar]]]]` and `[[foo|]]`. It's not in wikilinks.txt but I also verified that `[[[[foo|]]]]` parses as expected.

pulldown-cmark/src/parse.rs
```diff
@@ -953,14 +953,46 @@ impl<'input> ParserInner<'input> {
...
+                        let body_node = scan_nodes_to_ix(&self.tree, Some(body_node), rest);
+                        if let Some(body_node) = body_node {
+                            // break node so passes can actually format
+                            // the display text
+                            self.tree[body_node].item.start = rest;
+
+                            // Terminate the child chain so it does not extend
+                            // past the closing `]]` into sibling content.
+                            let mut scan = body_node;
+                            loop {
```
notriddle: I've been wondering for awhile why this loop was needed, and, finally, I figured it out.

It shouldn't be needed. This code is supposed to take care of fixing the same problem.

https://github.com/wesc/pulldown-cmark/blob/65e5fd6c2e072fbc7d50c6935dd0d0685b159127/pulldown-cmark/src/parse.rs#L1022-L1024

As long as `prev` is correctly assigned, you shouldn't need that.

notriddle:
I've pushed a commit that fixes the `prev` assignment and removes up the unneeded loop.

notriddle (approved):

zoosky:
@wesc can you be more specific?

> this bug was discovered by fuzzing

wesc:
> @wesc can you be more specific?
> 
> > this bug was discovered by fuzzing

I used a fuzzer (cargo-fuzz), which injects a large amount of random input into select functions looking for crashes or pathological behavior. I used it on an downstream library (EmbedAnything) which flagged issues there as well as in pulldown-cmark. So I came here to work on a fix. After I opened the PR, I found this: https://github.com/pulldown-cmark/pulldown-cmark/issues/1076, and also evidence in this repo that others have been fuzzing.

Fuzzing is producing another crash bug: a buffer overflow and access to invalid memory that, in my tests, only crashes when running the fuzzer. This doesn't seem to affect production builds, but that's just luck. I'll open a separate PR for that but I don't understand the fix that AI has provided just yet so I want to get a handle on it first.

It would probably be a good idea to regularly fuzz this library.. cargo fuzz is finding these issues within a few minutes of running so it's a cheap thing to set up.

PR #1105, opened 2026-06-03
feat: Add support for Pandoc-style table captions

amiroo54:
Currently, there is no native way to add captions to tables. This PR introduces support for table captions inspired by the [Pandoc table_caption specification](https://pandoc.org/demo/example33/8.9-tables.html), using the : prefix syntax immediately following a table.

Changes included:

    Caption Parsing: Added logic to parse lines beginning with : immediately after a table as a caption node.
    Feature Flag: Wrapped the functionality in a new ENABLE_TABLE_CAPTIONS option flag to ensure strict backward compatibility for existing users.

Note on Spec Compliance:

To keep this initial implementation focused and lightweight, I scoped it to support captions after the table using the : prefix. Pandoc’s alternative syntaxes (like Table: Caption before the table) are not included in this PR, but the architecture allows for them to be added later if desired.


Martin1887:
Thanks! Please add more tests in a spec file like for other features. Also please note the errors in the automated workflows :slightly_smiling_face:.

amiroo54:
Done!

Martin1887:
Nice, it looks good to me. The only thing that concerns me is that if we are adopting Pandoc spec, I would adopt the full spec allowing also captions in top of tables.

What do you think, @notriddle?

pulldown-cmark/src/firstpass.rs
notriddle, line 653: Can you add a test case for this feature when table captions and definition lists are enabled at the same time?

I checked, and it works, but these features both use `:`, which means that the interaction between them should be tested.

pulldown-cmark/src/firstpass.rs
```diff
@@ -660,11 +660,42 @@ impl<'a, 'b> FirstPass<'a, 'b> {
         ) {
             return None;
         }
+        if scan_ch(&bytes[ix..], b':') == 1 && self.options.contains(Options::ENABLE_TABLE_CAPTIONS)
```
notriddle, line 663: Here's a tricky test case

~~~
```````````````````````````````` example_table_captions
Test|Table
----|-----
Test|row
: Test Caption
Test|ending
.
<table><thead><tr><th>Test</th><th>Table</th></tr></thead><tbody>
<tr><td>Test</td><td>row</td></tr>
<caption>Test Caption</caption>
<tr><td>Test</td><td>ending</td></tr>
</tbody></table>
~~~

It reflects the behavior [of pandoc](https://pandoc.org/try/?params=%7B%22text%22%3A%22Test%7CTable%5Cn----%7C-----%5CnTest%7Crow%5Cn%3A+Test+Caption%5CnTest%7Cending%5Cn%22%2C%22to%22%3A%22html5%22%2C%22from%22%3A%22markdown%22%2C%22standalone%22%3Afalse%2C%22embed-resources%22%3Afalse%2C%22table-of-contents%22%3Afalse%2C%22number-sections%22%3Afalse%2C%22citeproc%22%3Afalse%2C%22html-math-method%22%3A%22plain%22%2C%22wrap%22%3A%22auto%22%2C%22highlight-style%22%3Anull%2C%22files%22%3A%7B%7D%2C%22template%22%3Anull%7D), except that `<caption>` is supposed to be the first child of the table element.

File: pulldown-cmark/src/firstpass.rs [hunk 2]
```diff
@@ -660,11 +660,42 @@ impl<'a, 'b> FirstPass<'a, 'b> {
...
+        }
 
         let (ix, row_ix) = self.parse_table_row_inner(ix, row_cells, missing_empty_cells)?;
         Some((ix, row_ix))
     }
 
+    fn parse_table_caption(&mut self, mut ix: usize) -> Option<(usize, TreeIndex)> {
+        let bytes = self.text.as_bytes();
+        ix += scan_ch(&bytes[ix..], b':');
+        ix += scan_whitespace_no_nl(&bytes[ix..]);
```
notriddle, line 676: Note that there are no test cases for the zero-spaces caption:

```
Test|Table
----|-----
Test|row
:Test Caption
Test|ending
```

That seems like a good thing to test, also.

pulldown-cmark/specs/table_captions.txt
```diff
@@ -0,0 +1,123 @@
...
+Test|Table
+----|-----
+Test|row
+: Test Caption
+
+Test ending
+.
+<table><thead><tr><th>Test</th><th>Table</th></tr></thead><tbody>
+<tr><td>Test</td><td>row</td></tr>
+<caption>Test Caption</caption>
```
notriddle, line 34: This isn't valid HTML.

According to [the standard](https://html.spec.whatwg.org/#the-caption-element), you are required to put the `<caption>` tag as the first child of the `<table>` element.

I've pushed a commit that fixes this to https://github.com/amiroo54/pulldown-cmark/compare/main...notriddle:pulldown-cmark:html-validation

pulldown-cmark/src/firstpass.rs
amiroo54: Makes sense. Done it.

pulldown-cmark/specs/table_captions.txt
```diff
@@ -0,0 +1,123 @@
+Run this with `cargo test --features gen-tests suite::table_captions`.
+
+# Table Captions, inspired by Pandoc's table captions
```
notriddle, line 3: Pandoc supports captions *before* tables, too. How hard would it be?

notriddle (changes requested):

pulldown-cmark/specs/table_captions.txt
amiroo54: Thanks, appreciated.

amiroo54:
A quick reminder, isn't this going to be merged?

Martin1887:
I think that we are both waiting to support for caption before tables to match Pandoc spec :slightly_smiling_face:.

amiroo54:
I was hoping we could do without that. I'm in the middle of some exams and can't work on it for a while. It'll have to wait for the duration I guess.

notriddle:
I submitted [a port of this feature to commonmark-hs](https://github.com/jgm/commonmark-hs/pull/173), because if we’re going to invent a new syntax, I wanted at least one other implementation of it to exist.

In the past, pulldown-cmark implemented a limited version of footnotes, and we got constant complaints until we fixed it by implementing a more typical version. I don’t want to do that again. If we’re adding this, I want *cross-markdown implementation consensus*.

PR #1111, opened 2026-07-08
Fix subtract-overflow panic in handle_wikilink on malformed input

### LINKED ISSUE #1108: Abort with input  `�  ![[] ]()]]`
A fuzzer found this bug.

Input:

```
 �  ![[] ]()]]
```

More details in https://oss-fuzz.com/testcase-detail/5107726603845632.

teddytennant:
`FirstPass::handle_wikilink` computes `start_ix` (the wikilink body node's start) and `end_ix` (the current node's start), then uses `end_ix - start_ix` as a length and slices `&block_text[start_ix..end_ix]`. On some malformed image/wikilink inputs the body node begins at or after the closing delimiter, so `start_ix > end_ix`: the subtraction underflows and panics (`attempt to subtract with overflow` in debug builds), and the slice range is inverted.

This adds an early bail-out — matching the function's documented "bail early in case the link is malformed" behavior and its existing `return None;` idiom — right after both indices are computed:

```rust
if end_ix <= start_ix {
    return None;
}
```

### Test

Added a regression test that runs the reproducer from the issue (`" \u{fffd}  ![[] ]()]]"`) through `Parser::new_ext(.., Options::all())` and `html::push_html`, asserting it no longer panics. Without the guard this input panics with the subtraction overflow; with it, the full test suite (1212 tests) passes.

Closes #1108.

### DIFF
--- a/pulldown-cmark/specs/regression.txt
+++ b/pulldown-cmark/specs/regression.txt
@@ -2879,3 +2879,14 @@ Link must be separated from title by at least one space
 <a href="https://example.com" title="test">a</a>
 <a href="https://example.com" title="test">a</a></p>
 ````````````````````````````````
+
+https://github.com/pulldown-cmark/pulldown-cmark/issues/1108
+
+A malformed image-wrapped wikilink must not underflow when the link body
+starts at or after the closing delimiter.
+
+```````````````````````````````` example_wikilinks
+![[] ]()]]
+.
+<p><img src="" alt="[] " />]]</p>
+````````````````````````````````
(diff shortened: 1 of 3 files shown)


Martin1887:
Thanks! Please write the test in the form of the other regression tests in `specs/regression.txt`.

teddytennant:
Done — moved it into `specs/regression.txt` as a wikilinks regression example (with the minimal repro `![[] ]()]]`), matching the other entries, and dropped the standalone unit test. Regenerated `tests/suite/regression.rs` and confirmed the suite passes.

PR #1113, opened 2026-07-17
Fix panic on HTML block ending with a lone carriage return

### LINKED ISSUE #1112: Panic on `<!S\r:`
```
echo -en '<!S\r:' | cargo run

thread 'main' (939493) panicked at pulldown-cmark/src/parse.rs:2412:50:
byte range starts at 4 but ends at 3
<!S
```

Found with the fuzzer.

SAY-5:
Fixes #1112

`append_html_line` normalizes a trailing CRLF into two separate `Html` items by checking `bytes[end - 2] == b'\r'`. That check assumes the line ends with a two-byte `\r\n`, but `scan_nextline` also treats a lone `\r` as a line ending (per CommonMark), so `end - 2` can point at the previous line's carriage return. On input like `<!S\r:` the second line's slice is built with `start = 4` and `end = 3`, which panics with "byte range starts at 4 but ends at 3".

The guard now only splits when the line actually ends with `\r\n` (`bytes[end - 1] == b'\n' && bytes[end - 2] == b'\r'`), and requires at least two bytes so the indexing can't underflow. Added the fuzzer input as a regression test in `tests/errors.rs`.

### DIFF
--- a/pulldown-cmark/src/firstpass.rs
+++ b/pulldown-cmark/src/firstpass.rs
@@ -1602,7 +1602,8 @@ impl<'a, 'b> FirstPass<'a, 'b> {
                 body: ItemBody::SynthesizeText(cow_ix),
             });
         }
-        if self.text.as_bytes()[end - 2] == b'\r' {
+        let bytes = self.text.as_bytes();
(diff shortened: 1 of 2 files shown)


ollpu:
Fixed by another PR, so closing this.

PR #1121, opened 2026-07-27
Fix subscript bug with strikethrough disabled and wrong pair

### LINKED ISSUE #1118: ENABLE_SUBSCRIPT without ENABLE_STRIKETHROUGH is buggy
```
$ echo '~~**a**~~' | cargo run -- -B

<p>~~**a**~~</p>
```

```
$ echo '~~[x](https://x)~~' | cargo run -- -B

<p>~~[x](https://x)~~</p>
```

It seems like no other syntax inside the "strikethrough" is recognized. I guess the subscript implementation is trying to delegate to the strikethrough code here, but it's not enabled.

notriddle:
Fixes #1118

### DIFF
--- a/pulldown-cmark/build.rs
+++ b/pulldown-cmark/build.rs
@@ -86,7 +86,7 @@ fn {}_test_{i}() {{
     let original = r##"{original}"##;
     let expected = r##"{expected}"##;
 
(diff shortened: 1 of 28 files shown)


ollpu (approved):
Ah, I see what the problem was.

There are still some fishy interactions, but maybe these are unavoidable:

```
$ cargo run -- -SB

~a ~~b~~ c~
<p><sub>a <del>b</del> c</sub></p>
```

```
$ cargo run -- -B

~a ~~b~~ c~
<p>~a ~~b~~ c~</p>
```

PR #1122, opened 2026-07-27
RFC: Specify Options in spec tests using bitflags from_str

ollpu:
I feel that the current system where we define ad-hoc test types `example_deflists` etc. and pass those around as booleans is starting reach its limits. Instead, we could use `bitflags::parser::from_str` to define the `Options` directly.

For convenience, I added two ways to set options. The effective options are the union of both.

Global options can be added to a file by writing on the first line:

```
Parser options: `ENABLE_ABC | ENABLE_XYZ`
```

Further options can be written in the lang string:

```````````````````````````````````
```````````````````````````````` example ENABLE_YUV
```````````````````````````````````

I also made it so that the test generator `struct Spec` keeps track of the current line in the file and names the test like `blockquotes_tags_test_0005_l44`, so it's easier to locate.


Martin1887:
Nice!

pulldown-cmark/specs/cjk_friendly_emphasis.txt
ollpu: This one test prevents setting `ENABLE_CJK_FRIENDLY_EMPHASIS` on the whole file. Should we just move it to regression.txt? No other spec seems to have negative tests like this.

pulldown-cmark/specs/heading_attrs.txt
notriddle: `.example`?

Have these tests ever actually been run?! 🤦

pulldown-cmark/specs/cjk_friendly_emphasis.txt
notriddle: I agree. Move it to regression.txt

notriddle (changes requested):
I pushed a fix to the heading_attrs.txt problem here: https://github.com/notriddle/pulldown-cmark/commit/26306b4bda68cc53146bd1287b52c2de4011a92e

pulldown-cmark/specs/heading_attrs.txt
ollpu: Damn. When changing the test collector, I was kind of worried that it's easy to accidentally write a non-test, since it needs the exact number of backticks and ` example`.

Would it make sense to warn about lines that have a large number of backticks but don't match, or something like that?

Martin1887: Makes sense. Accepting anything with more than 8 backticks is another alternative.

We could also warn about things with a hamming distance lower than 4 for instance or matching a regular expression similar to the accepted values to avoid things like `.example`, but probably not so useful and prone to false positives.

notriddle (approved):

PR #1124, opened 2026-07-31
Keep the task list marker as literal text when item content is a setext heading

### LINKED ISSUE #1115: Task list marker with `ENABLE_TASKLISTS` can end up inside a heading
```rust
use pulldown_cmark::{html, Options, Parser};

fn main() {
    let input = "- [ ] a\n  - ";
    let mut options = Options::empty();
    options.insert(Options::ENABLE_TASKLISTS);

    for event in Parser::new_ext(input, options) {
        println!("{:?}", event);
    }

    let parser = Parser::new_ext(input, options);
    let mut html_out = String::new();
    html::push_html(&mut html_out, parser);
    println!("---HTML---");
    println!("{}", html_out);
}
```

Output:

```
Start(List(None))
Start(Item)
Start(Heading { level: H2, id: None, classes: [], attrs: [] })
TaskListMarker(false)
Text(Borrowed("a"))
End(Heading(H2))
End(Item)
End(List(false))
---HTML---
<ul>
<li>
<h2><input disabled="" type="checkbox"/>
a</h2>
</li>
</ul>
```

The `TaskListMarker` event, and the resulting `<input type="checkbox">`, land inside the `Heading` (a setext `h2` formed by the `  -` underline), producing `<h2><input .../>a</h2>`. Per GFM a task list marker is only produced when the list item's content is a paragraph; here the content is a heading, so no marker should be emitted.

Tested on pulldown-cmark 0.13.4.

BTW, this bug was found using [hegel](https://crates.io/crates/hegeltest). Happy to contribute the tests if you're interested.

hdimer:
Fixes #1115.

With `ENABLE_TASKLISTS`, a task list marker is appended as the first child of the item's paragraph before the paragraph is parsed. If that paragraph turns out to be a setext heading (`  -` underline), the body is rewritten to `Heading` but the marker stayed inside, producing `<h2><input type="checkbox">a</h2>`.

Per GFM a task list marker is only valid when the item content is a paragraph, so drop the leading marker when the paragraph becomes a setext heading. A marker on an earlier paragraph (when only a later block becomes a heading) is unaffected.

Regression tests cover the reported case, the earlier-paragraph case, and heading attributes.

Used AI assistance on this; I reviewed and tested it.

### DIFF
--- a/pulldown-cmark/specs/regression.txt
+++ b/pulldown-cmark/specs/regression.txt
@@ -2909,8 +2909,9 @@ d ~~x ~[x](https://x)~ x~~
 https://github.com/pulldown-cmark/pulldown-cmark/issues/1115
 
 A task list marker is only valid when the list item content is a
-paragraph. When the paragraph becomes a setext heading, the marker
-must not be emitted inside the heading.
+paragraph. When the paragraph becomes a setext heading, the task-list
+interpretation is canceled and the marker bytes are kept as literal
+text instead of being emitted inside the heading.
 
 ```````````````````````````````` example
 - [ ] a
@@ -2918,7 +2919,7 @@ must not be emitted inside the heading.
 .
 <ul>
 <li>
-<h2>a</h2>
+<h2>[ ] a</h2>
 </li>
 </ul>
 ````````````````````````````````
@@ -2941,7 +2942,7 @@ x</p>
 </ul>
 ````````````````````````````````
 
-Heading attributes still apply when the marker is dropped (#1115).
+Heading attributes still apply when the marker is kept as text (#1115).
 
 ```````````````````````````````` example
 - [x] a {#id}
@@ -2949,7 +2950,21 @@ Heading attributes still apply when the marker is dropped (#1115).
 .
 <ul>
 <li>
-<h2 id="id">a</h2>
+<h2 id="id">[x] a</h2>
+</li>
+</ul>
+````````````````````````````````
+
+Only the separating whitespace is kept with the marker text, so a
+backslash escape right after it still resolves normally (#1115).
+
+```````````````````````````````` example
+- [ ] \*a\*
+  -
+.
+<ul>
+<li>
+<h2>[ ] *a*</h2>
 </li>
 </ul>
 ````````````````````````````````
--- a/pulldown-cmark/specs/regression.txt
+++ b/pulldown-cmark/specs/regression.txt
@@ -2905,3 +2905,51 @@ d ~~x ~[x](https://x)~ x~~
 <p>c ~x ~~<a href="https://x">x</a>~~ x~</p>
 <p>d ~~x <sub><a href="https://x">x</a></sub> x~~</p>
 ````````````````````````````````
+
+https://github.com/pulldown-cmark/pulldown-cmark/issues/1115
+
+A task list marker is only valid when the list item content is a
+paragraph. When the paragraph becomes a setext heading, the marker
+must not be emitted inside the heading.
+
+```````````````````````````````` example
+- [ ] a
+  -
+.
+<ul>
+<li>
+<h2>a</h2>
+</li>
+</ul>
+````````````````````````````````
+
+A task list marker on an earlier paragraph is kept when a *later*
+block in the same item becomes a setext heading (#1115).
+
+```````````````````````````````` example
+- [ ] x
+
+  y
+  -
+.
+<ul>
+<li>
+<p><input disabled="" type="checkbox"/>
+x</p>
+<h2>y</h2>
+</li>
+</ul>
+````````````````````````````````
+
+Heading attributes still apply when the marker is dropped (#1115).
+
+```````````````````````````````` example
+- [x] a {#id}
+  -
+.
+<ul>
+<li>
+<h2 id="id">a</h2>
+</li>
+</ul>
+````````````````````````````````
--- a/pulldown-cmark/src/firstpass.rs
+++ b/pulldown-cmark/src/firstpass.rs
@@ -884,11 +884,21 @@ impl<'a, 'b> FirstPass<'a, 'b> {
         );
 
         // A task list marker is only valid when the list item's content is a
-        // paragraph. If this paragraph turned out to be a setext heading, drop
-        // the leading marker so it doesn't end up inside the heading (#1115).
+        // paragraph. If this paragraph turned out to be a setext heading, the
+        // marker would render inside the heading (#1115). Cancel the task-list
+        // interpretation and keep the marker's `[ ]`/`[x]` bytes as literal
+        // text (the node already spans them) rather than silently dropping it.
         if let Some(child_ix) = self.tree[node_ix].child {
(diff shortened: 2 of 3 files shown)


ollpu:
Thanks for the contribution. +1 for AI disclosure.

This test is not what I expected:

```
- [ ] a
  -
.
<ul>
<li>
<h2>a</h2>
</li>
</ul>
```

The square brackets should be preserved (and not just reinserted, the parsing should be canceled completely in this case). That said, I'm not sure how to accomplish that cleanly.

notriddle:
GitHub's own behavior seems correct, and we should try to match it, I think?

```markdown
- [ ] a
  -

- [ ] b
  =

- [ ] c
  x
```

- [ ] a
  -

- [ ] b
  =

- [ ] c
  x

notriddle:
Alternatively, we could try to match [what pandoc does](https://pandoc.org/try/?params=%7B%22text%22%3A%22-+%5B+%5D+a%5Cn++-%5Cn%5Cn-+%5B+%5D+b%5Cn++%3D%5Cn%5Cn-+%5B+%5D+c%5Cn++x%22%2C%22to%22%3A%22gfm%22%2C%22from%22%3A%22gfm%22%2C%22standalone%22%3Afalse%2C%22embed-resources%22%3Afalse%2C%22table-of-contents%22%3Afalse%2C%22number-sections%22%3Afalse%2C%22citeproc%22%3Afalse%2C%22html-math-method%22%3A%22plain%22%2C%22wrap%22%3A%22auto%22%2C%22highlight-style%22%3Anull%2C%22files%22%3A%7B%7D%2C%22template%22%3Anull%7D), and just put the task marker outside the header.

Both answers seem defensible to me. Just, please don't silently eat the `[ ]`.

hdimer:
Thanks both, agreed the marker shouldn't vanish. I pushed a revision that cancels the task-list interpretation and keeps the marker's own `[ ]`/`[x]` source bytes as literal text:

```
- [ ] a          ->  <h2>[ ] a</h2>
  -
- [x] a {#id}    ->  <h2 id="id">[x] a</h2>
  -
```

It reuses the marker node's existing span (only growing it across the separating whitespace) rather than synthesizing a new string, so it's the "cancel completely" variant rather than a reinsertion. A following backslash escape isn't pulled in (`- [ ] \*a\*` over a rule still gives `<h2>[ ] *a*</h2>`), and normal task lists are untouched.

One edge worth flagging: since the brackets become a single literal text node, they no longer take part in reference-link resolution, so `- [x] foo` over a rule with a matching `[x]:` definition would render `[x]` literally rather than as a shortcut link (`[ ]` is always safe, since a whitespace label can't match a definition). If you'd rather match pandoc and put the marker outside the heading instead, I'm happy to switch, both read fine to me.

notriddle:
I'm not picky about which implementation we follow, as long as we actually do it. Since GitHub resolves link defs, we should too if we're following their lead.

```markdown

- [x] h1
  =

[x]: https://example.com
```

- [x] h1
  =

[x]: https://example.com

ollpu:
> Since GitHub resolves link defs

Ah, about that...

Even if you have a perfectly normal checked task, it will prioritize the link because this is done in postprocessing.

```
- [x] abc

[x]: https://example.com
```

- [x] abc

[x]: https://example.com

But yes, if we reject the`[x]` as a tasklist marker, I think it should be eligible to be parsed as a link.

hdimer:
Makes sense, thanks. So the agreed behavior: reject `[x]` as a task marker, but let those bytes go through normal inline parsing, so a matching `[x]:` definition still resolves as a link (matching GitHub's postprocessing order).

My current commit converts the marker to a single literal-text node, which is inert. To get link-eligibility I'll re-scan the marker's span as inline content instead, so it yields the usual `MaybeLinkOpen` / text / `MaybeLinkClose` nodes and the second pass resolves it. Then `- [x] foo` over a setext rule with a `[x]: url` definition renders the heading with a link, while `- [ ] a` stays literal (a whitespace-only label can't match a definition).

Sound right? If so I'll push that version.

hdimer:
Pushed that version. When the task-list interpretation is canceled (item content is a setext heading), `[x]`/`[ ]` now split back into the normal `MaybeLinkOpen` / text / `MaybeLinkClose` inline nodes instead of a single inert text node, so a matching `[x]: url` definition resolves as a link (matching GitHub's postprocessing). Added a regression example covering it. A whitespace-only label like `[ ]` can't match a definition, so it still renders literally. Full suite + `fmt` + `clippy` clean; the existing setext cases are unchanged when no definition is present.

notriddle:
I found another problem with this PR. It turns out, you can reproduce the same bug using tables, not just headings.

```markdown
- [x] | one | two | three    |
  |-----|-----|-------|------|
  | one | two | three | four |
```

GitHub:

- [x] | one | two | three    |
  |-----|-----|-------|------|
  | one | two | three | four |

pulldown-cmark:

```console
$ cargo run --release -- --enable-tables --enable-tasklists
    Finished `release` profile [optimized] target(s) in 0.07s
     Running `target/release/pulldown-cmark --enable-tables --enable-tasklists`
- [x] | one | two | three
  |-----|-----|-------
  | one | two | three
<ul>
<li><table><thead><tr><th>one</th><th>two</th><th>three</th></tr></thead><tbody>
<tr><td>one</td><td>two</td><td>three</td></tr>
</tbody></table>
</li>
</ul>
```

And definition lists have a slightly different, but still basically similar, bug:

```console
$ cargo run --release -- --enable-definition-list --enable-tasklists
    Finished `release` profile [optimized] target(s) in 0.07s
     Running `target/release/pulldown-cmark --enable-tables --enable-tasklists`
- [x] a
  : b
<ul>
<li>
<dl>
<dt><input disabled="" type="checkbox" checked=""/>
a</dt>
<dd>b</dd>
</dl>
</li>
</ul>
```

hdimer:
Both reproduce, both fixed.

Tables were the more interesting one. The marker's bytes are ordinary row content, so I moved the header-row scan back to the marker's start instead of the paragraph's. That means the marker counts as a leading cell, which is what makes your example agree with GitHub: four header cells (`[x]`, one, two, three) against the four-column delimiter row, rather than three-vs-four and no table at all.

Worth flagging the flip side, since it's a behavior change beyond the bug: `- [x] | a | b |` over `|---|---|` used to be a two-column table and is now a plain task list item, because the header is three cells now. GitHub does the same, and there's a spec example pinning it.

Definition lists just reuse the setext cancellation, so `- [x] a` / `  : b` gives `<dt>[x] a</dt>`, and a matching `[x]:` definition still resolves to a link there too.

One thing I noticed while doing this: cancelling left the marker's bytes outside the heading's own range, so `into_offset_iter` reported text events starting before the node containing them. Fixed for all three paths and added a test, since nothing asserted it.

Full suite + `fmt` + `clippy` clean. I also brute-forced ~540k generated bullet/marker/pipe/setext/deflist combinations for panics and bad ranges, and confirmed the structural table decision is now identical with `ENABLE_TASKLISTS` on and off, which it wasn't before.

Martin1887:
Closed in favor of #1135.

